PR #25, opened 2015-03-25
Deprecate log4go and fix gometalinter errors

fern4lvarez:
Pending:
- [x] Fix gometalinter errors in `server` package (and fix #24)
- [x] Document `log` package

- ~~[ ] Normalize log messages (i.e. start all with capital letter, s/unable/failed, etc.)~~
- ~~[ ] Log to stderr when it comes to error logs (ignore if writing to log file)~~

Pull request powered by [oscillatingworks](http://oscillating.works).

PR #33, opened 2015-04-12
Fix data path creation and merge #25 again

fern4lvarez:
With #25 a bug was introduced, so rqlite raised a panic whenever
starting because data path wasn't correctly created.

This commit attempts to fix this bug, and so #25 changes can be
finally applied.


otoolep:
Thanks @fern4lvarez

PR #49, opened 2015-12-14
New API for Retrieval / Bulk Ingest

heckdevice:
Implemented the following two apis:

1. /db/{tablename} GET operation internally transforms to a Select * from {tablename} and returns the JSON

2. /db/bulk POST operation bulk ingests the data into tables. The payload has to be in the following format:
Taking e.g. of table foo with columns: id and name
{
"data":
[
 {
  "table":"foo",
  "payload":
  [{"name":"testdata6","id":"601"},
  {"name":"testdata7","id":"701"}]
 }
]
}

The same can be used to insert into multiple tables as well.

### DIFF
--- a/main.go
+++ b/main.go
@@ -23,7 +23,7 @@ import (
 	"strings"
 	"time"
 	
-	"github.com/jsrCorp/rqlite/log"
+	"github.com/otoolep/rqlite/log"
 	"github.com/jsrCorp/rqlite/server"
 )
 
--- a/main.go
+++ b/main.go
@@ -22,7 +22,7 @@ import (
 	"runtime/pprof"
 	"strings"
 	"time"
-
+	
 	"github.com/jsrCorp/rqlite/log"
 	"github.com/jsrCorp/rqlite/server"
 )
--- a/main.go
+++ b/main.go
@@ -24,7 +24,7 @@ import (
 	"time"
 	
 	"github.com/otoolep/rqlite/log"
-	"github.com/jsrCorp/rqlite/server"
+	"github.com/otoolep/rqlite/server"
 )
 
 var host string
--- a/main.go
+++ b/main.go
@@ -23,8 +23,8 @@ import (
 	"strings"
 	"time"
 
-	"github.com/otoolep/rqlite/log"
-	"github.com/otoolep/rqlite/server"
+	"github.com/jsrCorp/rqlite/log"
+	"github.com/jsrCorp/rqlite/server"
 )
 
 var host string
--- a/server/server.go
+++ b/server/server.go
@@ -313,6 +313,7 @@ func (s *Server) ListenAndServe(leader string) error {
 	s.router.HandleFunc("/diagnostics", s.serveDiagnostics).Methods("GET")
 	s.router.HandleFunc("/raft", s.serveRaftInfo).Methods("GET")
 	s.router.HandleFunc("/db", s.readHandler).Methods("GET")
+	s.router.HandleFunc("/db/{tablename}", s.TableReadHandler).Methods("GET")
 	s.router.HandleFunc("/db", s.writeHandler).Methods("POST")
 	s.router.HandleFunc("/join", s.joinHandler).Methods("POST")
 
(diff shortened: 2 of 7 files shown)


otoolep:
Thanks for the PR.

main.go
```diff
@@ -22,9 +22,9 @@ import (
 	"runtime/pprof"
 	"strings"
 	"time"
-
+	
 	"github.com/otoolep/rqlite/log"
-	"github.com/otoolep/rqlite/server"
+	"github.com/jsrCorp/rqlite/server"
```
otoolep: You need to revert this change.

server/server.go
```diff
@@ -313,7 +314,9 @@ func (s *Server) ListenAndServe(leader string) error {
 	s.router.HandleFunc("/diagnostics", s.serveDiagnostics).Methods("GET")
 	s.router.HandleFunc("/raft", s.serveRaftInfo).Methods("GET")
 	s.router.HandleFunc("/db", s.readHandler).Methods("GET")
+	s.router.HandleFunc("/db/{tablename}", s.tableReadHandler).Methods("GET")
 	s.router.HandleFunc("/db", s.writeHandler).Methods("POST")
+	s.router.HandleFunc("/db/ingest", s.tableWriteHandler).Methods("POST")
```
otoolep: How about using `/db/bulk` instead? Both `POST` endpoints allow data to be "ingested".

server/server.go
```diff
@@ -389,7 +392,7 @@ func (s *Server) joinHandler(w http.ResponseWriter, req *http.Request) {
 }
 
 func (s *Server) readHandler(w http.ResponseWriter, req *http.Request) {
-	log.Tracef("readHandler for URL: %s", req.URL)
+	log.Infof("readHandler for URL: %s", req.URL)
```
otoolep: Did you mean to change this?

server/server.go
```diff
@@ -436,7 +439,48 @@ func (s *Server) readHandler(w http.ResponseWriter, req *http.Request) {
...
 	}
 }
+func (s *Server) tableReadHandler(w http.ResponseWriter, req *http.Request) {
+	log.Infof("tableReadHandler for URL: %s", req.URL)
+	s.metrics.queryReceived.Inc(1)
+
+	var failures = make([]FailedSqlStmt, 0)
+	vars := mux.Vars(req)
+	tablename := vars["tablename"]
+	log.Infof("Getting data from table %s", tablename)
```
otoolep: I think this should be trace.

server/server.go
```diff
@@ -436,7 +439,48 @@ func (s *Server) readHandler(w http.ResponseWriter, req *http.Request) {
...
 
+	pretty, _ := isPretty(req)
+	var b []byte
+	if pretty {
+		b, err = json.MarshalIndent(rr, "", "    ")
+	} else {
+		b, err = json.Marshal(rr)
+	}
+	if err != nil {
+		log.Tracef("Failed to marshal JSON data: %s", err.Error())
```
otoolep: Why not log this at error lvel?

server/server.go
```diff
@@ -436,7 +439,48 @@ func (s *Server) readHandler(w http.ResponseWriter, req *http.Request) {
...
+		http.Error(w, err.Error(), http.StatusBadRequest) // Internal error actually
```
otoolep: It's more Go-style to return from this clause, and have no `else` clause. Can you make that change? (And in other places where it matter)?

otoolep:
I have feedback, but am up for merging this once some changes are made.

Once we're both happy, I will ask that you squash the commits into a single commit.

server/server.go
heckdevice: Thats a valid point. OK

server/server.go
heckdevice: Mistake. Will revert.

server/server.go
heckdevice: Korrect. Will do

server/server.go
heckdevice: Ok. my first go exp so... :)

PR #62, opened 2016-02-26
Support backup at the Store level

otoolep:


otoolep:
Part of the fix for https://github.com/otoolep/rqlite/issues/57

PR #63, opened 2016-02-26
Move to POST for bulk queries

otoolep:
Backup endpoint also added, but not fully-functional.


otoolep:
Fixes: https://github.com/otoolep/rqlite/issues/61

Partial solution for https://github.com/otoolep/rqlite/issues/57

PR #66, opened 2016-02-28
Update build script

prateek1306:
Hey, I was setting up rqlite and realised that I need to install raft and raft-boltdb to start the daemon.
Do you think it should be added to README.md? 

I am using OSX 10.11.3 and go1.4


otoolep:
You're right @prateek1306 -- the build instructions are incomplete. :-(

Thanks for your report. I actually fixed it slightly differently.

https://github.com/otoolep/rqlite/commit/21952567b145a06f45ce867cc4595faad83da665

otoolep:
@prateek1306 -- I made the build instructions even shorter. Let me know if they don't work for you.

prateek1306:
Thanks, this works :)

Issue #52, opened 2016-02-17
Descriptions of column data types for query results in JSON response would be useful


zmedico:
In order for a database API to map query results to specific types other than string, such as DATETIME, it would be useful to include the data types in the JSON response. Without this information, the database API implementation has to do a separate `PRAGMA table_info` query, and then parse the SQL statement to determine how those types map to result columns. It would greatly simplify database API implementations if the JSON response described the result column types.

otoolep:
Not committing to fixing this for v2.0, but might get to it.

zmedico:
The relevant info is stored in the sqlite3.SQLiteRows.decltype field (which is currently inaccessible outside the sqlite3 package).

zmedico:
I have [a branch](https://github.com/zmedico/rqlite/tree/decltype) containing a patch that changes DB.Query to bypass sql.DB, allowing for access to SQLiteRows instances. Now sqlite3.SQLiteRows just needs a method to export the declaration types.

zmedico:
Fixed by https://github.com/otoolep/rqlite/pull/71.

PR #71, opened 2016-03-06
Use SQLiteRows.DeclTypes() to fix issue #52

zmedico:
Merge this only after https://github.com/mattn/go-sqlite3/pull/282 has been merged.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -69,6 +69,7 @@ type Result struct {
 // Rows represents the outcome of an operation that returns query data.
 type Rows struct {
 	Columns []string        `json:"columns,omitempty"`
+	Types   []string        `json:"types,omitempty"`
 	Values  [][]interface{} `json:"values,omitempty"`
 	Error   string          `json:"error,omitempty"`
 	Time    float64         `json:"time,omitempty"`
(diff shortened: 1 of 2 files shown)


otoolep:
Cool, let's wait until that PR is merged and then I will take a closer look.

zmedico:
The go-sqlite3 PR has been merged now. However, circleci still fails because apparently it still uses an older copy of go-sqlite3.

otoolep:
Selected "rebuild without cache".

db/db.go
```diff
@@ -227,12 +231,19 @@ func (db *DB) Query(queries []string, tx, xTime bool) ([]*Rows, error) {
 			}
 		}()
 
+		if db.sqlite3conn == nil {
```
otoolep: I don't think this will work because another call to `Open` could take place and set a new value for `sqlite3conn`. That said, I'm going to fix this connection handling in another PR so the issue will soon be moot.

otoolep:
As I thought -- you need to update the unit tests too.

db/db.go
```diff
@@ -67,6 +69,7 @@ type Result struct {
 // Rows represents the outcome of an operation that returns query data.
 type Rows struct {
 	Columns []string        `json:"columns,omitempty"`
+	DeclTypes []string      `json:"decltypes,omitempty"`
```
otoolep: Let's pick a different name for this. `decltypes` is too low-level. How about just `type`?

zmedico: Renamed to just `types`, and fixed tests.

otoolep:
Thanks @zmedico

Issue #73, opened 2016-03-07
Move to sqlite3 driver directly


otoolep:
Started with https://github.com/otoolep/rqlite/issues/72

It probably makes most sense to use the SQLite3-specific code, and forget about using the generic driver.

otoolep:
Fixed by #74

PR #74, opened 2016-03-14
Use sqlite3.SQLiteDriver directly, fix issue #73

zmedico:


db/db.go
```diff
@@ -80,62 +69,45 @@ func Open(dbPath string) (*DB, error) {
 	openMu.Lock()
 	defer openMu.Unlock()
 
-	dbc, err := sql.Open("sqlite3_with_hook", dbPath)
+	os.MkdirAll(filepath.Dir(dbPath), 0755)
```
otoolep: The error should be checked here.

db/db.go
```diff
@@ -80,62 +69,45 @@ func Open(dbPath string) (*DB, error) {
...
+	d := sqlite3.SQLiteDriver{}
+	dbc, err := d.Open(dbPath)
+
```
otoolep: Nit: please remove this blank line.

db/db.go
```diff
@@ -17,21 +18,10 @@ import (
 // to the driver's internal conn object. The mutex ensures only 1 Open() call
 // can run concurrently, and when it has returned, sqlite3conn will be set to
 // the connection. This connection can be used for backups.
-var sqlite3conn *sqlite3.SQLiteConn
 var openMu sync.Mutex
```
otoolep: This mutex and all the comment above can be removed now, right?

db/db.go
otoolep: Also, I don't think this is necessary. The directory tree will be created, right?

db/db.go
```diff
@@ -80,62 +69,45 @@ func Open(dbPath string) (*DB, error) {
...
 // Execute executes queries that modify the database.
 func (db *DB) Execute(queries []string, tx, xTime bool) ([]*Result, error) {
 	type Execer interface {
-		Exec(query string, args ...interface{}) (sql.Result, error)
+		Exec(query string, args []driver.Value) (driver.Result, error)
 	}
 
 	var allResults []*Result
-	err := func() error {
+	err := func() (err error) {
```
otoolep: Why did you decide to name the error?

otoolep:
Thanks for doing this @zmedico -- it's much better. I do have some feedback before I can merge however.

otoolep:
I also started a CHANGELOG. Can you update it?

https://github.com/otoolep/rqlite/blob/master/CHANGELOG.md
 
Add the following line to the 2.0 section:

```
- [PR 74](https://github.com/otoolep/rqlite/pull/74): Use SQLite3 connection directory
```

db/db.go
zmedico: Apparently sql.Open creates the directory. I just tested again, and this is the error if we don't create the directory:
```2016/03/14 21:05:39 failed to open store: unable to open database file```

File: db/db.go (hunk 4 above)
Comment (zmedico) [inline db/db.go]: If I don't name the error, then I get this:
```
src/github.com/otoolep/rqlite/db/db.go:126: undefined: err
src/github.com/otoolep/rqlite/db/db.go:127: undefined: err
src/github.com/otoolep/rqlite/db/db.go:128: undefined: err
```
If use `:=` to try and fix the above problem, I get this:
```
src/github.com/otoolep/rqlite/db/db.go:126: t declared and not used
```
I'm open to suggestions if there is a better way to achieve the intended result.

File: db/db.go (hunk 1 above)
Comment (otoolep) [inline db/db.go]: OK, I see what you're saying -- my mistake.

The directory creation should be performed up a level, which I've done here: https://github.com/otoolep/rqlite/commit/e7e212411a4831fa3b24aa3a02b102ad162ac431

That way you don't need to worry about it now.

Comment (zmedico) [inline db/db.go]: Updated to check for MkdirAll error.

Comment (otoolep) [inline db/db.go]: Just to explain my reasoning. The directory is also needed by the Store, but the Store is more fundamental to the system than the SQLite file, hence have the Store create it.

File: db/db.go (hunk 2 above)
Comment (zmedico) [inline db/db.go]: Done.

File: db/db.go [hunk 5]
```diff
@@ -77,65 +57,48 @@ type Rows struct {
...
 		var execer Execer
 		var rollback bool
+		var t driver.Tx
```
otoolep: OK, for the sake of consistency, how about just doing `var err error` here?

db/db.go
zmedico: Yes, those are removed now.

db/db.go
```diff
@@ -77,65 +57,48 @@ type Rows struct {
 
 // Open an existing database, creating it if it does not exist.
 func Open(dbPath string) (*DB, error) {
-	openMu.Lock()
-	defer openMu.Unlock()
-
-	dbc, err := sql.Open("sqlite3_with_hook", dbPath)
+	err := os.MkdirAll(filepath.Dir(dbPath), 0755)
```
otoolep: Yeah, you can delete this now, since it's done at the `Store` code.

otoolep:
Thanks again. Minor feedback. If I don't see changes soon, I'll merge yours and make the changes.

db/db.go
zmedico: Done.

db/db.go
zmedico: Done.

otoolep:
Thanks @zmedico !

zmedico:
I didn't push the changelog entry before you merged it.

otoolep:
No worries @zmedico -- I updated the log for you.

PR #88, opened 2016-04-17
Re-enable multi-node testing on CirceCI

otoolep:


otoolep:
Fixes https://github.com/otoolep/rqlite/issues/64

PR #102, opened 2016-04-25
Add initial cluster service package

otoolep:


otoolep:
This package will interact with the Raft consensus module, and allow nodes in a cluster to set and get metadata about themselves, with the cluster, in a consistent manner. 

Part of the work for https://github.com/otoolep/rqlite/issues/67

PR #104, opened 2016-04-26
Don't join unless actually necessary

otoolep:


otoolep:
Fix https://github.com/otoolep/rqlite/issues/103

Issue #67, opened 2016-02-29
Add support for leader redirection


otoolep:
If a node receives a query which can only be served by the leader, and it is not the leader, it should support leader redirection (HTTP status 301).

adamierymenko:
:+1:

Would make this considerably more transparent.

Also wanted to thank you for this! We plan to use this a lot for things like this:

https://github.com/zerotier/ZeroTierOne/tree/master/controller

Using this we could cluster our stock network controller implementation (that uses SQLite3) without writing another to use some other kind of database.

We're also considering using this for other stuff. Got any place to send donations?

otoolep:
It's easy to do, I've started it here: https://github.com/otoolep/rqlite/pull/69

otoolep:
This is not actually as easy to do as initially thought, because it is not straightforward for a node to learn the query port of the leader. It's easy to find the Raft port, but that is not sufficient.

otoolep:
Until I get this in, you can learn the current leader by querying the status endpoint. The current leader will be listed in the section titled "raft".

otoolep:
One possible solution is: each node will have to send a message back down the Raft connection to the leader, with a special message in there telling the leader its HTTP address. On receipt of these messages the leader broadcasts the information (Raft address, HTTP address) to all other nodes. This could be done via a message through the Raft log -- a message that is not actually applied to the database. 

Alternatively, perhaps the leader broadcasts this out the Raft connection.

otoolep:
Basically, the key here is that each node should know how to map a Raft address of a node to the HTTP address of that node, so that HTTP redirection can be setup properly.

otoolep:
Just sending the HTTP(S) API address on join is not enough, as it could change on a node after a node is restarted. The follower node must actually ensure it is sent once to the leader after startup (and the leader broadcasts it out). Once sent, the sending can stop.

otoolep:
This is almost complete. The 307 is coming back from followers, but the returned address isn't quite right yet.
```
* Hostname was NOT found in DNS cache
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 4005 (#0)
> GET /db/query?pretty&timings&q=SELECT%20%2A%20FROM%20foo HTTP/1.1
> User-Agent: curl/7.38.0
> Host: localhost:4005
> Accept: */*
> 
< HTTP/1.1 307 Temporary Redirect
< Location: /db/
[...]

PR #111, opened 2016-04-30
Fill our API peers lookups

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/67

PR #131, opened 2016-05-06
db: only convert row value to string if the column type is text

### LINKED ISSUE #130: db.Query silently converts []byte to string
Here: https://github.com/rqlite/rqlite/blob/master/db/db.go#L283-L291

This means that any ``BLOB`` data types get silently returned as strings. I feel like the more appropriate solution would be to hex or base64-encode the value in the http service.

bkeroackdsc:
@otoolep Fixes #130. Tests pass.

Turns out that json.Marshal will base64-encode any []byte values it finds by default, so no changes to the http service are necessary. It's still an open question why go-sqlite is returning ``[]byte`` values for text columns.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -277,16 +277,20 @@ func (db *DB) Query(queries []string, tx, xTime bool) ([]*Rows, error) {
 					}
 					break
 				}
-
 				values := make([]interface{}, len(rows.Columns))
-
-				// Special case -- convert []uint8 to string. Perhaps this should be a config option.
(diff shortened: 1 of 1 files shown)


otoolep:
Thanks @bkeroackdsc

otoolep:
@bkeroackdsc - you wish to verify your e-mail with GitHub so you get credit as an rqlite contributor.

https://github.com/rqlite/rqlite/commit/3f064afc317a1776a209170126cb1d9b504d6911

PR #141, opened 2016-05-18
Store methods to expose node Raft state

bkeroack:
I'm using store as a library in one of my projects, and it's become necessary to introspect the current node's Raft state (currently that's only available by the ``Leader()`` method which requires string matching).

This PR adds an ``IsLeader()`` bool method as well as a ``State()`` method (wrapper around ``raft.State()``).

Tests pass. Let me know if there are any issues/modification you would like to see.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -91,6 +91,18 @@ const (
 	Strong
 )
 
+// ClusterState defines the possible Raft states the current node can be in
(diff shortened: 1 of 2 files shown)


otoolep:
Looks good thanks @bkeroack

Issue #166, opened 2016-06-25
Allow config to be set via environment variables


otoolep:
This was highlighted as part of the work for https://github.com/rqlite/rqlite/issues/163

This will allow rqlite running inside a Docker container to be configured dynamically.

otoolep:
On second thoughts, this might not be the best way to do this.

PR #167, opened 2016-06-25
Add new env package and env config variables

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/166

otoolep:
Not going anywhere with this right now.

PR #204, opened 2016-10-23
Use Go 1.7.1 on CircleCI

otoolep:


otoolep:
This doesn't appear to work. Probably need to some more here.

PR #221, opened 2016-11-19
Start handling the various SQLite TEXT types

otoolep:
This is the start of a code change to handle this comprehensively.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -7,6 +7,7 @@ import (
 	"expvar"
 	"fmt"
 	"io"
+	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


otoolep:
The code could check the encoding of the database, and perform the correct conversions.
PRAGMA encoding;
PRAGMA encoding = "UTF-8";
PRAGMA encoding = "UTF-16";
PRAGMA encoding = "UTF-16le";
PRAGMA encoding = "UTF-16be";

otoolep:
I'm going to merge this, as I think it's a step in the right direction.

PR #231, opened 2016-11-29
Add embedded single quote test

otoolep:
This fails. I need a more sophisticated parser, that tracks more state.


otoolep:
The SQLite parser works fine.
```
sqlite> SELECT * FROM "foo;'barx";
Error: no such table: foo;'barx
sqlite> 
```

otoolep:
Moot, after the simple loading of SQLite dump files.

PR #232, opened 2016-11-29
Add rqlite flexibility

tych0:
Hi,

I've been playing around with embedding rqlite into another application. Here's some patches that I found useful while experimenting.

### DIFF
--- a/cmd/rqlite/execute.go
+++ b/cmd/rqlite/execute.go
@@ -21,7 +21,7 @@ type executeResponse struct {
 }
 
 func execute(ctx *cli.Context, cmd, line string, argv *argT) error {
-	urlStr := fmt.Sprintf("%s://%s:%d/db/execute?pretty&timings", argv.Protocol, argv.Host, argv.Port)
+	urlStr := fmt.Sprintf("%s://%s:%d%sdb/execute?pretty&timings", argv.Protocol, argv.Host, argv.Port, argv.Prefix)
 	ret := &executeResponse{}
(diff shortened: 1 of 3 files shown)


otoolep:
Thanks very much for your contribution.

cmd/rqlite/execute.go
```diff
@@ -21,9 +21,9 @@ type executeResponse struct {
 }
 
 func execute(ctx *cli.Context, cmd, line string, argv *argT) error {
-	urlStr := fmt.Sprintf("%s://%s:%d/db/execute?pretty&timings", argv.Protocol, argv.Host, argv.Port)
+	urlStr := fmt.Sprintf("%s://%s:%d%sdb/execute?pretty&timings", argv.Protocol, argv.Host, argv.Port, argv.Prefix)
```
otoolep: What exactly is the point of this change? I don't follow what one would use this switch for.

tych0: Right now, the client expects the API to live at `/`. I'm embedding rqlite (i.e. not using rqlited, but using the go library) into another application at some level that's not the root, e.g. `/rqlite`. This just allows you to specify something besides `/`.

otoolep: Got it, thanks @tych0

otoolep:
Thanks @tych0 -- I might tweak the help output slightly, but this does seem useful, particularly the TLS support.

tych0:
No problem, thanks for the quick response!

PR #236, opened 2016-12-05
Merge pull request #2 from rqlite/master

jmptrader:
update from original


otoolep:
I think this is a mistake. You need to generate a PR against your own repo.

PR #239, opened 2016-12-15
add an API for using custom raft.PeerStore/log.Logger

tych0:
This is useful in case the server needs to store other metadata (e.g. auth
data) along side the peer list.

The logging bit is handy in case something has its own logging framework
that it wants to use.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/cmd/rqlited/main.go
+++ b/cmd/rqlited/main.go
@@ -167,7 +167,11 @@ func main() {
 	}
 	dbConf := store.NewDBConfig(dsn, !onDisk)
 
-	store := store.New(dbConf, dataPath, raftTn)
+	store := store.New(&store.StoreConfig{
+		DBConf: dbConf,
+		Dir:    dataPath,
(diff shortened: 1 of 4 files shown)


tych0:
Seems I forgot a go fmt. Let me know if you want me to squash these together or leave them separate.

otoolep:
Yeah, if you squash I will then merge -- thanks!

https://github.com/rqlite/rqlite/blob/master/CONTRIBUTING.md#clean-commit-histories

tych0:
On Thu, Dec 15, 2016 at 04:01:32PM -0800, otoolep wrote:
> Yeah, if you squash I will then merge -- thanks!
> 
> https://github.com/rqlite/rqlite/blob/master/CONTRIBUTING.md#clean-commit-histories

Ok; I've pushed a squashed version. Thanks!

PR #242, opened 2016-12-22
add a raw snapshotting API

tych0:
This allows us to grab a raw snapshot of the databse.

Signed-off-by: Tycho Andersen <tycho.andersen@canonical.com>

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -683,13 +683,17 @@ func (s *Store) Apply(l *raft.Log) interface{} {
 	}
 }
 
-// Snapshot returns a snapshot of the database. The caller must ensure that
-// no transaction is taking place during this call. Hashicorp Raft guarantees
-// that this function will not be called concurrently with Apply.
+// Database() returns a copy of the underlying database. The caller should
+// ensure that no transaction is taking place during this call, or an error may
+// be returned.
 //
 // http://sqlite.org/howtocorrupt.html states it is safe to do this
 // as long as no transaction is in progress.
-func (s *Store) Snapshot() (raft.FSMSnapshot, error) {
+func (s *Store) Database(leader bool) ([]byte, error) {
+	if leader && s.raft.State() != raft.Leader {
+		return nil, ErrNotLeader
+	}
+
 	// Ensure only one snapshot can take place at once, and block all queries.
 	s.mu.Lock()
 	defer s.mu.Unlock()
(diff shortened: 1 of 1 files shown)


store/store.go
```diff
@@ -683,13 +683,13 @@ func (s *Store) Apply(l *raft.Log) interface{} {
...
 
-// Snapshot returns a snapshot of the database. The caller must ensure that
+// RawDBSnapshot returns a snapshot of the database. The caller must ensure that
 // no transaction is taking place during this call. Hashicorp Raft guarantees
 // that this function will not be called concurrently with Apply.
 //
 // http://sqlite.org/howtocorrupt.html states it is safe to do this
 // as long as no transaction is in progress.
-func (s *Store) Snapshot() (raft.FSMSnapshot, error) {
+func (s *Store) RawDBSnapshot() ([]byte, error) {
```
otoolep: I think I prefer a simpler name for this call. How about just `Database()`?

store/store.go
```diff
@@ -683,13 +683,13 @@ func (s *Store) Apply(l *raft.Log) interface{} {
 	}
 }
 
-// Snapshot returns a snapshot of the database. The caller must ensure that
+// RawDBSnapshot returns a snapshot of the database. The caller must ensure that
 // no transaction is taking place during this call. Hashicorp Raft guarantees
```
otoolep: The docstring for this function shouldn't contain any reference to Hashicorp Raft, it's not relevant in this context. This is function that is handing control over to the user, and they better be sure they know what they are doing.

There are multiple potential ways to shoot yourself in the foot here. There is no check that the store is the leader of the cluster, for example, and therefore no guarantee that the copy you get is up-to-date.

tych0:
The idea here is to offer introspection of the node's view of the database. This is something we've found useful in the past when debugging issues with LXD's database, and so we want some way to export each node's database.

store/store.go
tych0: Right, the lack of a leader check is intentional. We want the current state of the node's database as-is.

otoolep:
OK, I think it's fair to provide an option to get to the raw database underneath. It does no real harm.

However, can you add a bool flag to the function signature, like so:

 `func (s *Store) Database() ([]byte, bool leader, error)`

and add the check that you find in `Backup()`?  on the `Store` object.  I would feel better if this function explicitly forced the user to choose whether he wants a leader check or not, before returning the data of the underlying SQLite database.

When calling `Database()` from within the Snapshot function, you can set leader check to `false`, since I assume the Raft infra knows what it is doing.

What do you think?

tych0:
Assuming you mean `func (s *Store) Database(leader bool) ([]byte, error)` then yep, sounds good and I just pushed a fix.

otoolep:
Sorry, yeah @tych0 -- that's what I mean, thanks.

store/store.go
```diff
@@ -683,13 +683,16 @@ func (s *Store) Apply(l *raft.Log) interface{} {
 	}
 }
 
-// Snapshot returns a snapshot of the database. The caller must ensure that
-// no transaction is taking place during this call. Hashicorp Raft guarantees
-// that this function will not be called concurrently with Apply.
+// RawDBSnapshot returns a snapshot of the database. The caller must ensure that
```
otoolep: Docstring isn't quite right. Can you change it to read:

```
Database() returns a copy of the underlying database. The caller should ensure that no transaction is taking place during this call, or an error may be returned.
```

Then I'll merge.

tych0:
Yup, just pushed a fix.

otoolep:
Thanks @tych0

tych0:
No problem, thank you!

Issue #241, opened 2016-12-22
better way of extracting results?


tych0:
Hi,

I was using some code like this:

```
for _, r := range rows.Values {
    n := make([]byte, len(r[0].([]byte)))
    copy(n, r[0].([]byte))
    real := string(n)
}
```

to extract sql results. 75417f6d1e78dc5e9556b7732fbefbbe1db734cf breaks this, but ends up making it substantially similar (`r[0].(string)` is all that's needed now).

My question is: is there some better way I can extract results with less brittle code so future changes like this work? Maybe that's a pipe dream, but I figured I'd ask :)

otoolep:
This line:
`copy(n, r[0].([]byte))`
of your previous code was already making an assumption that the value was of type `[]byte`. If you were storing integers, for example, in the column, your code would have panicked.

The type for the values in question is `interface{}` which tells you nothing really. It would be nice to tighter, I am open to suggestions.

But the fact that it is `interface{}` already leaves the interface quite loose.

tych0:
Right; the problem is that the commit above changes the return type of strings from []byte to string, so the code does panic. I'm just wondering if there's a cleaner way. I don't have any bright ideas right now, but I'll think about it :)

otoolep:
OK, perhaps the new code should be more explicit about the contract of this function, as put in place by the recent change.

However, my point about your code (correct me if I am missing something) is that if you stored, say, integers in a column and attempted to type-assert them to `[]bytes`, that too would have panicked, even with the previous version of rqlite code.

tych0:
Yes, the cast will fail if I try to cast a column to a type it's not. It's my job to cast it to the right thing. The problem is when the right thing changes underneath me :)

otoolep:
Right, I see. Let me allow you to control this.

otoolep:
OK, now that I look at it, this totally changed the contract. Bad on my part. Let me tweak the store now, so you have control over whether this conversion takes place. I do think the conversion is useful, but perhaps not to people who just care about the store package.

I'l push up a PR, and you can tell me what you think.

otoolep:
Actually @tych0 -- can you tell me something else? Is the conversion to `string` when the underlying type is of TEXT affinity useful? If so, and it was an explicit part of the function's contract, would that be OK? I'm trying to determine if this conversion is actually useful at this layer.

So is the change in contract, or the conversion itself, that is awkward?

otoolep:
Obsolete.

PR #244, opened 2016-12-23
Control text conversion

otoolep:
This change is a proposed solution to locking down behaviour of the `Store` package, to help with https://github.com/rqlite/rqlite/issues/241

This gives users of the `Store` package a new call, which doesn't perform any conversion of values which have TEXT affinity (see http://www.sqlite.org/datatype3.html) to strings. 

This functionality has not been exposed beyond the `Store` layer for now. It could be, and controlled via a HTTP URL param to rqlited clients.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -50,8 +50,6 @@ type DB struct {
 	path        string              // Path to database file.
 	dsn         string              // DSN, if any.
 	memory      bool                // In-memory only.
-
-	TextDisable bool // Set to disable TEXT affinity []byte to string
 }
 
 // Result represents the outcome of an operation that changes rows.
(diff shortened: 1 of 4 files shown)


otoolep:
@tych0 -- with this PR you've got a new call on the `Store` type, `QueryRaw`. This acts the way `Query` did previously.

As far as I can see, the automatic conversion of TEXT affinity values to strings is useful to people, so I want to keep that they way it is. But `QueryRaw` might be better for you. Let me know what you think.

tych0:
Hey, just saw this. I actually don't mind the new way -- as I mentioned in the other issue it makes my code much simpler. My question was just if we can find a way to avoid this breakage entirely, e.g. by doing what the stdlib's sql.Rows.Scan() does, or something similar. i.e. have the user pass in types, and then shoehorn the values in to that type.

otoolep:
OK, I see what you mean @tych0 -- there are some ideas there.

I need to think about this for a short while. This is the first grey area, where the goals of the `Store` type are not quite in line with the level of control you seek. I want this to work however, and see how it _might_ work. There is no requirement that all methods added to the `Store` type have to be exposed by the HTTP service, so this does provide some flexibility around what it (`Store`) exposes.

Presumably you're not blocked right now, and I won't change the behavior of `Query` again so you can continue to build on it.

tych0:
Yep, I'm not blocked on it. Also, please feel free to change Query if it makes sense (I think it did in this string case). I can adapt my code easily enough. Mostly I'm just curious if we can make it so that I don't have to at some point in the future :)

PR #247, opened 2016-12-27
Use single load for dump processing

otoolep:
Writing a sufficiently sophisicated parser is too much work, and unlikely to be successful soon. Instead this change simply loads the entire dump as one command, and allows the underlying SQLite support to parse it correctly. This will definitely work, but since the load goes over Raft, it may hit limits with regards to network transfer sizes. Right now any limitations in that area are unknown.

Therefore this functionality remains somewhat experimental.


kikitux:
@otoolep for the message limit. 
whats the raft function you are using? is this?

https://github.com/rqlite/rqlite/blob/e6586f41f2350a2e3516eccb0f2fdc7db33a495e/store/store.go#L637-L672

I can ask to the HashiCorp people for the message length limit

otoolep:
Yes @kikitux -- just speculating that there might be an internal Raft limit. I have not examined the code, nor run any tests, to know if this is the case however.

Issue #255, opened 2017-01-13
Common Table Expression support


alanjds:
I am updating [pyrqlite](https://github.com/rqlite/pyrqlite) to complain with the official sqlite driver of Python as much as possible. 

Most of the work done is on my fork on branch [test-types]( https://github.com/alanjds/pyrqlite/tree/tests-types) for now.

However, one test ported from the official module suggests that RQLite does not support a feature so called "Common Table Expression". The code that tests this support is here: https://github.com/alanjds/pyrqlite/blob/tests-types/src/test/test_types.py#L333

I am running the tests with DEBUG loglevel and, before the test fail, the SELECT is issued with success and but I saw no row nor columns info returned. It suggests me that, as the underlying SQLite version does support the feature, RQLite had not implemented it.

Could someone please clarify, as I am not "that" used to SQLite nor RQLite internals?

Thanks.

otoolep:
Looking into this now, I too am seeing some unexpected behaviour.

otoolep:
```
$ curl -v -XGET 'localhost:4001/db/query?pretty&timings' -H 'Content-Type: application/json' -d '[
    "with bar as (select * from test) select * from test where x = 1"
]'
*   Trying ::1...
* connect to ::1 port 4001 failed: Connection refused
*   Trying 127.0.0.1...
* Connected to localhost (127.0.0.1) port 4001 (#0)
> GET /db/query?pretty&timings HTTP/1.1
> Host: localhost:4001
> User-Agent: curl/7.43.0
> Accept: */*
> Content-Type: application/json
> Content-Length: 73
> 
* upload completely sent off: 73 out of 73 bytes
< HTTP/1.1 400 Bad Request
< Content-Type: application/json; charset=utf-8
< X-Rqlite-Version: 3
< Date: Fri, 13 Jan 2017 22:19:25 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```
Don't know why this is a bad request.

otoolep:
It's not an issue with the database or SQL layer, since this passes:

https://github.com/rqlite/rqlite/pull/256

Continuing to investigate.

otoolep:
Actually, I take it back, my `curl` request is wrong. rqlite supports this kind of thing fine:

```
$ curl -G 'localhost:4001/db/query?pretty&timings' --data-urlencode 'q=with bar as (select * from test) select * from test where x = 1'
{
    "results": [
        {
            "columns": [
                "x"
            ],
[...]

PR #258, opened 2017-01-14
Not all queries begin with SELECT so pull check

otoolep:
Common Table Expressions can begin with WITH. This check was always dubious.


otoolep:
Fixes https://github.com/rqlite/rqlite/issues/255

Issue #260, opened 2017-01-16
rqlited does not load


omerkorenn:
Hi, 
I preloaded a single node of rqlited (created a 1.5gb rafb.db file).
After gracefully shutdown of the process i tried to reload it, but the process ends with "set peer timeout expired" before all the data is being loaded to memory.

Please advise,

Thanks !!!

See the log here:
sudo ./rqlited --http 0.0.0.0:4001 ./node1

            _ _ _           _
           | (_) |         | |
  _ __ __ _| |_| |_ ___  __| |
 | '__/ _  | | | __/ _ \/ _  |  The lightweight, distributed
 | | | (_| | | | ||  __/ (_| |  relational database.
 |_|  \__, |_|_|\__\___|\__,_|
         | |
         |_|

[rqlited] 2017/01/16 16:22:49 rqlited starting, version v3.9.2, commit 46589e91248e10043bb4fe4b8c9522564c35b838, branch master
[store] 2017/01/16 16:22:49 SQLite in-memory database opened
[store] 2017/01/16 16:22:49 enabling single-node mode
[tcp] 2017/01/16 16:22:49 mux serving on 127.0.0.1:4002, advertising 127.0.0.1:4002
[cluster] 2017/01/16 16:22:49 service listening on 127.0.0.1:4002
2017/01/16 16:22:49 [INFO] raft: Node at 127.0.0.1:4002 [Follower] entering Follower state (Leader: "")
[rqlited] 2017/01/16 16:22:50 failed to set peer for 127.0.0.1:4002 to 0.0.0.0:4001: no leader available (retrying)
2017/01/16 16:22:50 [WARN] raft: Heartbeat timeout from "" reached, starting election
2017/01/16 16:22:50 [INFO] raft: Node at 127.0.0.1:4002 [Candidate] entering Candidate state
2017/01/16 16:22:50 [DEBUG] raft: Votes needed: 1
2017/01/16 16:22:50 [DEBUG] raft: Vote granted from 127.0.0.1:4002. Tally: 1
2017/01/16 16:22:50 [INFO] raft: Election won. Tally: 1
2017/01/16 16:22:50 [INFO] raft: Node at 127.0.0.1:4002 [Leader] entering Leader state
2017/01/16 16:22:50 [DEBUG] raft: Node 127.0.0.1:4002 updated peer set (2): [127.0.0.1:4002]
[...]

PR #261, opened 2017-01-18
Allow Raft Apply timeout to be configurable

otoolep:


otoolep:
May be a solution for https://github.com/rqlite/rqlite/issues/260

Issue #266, opened 2017-01-24
Transactions?


tych0:
Hi,

I'm interested in some sort of transaction functionality beyond what rqlite currently provides today. In particular, I'd like to be able to get the results of a query in the middle of a transaction, as opposed to just "process this set of queries", the primitive that rqlite provides today.

We could do transactions as the leader, applying them to the sqlite database and accumulating the queries to apply until the transaction was committed, but I don't think that would actually guarantee correctness (someone else who isn't in the transaction could take e.g. a primary key ID that was allocated in the transaction, since we're not committing things to the raft logs as they're added to the tx).

Do you have any ideas on how we'd go about doing this, or if it's even possible?

Thanks

tych0:
Actually, after a careful reading of:

https://www.sqlite.org/lang_transaction.html
https://www.sqlite.org/lockingv3.html

I think that this is possible. It would look something like this:

1. The leader does a BEGIN and returns a tx id.
2. That transaction accumulates queries (by using golang sql's tx.Execute/tx.Query methods to return results)
3. when done, the leader acquires a lock, commits the transaction to the sqlite locally, and then also commits the transaction to the raft logs. Once the transaction is committed to the raft logs, the lock is released.

I think this is safe because active transactions will never overlap with each other, so the potential race here is that things are committed in the wrong order. Since we only ever allow the leader to do transactions, and the leader synchronizes transactions in the order they're committed, things should be safe.

I'll see if I can come with an implementation.

tych0:
FWIW, there's an implementation of this here: https://github.com/tych0/rqlite/tree/transaction

The queries don't quite work, but I think this is likely just some mechanics about converting values from query results to actual usable values. I'll keep fiddling with that, but the transaction approach is there (and it is also quite small, all things considered).

otoolep:
Interesting! I think it's safe, but there might be one problem.

I do agree that if the leader only commits to the Raft log after successfully committing the transaction, it's pretty safe. Since, on start-up, the SQLite database is completely rebuilt from the Raft log any commits in the database, but not in the log, will not be applied to the newly-built database under the node.

However, there is a concept of snapshotting the SQLite database, as a part of compacting the log (and this state is shipped to other nodes that join). I'm not sure what guarantees there are around when that snapshot is taken. If it was taken while the commits were in the database, but before the log was applied to the database, it would be trouble.

[...]

PR #267, opened 2017-01-27
[RFC] Implementation of transactions

tych0:
This is an implementation of transactions as described in #266. There are a couple caveats:

* Two tests don't pass right now, although I believe this is a bug in the sqlite3 driver; I'm working on it.
* We're doing a little bit of voodoo with in memory databases. This doesn't seem very nice; I think we should probably just put all the DBs on disk, so easier to figure out what's going on. The DBs probably aren't ever going to be very big since we're coordinating everything over raft, so I think this would be an acceptable compromise.

Anyway, play around with it and let me know if you have any thoughts.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -3,6 +3,7 @@
 package db
 
 import (
+	"database/sql"
 	"database/sql/driver"
 	"expvar"
 	"fmt"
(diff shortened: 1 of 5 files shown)


tych0:
Actually, ignore this for now. I think all the stuff with memory needlessly complicates this; let me play around with it some more.

PR #279, opened 2017-02-13
Transaction testing

tych0:
Hi,

Here's a proposal for an implementation of transactions (#266). Let me know what you think; all the tests seem to pass, although the same caveat applies as the last PR (#267), that I think in-memory databases in SQLite + golang are a little bit wonky, and it's probably best just to get rid of the in-memory concept entirely.

Please let me know if you have any thoughts!

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -140,6 +140,50 @@ func (db *DB) Close() error {
 	return db.sqlite3conn.Close()
 }
 
+// Destroy drops all the tables from the database. This is useful for resetting
+// an in-memory database when necessary.
+func (db *DB) Destroy() error {
+	if !db.memory {
+		return fmt.Errorf("destroy not supported on non-memory databases")
+	}
+
+	/* Since we use the same in memory database (i.e. cache=shared)
+	 * to avoid oddities with how the stdlib's sqlite driver works,
+	 * we need to clean up after ourselves, since things will
+	 * persist.
+	 */
+	rows, err := db.sqlite3conn.Query("SELECT name FROM sqlite_master WHERE type='table'", nil)
+	if err != nil {
+		return err
+	}
+	defer rows.Close()
+
+	tables := []string{}
+	for {
+		vs := make([]driver.Value, 1)
+		err := rows.Next(vs)
+		if err != nil {
+			if err != io.EOF {
+				return err
+			}
+			break
+		}
+
+		switch val := vs[0].(type) {
+		case []byte:
+			tables = append(tables, string(val))
+		default:
+			return fmt.Errorf("unexpected type: %v", vs[0])
+		}
+	}
+
+	for _, t := range tables {
+		db.sqlite3conn.Exec(fmt.Sprintf("DROP TABLE %s", t), nil)
+	}
+
+	return nil
+}
+
 func open(dbPath string) (*DB, error) {
 	d := sqlite3.SQLiteDriver{}
 	dbc, err := d.Open(dbPath)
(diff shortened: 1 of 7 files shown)


otoolep:
OK, let me take a look, but before I do, why is there an issue with in-memory databases? It's just a SQLite database as far as rqlite is concerned. Maybe it's obvious from your PR.

In-memory databases can provide a big speed-up, I'm hesitant to remove it without a big, big win in return.

db/db.go
```diff
@@ -134,6 +140,50 @@ func (db *DB) Close() error {
 	return db.sqlite3conn.Close()
 }
 
+// Destroy drops all the tables from the database. This is useful for resetting
+// an in-memory database when necessary.
+func (db *DB) Destroy() error {
```
otoolep: Is there no way to do this by re-opening the database connection? Such that you get a new lump of RAM?

sql/sql.go
```diff
@@ -0,0 +1,250 @@
+// Package sql provides an way to create sql.DB objects out of an RQLite store.
+//
+// The goal of this package is to provide an interface that looks and feels
```
otoolep: I like this idea in principle. It's probably the interface I should have had the `Store` object implement in the first place. (Though how to do transactions is what you're trying to solve here0.

sql/sql.go
```diff
@@ -0,0 +1,250 @@
...
+	"github.com/rqlite/rqlite/store"
+)
+
+// RqliteSQLWrapper is the user-supplied interface that allows the driver to
+// resolve the rqlite store and communicate with its members.
+type RqliteSQLWrapper interface {
+	// Store returns the local store that this wrapper wraps.
+	Store() (*store.Store, error)
+
+	// ExecOn should execute the specified query on the specified address
```
otoolep: What is the idea behind these `On` functions?

db/db.go
```diff
@@ -81,12 +81,18 @@ func OpenWithDSN(dbPath, dsn string) (*DB, error) {
 
 // OpenInMemory opens an in-memory database.
 func OpenInMemory() (*DB, error) {
-	return open(fqdsn(":memory:", ""))
+	return OpenInMemoryWithDSN("mode=memory&cache=shared")
```
tych0: See the comments in the commit message: https://github.com/rqlite/rqlite/pull/279/commits/9c670e6b9e747adf3ff992d86374ff6b6f678db7

db/db.go
tych0: The problem is that you don't necessarily know when you want to do that. The stdlib's connection pooling mechanism may open or close connections at will, and you don't want to delete the entire database when one connection in the pool is closed, because the others are still valid.

sql/sql.go
tych0: If the current store is not the leader, you need some way to talk to the leader to actually execute the query; this is the way to provide that. This way, on every node that opens one of these, every query succeeds, instead of sometimes failing with ErrNotLeader and leaving you to handle it.

I can draft an implementation that's rqlite specific if you like to demo it.

tych0:
The problem with memory databases is something like the following: when users call `sql.Open()`, the stdlib creates a `sql.DB` object wrapping sqlite3, but it doesn't open any connections. If the database URL is just specified as ":memory:", then each call to `database/sql/driver.Conn.Open` gives a /unique/ in-memory database. Since the stdlib thinks it can reconnect to the same database and add another connection to its pool by `database/sql/driver.Conn.Open`ing again with the same database URL, things blow up (e.g. the table doesn't exist, because it was created on table 1, but inserted to on table 2).

See my comment here https://github.com/mattn/go-sqlite3/issues/204#issuecomment-275527228 on the issue linked from the commit message that has a program that demonstrates this behavior. Switch the commented out lines to break/unbreak things.

Issue #295, opened 2017-03-17
Problems quering from other  nodes


anfho93:
Hi.  I'm having problems with the cluster connection, im following the steps in the documentation, and when I connect directly to the leader through rqlite cmd and execute queries everything is ok, but when i do the same thing from other nodes, the answer to all the queries is:
 ERR! unexpected end of JSON input

I thinks this is related to the recent changes made. because i use curl to make a query in a non-leader node and this is what i get:

curl -G 'localhost:4003/db/query?pretty&timings' --data-urlencode 'q=.tables'
<a href="http://localhost:4001/db/query">Moved Permanently</a>.

btw, this is my configuration :

 rqlited ./node.1
 rqlited -http localhost:4003 -raft localhost:4004 -join http://localhost:4001 ./node.2
 rqlited -http localhost:4005 -raft localhost:4006 -join http://localhost:4001 ./node.3

**This is my first step on the leader node**

$ rqlite
127.0.0.1:4001> CREATE TABLE foo (id INTEGER NOT NULL PRIMARY KEY, name TEXT)
0 row affected (0.000668 sec)
127.0.0.1:4001> .schema
+-----------------------------------------------------------------------------+
| sql                                                                         |
+-----------------------------------------------------------------------------+
| CREATE TABLE foo (id INTEGER NOT NULL PRIMARY KEY, name TEXT)               |
+-----------------------------------------------------------------------------+
127.0.0.1:4001> INSERT INTO foo(name) VALUES("fiona")
1 row affected (0.000080 sec)
127.0.0.1:4001> SELECT * FROM foo
+----+-------+
| id | name  |
+----+-------+
| 1  | fiona |
+----+-------+

When i try to do the same on a non-leader node, this is the answer 

$ rqlite -H localhost -p 4003
localhost:4003> .schema
ERR! unexpected end of JSON input

I tried to make a curl request and this is what i get
$ curl -G 'localhost:4003/db/query?pretty&timings' --data-urlencode 'q=select * from foo'
<a href="http://localhost:4001/db/query">Moved Permanently</a>.

**This behavior was not present on the previous releases.** I hope you can help me with this.

otoolep:
Thanks for your report.
```
curl -G 'localhost:4003/db/query?pretty&timings' --data-urlencode 'q=.tables'
Moved Permanently.
```
[...]

PR #297, opened 2017-03-17
CLI should handle HTTP redirect explicitly

otoolep:
HTTP client behavior changed with Go 1.8. This change makes behavior that was implicit in Go 1.7 and earlier, explicit now.


otoolep:
Should fix https://github.com/rqlite/rqlite/issues/295

PR #337, opened 2017-08-07
Include any query params with 301 redirect URL

otoolep:
With this change, any URL set as `Location` in the HTTP headers includes the full request path *including* the query params.


otoolep:
@zmedico @justinmorant @jack-y -- I think I should just merge this change. Every client library I've seen would be compatible with this since those libraries are either a) parsing the host and port from the `Location` header, or just assuming the `Location` header is of the form this PR would make it.

In other words this might be a breaking change in principle, but not in practise. And it is bug.

otoolep:
OK, I'm going to merge this and release a new build, for the reasons outlined above.

PR #338, opened 2017-08-08
Support query params for POST

otoolep:


http/service.go
```diff
@@ -555,17 +555,26 @@ func (s *Service) handleExecute(w http.ResponseWriter, r *http.Request) {
...
+	// Prep storage for queries.
 	queries := []string{}
-	if err := json.Unmarshal(b, &queries); err != nil {
-		http.Error(w, err.Error(), http.StatusBadRequest)
+
+	if query, err := stmtParam(r); err != nil {
+		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
+	} else if query != "" {
+		queries = append(queries, query)
```
zmedico: This means that queries in the body will be ignored. Maybe return an error code if there are queries in both places?

otoolep: I was thinking about this. I do want URL params to take precedence over any in the body and making that explicit in the documentation.

Returning an error wouldn't be a bad idea, though the query endpoint doesn't do that in that case. 400 Bad Request would probably make most sense.

otoolep:
No plans to implement this.

PR #365, opened 2017-10-27
gRPC query and execute interface

otoolep:
This is the start of a gRPC interface, specifically for executing queries (which read data) and execute statements (which write data). All other functionality (backups, status, diagnostics) will remain on the HTTP interface. The goal here is to provide a lower-latency connection for reads and writes. How much lower is yet to be measured.

Feedback is welcome. Once this is finalized, libraries will be generated for various languages.

https://github.com/rqlite/rqlite/issues/350


otoolep:
@turbo -- I am benchmarking the performance of this new interface, and the HTTP API is just as fast. I still plan to proceed with this work, and there are other optimizations that gRPC will bring, but I wonder if it's that significant relevant to the network round trips between the nodes.

One thing is important to get good performance out of the HTTP API. You should use persistent HTTP connections. Are you doing so?

turbo:
Yes, I am. I will try to benchmark this in my own environment, too.

otoolep:
Perhaps, but it's not high-priority. Subsequent testing showed that the performance bottleneck is in the Raft subsystem, not the HTTP interface, and trying to improve performance was an important goal of this work.

I think the only way I'll get significantly better performance is via a different Raft implementation -- most likely the one from CoreOS (which is used by etc).

otoolep:
Closing for now.

bruth:
Understood thanks. That said I would argue the generated clients is still a nice benefit in lieu of performance gains.

Issue #375, opened 2017-10-31
Add new read-consistency level "stale"


otoolep:
There is a place for a 4th consistency level, call it _stale_.

This consistency level would be one step above [none](https://github.com/rqlite/rqlite/blob/master/DOC/CONSISTENCY.md) in that it would not check if it was leader, but it would check that it was part of the cluster within the last x milliseconds.

otoolep:
Fixed by https://github.com/rqlite/rqlite/pull/614. Should be in v5.1.

PR #376, opened 2017-11-01
Add stale read support

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/375

PR #377, opened 2017-11-02
Move to Hashicorp Raft v1

otoolep:


otoolep:
This is still work-in-progress. https://github.com/otoolep/hraftd/pull/15 is my guide.

store/store.go
```diff
@@ -273,10 +223,23 @@ func (s *Store) Open(enableSingle bool) error {
 	}
 
 	// Instantiate the Raft system.
-	ra, err := raft.NewRaft(config, s, logStore, logStore, snapshots, s.peerStore, transport)
+	ra, err := raft.NewRaft(config, s, logStore, logStore, snapshots, transport)
 	if err != nil {
 		return fmt.Errorf("new raft: %s", err)
 	}
+
+	if enableSingle {
```
otoolep: This no longer seems correct, and needs work. What is the right way to form the cluster?

otoolep: This seems relevant: https://www.consul.io/docs/guides/bootstrapping.html

PR #397, opened 2018-03-16
auth: Added cmdline option to specify passwords are bcrypted

### LINKED ISSUE #395: Feature: Hashed password in authFIle
Currently  `CredentialStore` expects passwords to be stored in cleartext in the `authFile`.

This is not really a problem as the long the access permissions to `authFile` are clean and controlled.

However, hashed password are always a better idea than cleartext. 
If that is acceptable, I would like to submit a patch for the same.

And Thank you for rqlite.

sum12:
`bcryptpass` cmdline informs the credentialStore that received passwords
should be hashed before comparing to the passwords obtained form the
authfile.

Closes #395

### DIFF
--- a/auth/credential_store.go
+++ b/auth/credential_store.go
@@ -5,6 +5,8 @@ package auth
 import (
 	"encoding/json"
 	"io"
+
+	"golang.org/x/crypto/bcrypt"
 )
 
 // BasicAuther is the interface an object must support to return basic auth information.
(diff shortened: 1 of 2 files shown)


otoolep:
Thanks for your PR. The functionality looks useful. However I think it's more complicated than it needs to be. There are two ways I think it could be simpler.

1) Instead of passing `bcrypt-pass` to rqlited when it starts up, just try the existing logic. If that doesn't work, then try the bcrypt hash check. Only if both fail is the access blocked.
2) Add a new field to the Passwords file, named, say, `password_bcrypt`. If this field exists check that, otherwise do the existing check. This change would be backwards compatible with existing files. Again you wouldn't need to pass `bcrypt-pass` to the binary.

This minimize the changes and means the way rqlited is launched is not changed.

sum12:
@otoolep thanks for the input.
I have update the PR to the second option. Let me know what you think.

otoolep:
Thanks @sum12, implementation generally looks good to me. -- I will merge this, and make a couple of changes on top of it.

PR #401, opened 2018-03-21
Make sure file is correctly closed and catch closing error

### LINKED ISSUE #400: File handle not closed cleanly
https://github.com/rqlite/rqlite/blob/77e345b97c5597c1ef86e75e690539de369b8dd3/db/db.go#L392 this should be deferred

sum12:
there could be errors during backup in which case the file handle is left
open. Also there could be errors in closing file, catch and return that
too unless there have been prior errors.
Both the cases now stand handled.

Closes #400

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -368,6 +368,13 @@ func (db *DB) Backup(path string) error {
 		return err
 	}
 
+	defer func(db *DB, err *error) {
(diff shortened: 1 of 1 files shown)


otoolep:
Thanks @sum12

Issue #369, opened 2017-10-27
CLI should accept username and password


otoolep:
The CLI doesn't current support adding Basic Auth information to the HTTP requests it makes, meaning it can't be used with a secure cluster. Ideally it would take something like the following as an option:

```
-u, --user username:password
```

and then set add Basic Auth credentials when making all requests.

wuxiaoxiaoshen:
How are you getting along with the work?

otoolep:
It hasn't started. It's an easy issue, which I plan to leave to the community.

PR #406, opened 2018-04-11
Add Basic Auth credentials support to the CLI

joaodrp:
Currently the CLI doesn’t support Basic Auth credentials. This change adds a new option, `-u, —user` to accept an username and password in the form of `username:password`. Credentials are then used to set the Authorization header for all HTTP requests.

Resolves: #369

### DIFF
--- a/cmd/rqlite/main.go
+++ b/cmd/rqlite/main.go
@@ -17,11 +17,12 @@ const maxRedirect = 21
 
 type argT struct {
 	cli.Helper
-	Protocol string `cli:"s,scheme" usage:"protocol scheme (http or https)" dft:"http"`
-	Host     string `cli:"H,host" usage:"rqlited host address" dft:"127.0.0.1"`
-	Port     uint16 `cli:"p,port" usage:"rqlited host port" dft:"4001"`
-	Prefix   string `cli:"P,prefix" usage:"rqlited HTTP URL prefix" dft:"/"`
-	Insecure bool   `cli:"i,insecure" usage:"do not verify rqlited HTTPS certificate" dft:"false"`
+	Protocol    string `cli:"s,scheme" usage:"protocol scheme (http or https)" dft:"http"`
+	Host        string `cli:"H,host" usage:"rqlited host address" dft:"127.0.0.1"`
+	Port        uint16 `cli:"p,port" usage:"rqlited host port" dft:"4001"`
+	Prefix      string `cli:"P,prefix" usage:"rqlited HTTP URL prefix" dft:"/"`
+	Insecure    bool   `cli:"i,insecure" usage:"do not verify rqlited HTTPS certificate" dft:"false"`
+	Credentials string `cli:"u,user" usage:"Basic Auth credentials (username:password)"`
 }
 
 const cliHelp = `.help				Show this message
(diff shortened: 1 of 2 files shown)


cmd/rqlite/main.go
```diff
@@ -127,7 +129,20 @@ func sendRequest(ctx *cli.Context, urlStr string, line string, argv *argT, ret i
 
 	nRedirect := 0
 	for {
-		resp, err := client.Post(url, "application/json", strings.NewReader(data))
+		req, err := http.NewRequest("POST", url, strings.NewReader(data))
+		if err != nil {
+			return err
+		}
+
+		if argv.Credentials != "" {
```
adelowo: `strings.TrimSpace` ?

cmd/rqlite/main.go
```diff
@@ -127,7 +129,20 @@ func sendRequest(ctx *cli.Context, urlStr string, line string, argv *argT, ret i
...
+			user, pass, err := net.SplitHostPort(argv.Credentials)
+			if err != nil {
+				return fmt.Errorf("invalid Basic Auth credentials format")
```
adelowo: Why is the error being discarded and not returned to the user.. should be `fmt.Errorf("inv..... %v", err)`

cmd/rqlite/main.go
```diff
@@ -127,7 +129,20 @@ func sendRequest(ctx *cli.Context, urlStr string, line string, argv *argT, ret i
...
 	nRedirect := 0
 	for {
-		resp, err := client.Post(url, "application/json", strings.NewReader(data))
+		req, err := http.NewRequest("POST", url, strings.NewReader(data))
+		if err != nil {
+			return err
+		}
+
+		if argv.Credentials != "" {
+			user, pass, err := net.SplitHostPort(argv.Credentials)
```
otoolep: While this works, `SplitHostPort` is not for splitting Basic Auth credentials. Instead please do:
```
creds := strings.Split(arg, ":")`
if len(creds) != 2 || creds[0] == "" || creds[1] == "" {
    return fmt.Errorf("invalid Basic Auth credentials format")
}
req.SetBasicAuth(creds[0], creds[1])
```
or something to this effect.

cmd/rqlite/main.go
joaodrp: No need for this. Spaces have already been trimmed at this stage from my tests.

cmd/rqlite/main.go
joaodrp: Because nothing useful would come from that error. Since I'm using the `SplitHostPort`, if the user would do `-u bob` (no password) this would raise the following message: `missing port in address`.

cmd/rqlite/main.go
joaodrp: Ok, that's fine. However, do you want to mandate a non-empty username and password? Because RFC 2617 doesn't. I'll make the change later today. Thanks.

cmd/rqlite/main.go
adelowo: Oh, great then

cmd/rqlite/main.go
adelowo: Then why not use `errors.New` since this would produce an error if ran through `go vet`

adelowo: or so I think

joaodrp: `go vet` is ok with that 👍

adelowo: 👍

otoolep:
Thank you @joaodrp

Issue #409, opened 2018-04-12
rqlite not allowing nodes to come up with different IPs


otoolep:
According to https://godoc.org/github.com/hashicorp/raft#Raft.AddVoter,  I believe nodes should be restartable with different Raft IP addresses. However rqlite is not operating in this manner, and the restarted node is starting an election, and failing to rejoin the cluster.

I need to figure out why. This blocks the 5.0 release.

otoolep:
Just to be clear, this is not a regression. rqlite prior to the port to Hashicorp Raft v1 didn't support changing IP addresses. It should now however.

otoolep:
The solution to this is shown in this PR: https://github.com/otoolep/hraftd/pull/23

This change needs to be ported to rqlite.

otoolep:
All done.

PR #411, opened 2018-04-13
Remove any matching node before a new node joins

otoolep:


otoolep:
Fixes https://github.com/rqlite/rqlite/issues/409

PR #412, opened 2018-04-13
Do not wait for log application before joining

otoolep:
This addresses the race between opening and the store and starting the
join. It's not robust however, and needs more work.


otoolep:
Further changes required to address https://github.com/rqlite/rqlite/issues/409

Issue #317, opened 2017-06-21
CLI should support displaying statement timings


otoolep:
The response from the nodes includes timing information. The CLI should support display of this, similar to the SQLite prompt.
```
sqlite> .timer on
sqlite> SELECT name FROM sqlite_master WHERE type="table";
Run Time: real 0.000 user 0.000000 sys 0.000000
```

PR #414, opened 2018-04-14
Add statement timings support to CLI

joaodrp:
Currently the CLI doesn’t display the statement timings after executing an SQL query. This change adds a new `.timer on|off` command to enable/disable this feature. When enabled, the CLI will append the `timings` URL query parameter to all query/execute requests sent to the API. When a response is received, the execution time is shown below the result.

Resolves: #317

### DIFF
--- a/cmd/rqlite/main.go
+++ b/cmd/rqlite/main.go
@@ -31,6 +31,7 @@ const cliHelp = `.help				Show this message
 .status				Show status and diagnostic information for connected node
 .expvar				Show expvar (Go runtime) information for connected node
 .tables				List names of tables
+.timer on|off			Turn SQL timer on or off
 `
 
 func main() {
@@ -41,7 +42,7 @@ func main() {
 			ctx.WriteUsage()
 			return nil
 		}
-
+		timer := false
 		prefix := fmt.Sprintf("%s:%d>", argv.Host, argv.Port)
 	FOR_READ:
 		for {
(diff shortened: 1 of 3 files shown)


joaodrp:
Hi @otoolep, would like your feedback about this change.

cmd/rqlite/main.go
```diff
@@ -31,6 +31,7 @@ const cliHelp = `.help				Show this message
 .status				Show status and diagnostic information for connected node
 .expvar				Show expvar (Go runtime) information for connected node
 .tables				List names of tables
+.timer on|off			Turn SQL timer on or off
```
otoolep: The help message looks like it needs to be pushed over move, so it lines up with the messages above.

cmd/rqlite/main.go
```diff
@@ -90,6 +93,14 @@ func main() {
 	})
 }
 
+func toggleTimer(op string, flag *bool) error {
+	if op != "on" && op != "off" {
+		return fmt.Errorf("not a boolean value: \"%s\". Assuming \"no\"", op)
```
otoolep: This message doesn't make sense to me. Did you mean "assuming on"? If so, why not just do that, and not return error?

joaodrp:
Here is a sample session: https://gist.github.com/joaodrp/f2ba9fb114aeef58a50dae12366095e9

cmd/rqlite/main.go
joaodrp: I thought the same, but if I add any more tabs it won't be properly aligned. If you try to edit the file [here](https://github.com/joaodrp/rqlite/edit/add-timings-support-to-cli/cmd/rqlite/main.go?pr=/rqlite/rqlite/pull/414) it's aligned. Also, if you take a look at the sample session below, you'll see that all lines are aligned when using the CLI.

cmd/rqlite/main.go
joaodrp: Ok, I should have mentioned that I copied the sqlite3 CLI behaviour and error message:

```
$ sqlite3
SQLite version 3.23.1 2018-04-10 17:39:29
Enter ".help" for usage hints.
Connected to a transient in-memory database.
Use ".open FILENAME" to reopen on a persistent database.
sqlite> .timer onn
ERROR: Not a boolean value: "onn". Assuming "no".
```

I can change it though.

otoolep: Yeah, probably change it.

otoolep:
Looks good, I'll wait for your final changes and then merge.

otoolep:
Great, thanks.

otoolep:
This change appears to have introduced a bug into the execute path. The CLI always shows number of rows affected as 0. When I roll back the change, it shows non-zero rows affected when an `INSERT` is performed.

otoolep:
Found it, easy fix. https://github.com/rqlite/rqlite/pull/416

joaodrp:
Indeed, I missed that one. Thanks.

PR #418, opened 2018-04-15
Close log when Store is closed

otoolep:


otoolep:
Done in https://github.com/rqlite/rqlite/pull/424

otoolep:
#424 may not merge now, re-opening.

Issue #422, opened 2018-04-17
Use Raft address as default node ID


otoolep:
Right now it's the hostname, it should be the Raft address since multiple different nodes could run on the same host.

PR #425, opened 2018-04-18
Use Raft network address as node ID, by default

otoolep:
Fixes issue #422.


otoolep:
Ported to "new" v5 via https://github.com/rqlite/rqlite/pull/601

Issue #138, opened 2016-05-09
HTTP service should remove peer API mapping on node remove


otoolep:
This will require the HTTP service getting passed the cluster service.

otoolep:
Yes, I should fix this. Now that someone has asked, I'll look into it within the next couple of weeks or so.

otoolep:
Didn't mean to close.

otoolep:
This might be better addressed in the move to Hashicorp Raft 1.0. That has even better support for dynamic environments, and is a better solution.

https://github.com/rqlite/rqlite/pull/377

Development is currently paused on that PR, but I do plan to complete it over the next few weeks.

otoolep:
I'm going to get back to this now, as part of release 5.0. I don't think addition of nodes is handled well now (with a dedicated cluster service). I may need to rewrite a bunch of this, so that node changes are handled more fully by the Raft system. With 5.0, it would be a good time.

Whatever I do will be a fully solution, handling addition and removal of peer API mappings properly.

otoolep:
Actually, comment is mistaken.

otoolep:
May be addressed once https://github.com/rqlite/rqlite/pull/424 is completed and merged.

otoolep:
Done in https://github.com/rqlite/rqlite/pull/434.

PR #434, opened 2018-04-21
Broadcast Store meta via standard consensus

otoolep:
With this change the cluster metadata (arbitrary key-value data associated with each node) is now broadcast across the cluster using the standard consensus mechanism. Specifically the use case for this metadata is to allow all nodes know the HTTP API address of all other nodes, for the purpose of redirecting requests to the leader.

This change removed the need for multiplexing two logical connections over the single Raft TCP connection, which greatly simplifies the networking code generally.

This change should also address https://github.com/rqlite/rqlite/issues/138, but the output of the status API has changed as a result.


otoolep:
https://github.com/rqlite/rqlite/issues/138

PR #449, opened 2018-04-26
Support for end-to-end TLS test

otoolep:
Fixes https://github.com/rqlite/rqlite/issues/441


otoolep:
Did this at the system-test level.

PR #450, opened 2018-04-26
Create a CLI command to restore the database from a dump file

eariassoto:
User's current FK constraint configuration is added to the restore file
so the DB preserves the same value


eariassoto:
Command example:

```
$ ./rqlite
127.0.0.1:4001> .restore restore.dump
last inserted ID: 1
rows affected: 1
database restored successfully
```

otoolep:

PR #461, opened 2018-05-05
 write backup directly to ResponseWriter instead of returning

sum12:
returned data could be of any size and would take up additional space in
ram, instead write the data directly to the passed writer, which in this
case would be http.ResponseWriter

Later this writer could be gzipped buffer writer

There is a wipped commit, as I wanted some input on how to test the backedup file.  I tested manually and it was fine.

I would have to use the db package. But not sure if is a good idea

### DIFF
--- a/http/service.go
+++ b/http/service.go
@@ -9,6 +9,7 @@ import (
 	"errors"
 	"expvar"
 	"fmt"
+	"io"
 	"io/ioutil"
 	"log"
 	"net"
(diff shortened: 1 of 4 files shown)


sum12:
Argh... Please let me know if such a change is acceptable, then I will proceed to resolve the conflict

otoolep:
Change looks great @sum12 -- let me know if it's ready to merge.

otoolep:
Thanks for your contribution @sum12

PR #505, opened 2018-06-01
Move connection management to Store

otoolep:


otoolep:
This seems to be still failing because `Close()` can be explicitly called while the `run()` is running. The solution is probably to move connection manager to the `Store` object and have it run the goroutine.

Issue #499, opened 2018-06-01
HTTP service occassionally timing out


otoolep:
HTTP service is occasionally failing to report ready, during full system test. This accounts for some recent CircleCI failures such as: https://circleci.com/gh/rqlite/rqlite/1983#tests/containers/2

Need to address.

otoolep:
Another: https://circleci.com/gh/rqlite/rqlite/2002#tests/containers/2

otoolep:
Perhaps the code that picks a free port and returns that address to the test port (`random_addr()`) is not quite finished closing the port down when it returns it to the node, for use by that node. A less racy way of finding free ports is probably required.

otoolep:
It might be best to actually push port 0 into the test node, and let is open the port. Then interrogate the node for its actual port, in the `start()` function, setting it on the node object. However there is a chicken-and-egg issue here -- can't interrogate unless you know the port.

PR #509, opened 2018-06-01
Hold onto random addresses until all assigned

otoolep:
Trying to fix https://github.com/rqlite/rqlite/issues/499


otoolep:

PR #524, opened 2018-07-20
add multistage dockerfile

lalyos:
This Dockerfile produces a 16MB image, compared to the 99MB original:
- https://hub.docker.com/r/lalyos/rqlite/tags/
- https://hub.docker.com/r/rqlite/rqlite/tags/

Other advantages an automated build provides:
- its more trusted by the community than a "pushed" image
- whenever you create a git tag, the appropriate docker image version is built automatically. (for example 4.3.0 is missing right now)
- the fresh version from master branch is always available as "latest" image tag. So some can play with new command, like `.dump` and `.restore`


otoolep:
Thank you for your contribution. Couple of questions:

-- how is the image built automatically? Where is that hooked in?
-- is this better done in https://github.com/rqlite/rqlite-docker (perhaps not, depending on the answer to the previous question).

lalyos:
Automated builds are described [here](https://docs.docker.com/docker-hub/builds/), but basically its connecting a git repository (with the Dockerfile in it), with dockerhub. Than every commit triggers dockerhub to check if it has any image to build (based on tag/branch patterns).

This Dockerfile is building the binary in a container, and not downloading it form github release page. Therefore the only needed 2 additional files are:
-  Dockefile
- entrypoint.sh
This way you dont, have to maintain a new directory for each version, just keep them in the same repo as the source.

Of course there are valid reasons for having a separate docker-myproject repo, but in this case its only 2 files, so it seems easier to include with the source.

lalyos:
I can see, there are alpine versions of the docker image: https://github.com/rqlite/rqlite-docker/tree/master/4.2.2-alpine but there are not pushed to dockerhub. 

So maybe you can just keep both repos (rqlite/rqlite-docker) but restructure rqlite-docker, to have the versions in branches, insted of directories.

Automated builds can be set-up also to build different image tags, from directories, but in that case, every commit will trigger all your images.

lalyos:
I know that a lot of docker images using the per-directory pattern (line [nginx](https://hub.docker.com/r/_/nginx/) or [alpine](https://hub.docker.com/r/_/alpine/)) for maintaining versions. But mostly they are the official/library images. They are using some different images build process, and therefor, those images are not "automated" images. Which probably can be trusted in the case of official images, but automated build is the recommended way.

otoolep:
@lalyos  -- thanks very much for this. This looks really helpful, as I hadn't been happy with the manual work involved with pushing up Docker images. I'm going to dig into this a little more, to ensure that I generally understand it, but your image is so much smaller than what I have pushing up until now that I think it makes sense just to drop the dockerfile here, and deprecate the other repo.

otoolep:
(Deprecate rqlite/rqlite-docker, that is.)

Dockerfile
```diff
@@ -0,0 +1,16 @@
+FROM golang:1.10-stretch
+ADD . /go/src/github.com/rqlite/rqlite
+WORKDIR /go/src/github.com/rqlite/rqlite
+RUN go get ./...
+RUN go install -ldflags="-X main.commit=$(git rev-parse HEAD) -X main.branch=$(git rev-parse --abbrev-ref HEAD)" ./cmd/...
+
+FROM frolvlad/alpine-glibc:alpine-3.8
```
otoolep: How do I know I can trust this container?

otoolep: I am assuming that this image forms the basis of the rqlite image. Let me know if I am missing something.

lalyos: yes that would be the base of your image. Its an automated build you can see the dockerfile here: https://hub.docker.com/r/frolvlad/alpine-glibc/~/dockerfile/

The reason behind this base image is, that I wasn't able to get a fully statically compiled binary version of rqlite. What I tried was the usual `CGO_ENABLED=0` but than I got some other error messages. I guess some of your dependencies does need cgo...

An alternative is that you start from an official alpine, and add the glibc binary packages (they are not officially released) so its 4 lines:
```
apk --no-cache add ca-certificates wget
wget -q -O /etc/apk/keys/sgerrand.rsa.pub https://raw.githubusercontent.com/sgerrand/alpine-pkg-glibc/master/sgerrand.rsa.pub
wget https://github.com/sgerrand/alpine-pkg-glibc/releases/download/2.27-r0/glibc-2.27-r0.apk
apk add glibc-2.27-r0.apk
```

otoolep: OK, thanks @lalyos for the extra information. Need to do a little bit more research. If you can provide any references to this glibc being used by others, that would be helpful.

otoolep:
Just an FYI: I am keen to merge this, but am away from rqlite development for a little while longer due to demands other than programming.

PR #531, opened 2018-09-28
Fix possible nil pointer dereference

runsisi:
Signed-off-by: runsisi <runsisi@zte.com.cn>


otoolep:
Thanks @runsisi -- this does seem necessary. Did you hit this during use, or you just noticed it?

runsisi:
yeah, it occurred while restoring from a very big postgresql data set.

PR #567, opened 2019-08-04
Added go-hclog support and formatted log messages to conform.

elliotcourant:
hclog uses a weird format for their log messages, rather than doing a
string format on the back end it takes the args for the log message
and prints them as alternating key/values. This replaces the current
logging implementation with hclog (so that the logging interface can be
passed directly down to the underlying raft implementation) and does its
best to make log messages conform to hclogs format.

I'm not sure I really like hclog, but it would be nice to have the uniform
logging in the mean time until another solution comes along. At the very
least this allows me to pass in my own logging interface that does match
hclog's interface and use that instead.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -10,9 +10,9 @@ import (
 	"errors"
 	"expvar"
 	"fmt"
+	"github.com/hashicorp/go-hclog"
 	"io"
 	"io/ioutil"
-	"log"
 	"math/rand"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


otoolep:
I'll be honest, I'm not sure how I feel about this. I really do prefer the simplicity of the standard log library. You might like to review this:

https://github.com/rqlite/rqlite/blob/master/CONTRIBUTING.md#third-party-libraries

otoolep:
Sure, I know there are lots of third-party libraries in the code already, but I prefer only bring them in if there is no alternative.

elliotcourant:
That's fair; It would technically add that library as a dependency to this code independent of raft. My use case for this project is kind of weird, I'm not using it as a standalone application but as an embedded database. So it really only makes sense for my use case to change the logging implementation.

otoolep:
Yeah, I think forking the code would be your best bet at this point, if you really want this in.

elliotcourant:
It's not that important I don't think. Thank you very much though!

PR #573, opened 2019-09-11
add proxy support

paulstuart:
To be able to proxy into a docker network would be handy. I would imagine it wouldn't hurt to have the flexibility in general.


otoolep:
Thanks for your contribution.

Issue #577, opened 2019-11-27
Https redirect to http protocol


weenxin:
I am using service discovery deployed in localhost.

you can think there is an dns routine:

node0 :  rqlite-quickstart-0.rqlite-quickstart-gvr.demo
node1 :  rqlite-quickstart-1.rqlite-quickstart-gvr.demo
node2 :  rqlite-quickstart-2.rqlite-quickstart-gvr.demo

and I deploy node0 using this command line

```
rqlited -http-addr 0.0.0.0:4001 -raft-addr 0.0.0.0:4002 \ 
-http-adv-addr rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4001 \
-raft-adv-addr rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4002 \
-disco-id 0a863868-38d6-46bb-b8b1-ff8a0c1189f0 \
-disco-url http://rqlite-service-discovery.rqlite:8090 \
-http-key /data/configdb/rqlite.pem \
-http-cert /data/configdb/client.pem \
-http-no-verify 
-node-encrypt 
-node-key /data/configdb/rqlite.pem \
-node-cert /data/configdb/client.pem \
-node-no-verify \
/data/db
```

node0 work well.

and then I start my second and third node using 

```
rqlited -http-addr 0.0.0.0:4001 -raft-addr 0.0.0.0:4002 \ 
-http-adv-addr rqlite-quickstart-1.rqlite-quickstart-gvr.demo:4001 \
-raft-adv-addr rqlite-quickstart-1.rqlite-quickstart-gvr.demo:4002 \
-disco-id 0a863868-38d6-46bb-b8b1-ff8a0c1189f0 \
-disco-url http://rqlite-service-discovery.rqlite:8090 \
-http-key /data/configdb/rqlite.pem \
-http-cert /data/configdb/client.pem \
-http-no-verify 
-node-encrypt 
-node-key /data/configdb/rqlite.pem \
-node-cert /data/configdb/client.pem \
-node-no-verify \
/data/db
```

```
rqlited -http-addr 0.0.0.0:4001 -raft-addr 0.0.0.0:4002 \ 
-http-adv-addr rqlite-quickstart-2.rqlite-quickstart-gvr.demo:4001 \
-raft-adv-addr rqlite-quickstart-2.rqlite-quickstart-gvr.demo:4002 \
-disco-id 0a863868-38d6-46bb-b8b1-ff8a0c1189f0 \
[...]

PR #578, opened 2019-11-27
Fix https redirect http protocol

weenxin:
[Https redirect to http protocol #577](https://github.com/rqlite/rqlite/issues/577)


otoolep:
While this would probably solve the problem in practice, it does assume that the other node is using HTTP just because this node is running HTTPS. This is not guaranteed to be true.

weenxin:
@otoolep  yes, Absolutely right.

But This will be very varied , When a cluster has three nodes,  node0 use https, node1 use http, and node2 use https.

I think we have  done a  reasonable assumption. All nodes in same cluster should use the same configuration.

What do you think ?

otoolep:
@weenxin -- take a look at https://github.com/rqlite/rqlite/pull/579. I think I prefer that approach. During the join logic the system tries HTTPS is HTTP fails. Can you try this patch, and see if it works for you? I tested it and it works for me.

otoolep:

Thanks for the PR however, I appreciate the contribution.

PR #579, opened 2019-11-28
When joining try HTTPS if HTTP fails

otoolep:


otoolep:
To avoid making a breaking change to the Disco format, this patches the join logic such that if a HTTP join fails, then system next tries a HTTPS join. I've tested this, and it seems to work well.

weenxin:
@otoolep Thanks for replying.

Maybe we have another problem to fix, if we approach this PR.

If we send request to a secondary node, It will reply
1. A http response with code=301 
2. The redirected address of primary node 

We need the protocol(http/https) in the redirected address. and we have no information about the leader protocol .

if we approach #578 , We have a reasonable protocol.

weenxin:
hi @otoolep 

I can not compile this PR

I run 
```
GOOS=linux  CGO_ENABLED=1  go build -o rqlited-linux -a   ./cmd/rqlited/...
```

Errors is : 
```
# github.com/rqlite/rqlite/store
store/store.go:178:16: undefined: raft.PeerStore
```

Seems like raft v1.1.1 has a big change:

https://github.com/hashicorp/raft/blob/a3fb4581fb07b16ecf1c3361580d4bdb17de9d98/membership.md#code-implications

otoolep:
Probably because you're trying to build it without vendoring? The 4.x branches use an older Go method for pinning down dependencies.

In any event, I decided to revert (forced) master to the 4.x code. I will close this PR and re-open it. I suggest you reclone the rqlite repo if you want to build.

otoolep:
This code is now in master, and master represents 4.x development again. No need for this PR.

Issue #580, opened 2019-11-28
redirect address lost path segment


weenxin:
## Version
master branch :  v5

## Test Environment
I provision a rqlite cluster with 3 nodes

using some command line like :

```
rqlited -http-addr 0.0.0.0:4001 \
-raft-addr 0.0.0.0:4002\
-http-adv-addr https://rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4001 \
-raft-adv-addr rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4002 \
-disco-id e7f33e53-a02f-4256-b405-221288ef7f72 \
-disco-url http://rqlite-service-discovery.rqlite:8090 \
-http-key /data/configdb/rqlite.pem \
-http-cert /data/configdb/client.pem -http-no-verify \
-http-ca-cert /data/configdb/rqlite.pem /data/db
```

and of course node1/node2 is just different in 
```
-http-adv-addr https://rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4001 \
-raft-adv-addr rqlite-quickstart-0.rqlite-quickstart-gvr.demo:4002 \
```

## How to replay this issue

And then i run command :

```
curl -v  -G 'https://rqlite-quickstart-2.rqlite-quickstart-gvr.demo.svc:4001/db/query?pretty&timings' --data-urlencode 'q=SELECT * FROM foo' --cacert keydir-readonly/rqlite.pem
```

It returns:

```
*   Trying 172.17.0.9...
* Connected to rqlite-quickstart-2.rqlite-quickstart-gvr.demo.svc (172.17.0.9) port 4001 (#0)
* found 1 certificates in keydir-readonly/rqlite.pem
* found 592 certificates in /etc/ssl/certs
* ALPN, offering http/1.1
* SSL connection using TLS1.2 / ECDHE_RSA_AES_128_GCM_SHA256
* 	 server certificate verification OK
* 	 server certificate status verification SKIPPED
* 	 common name: rqlite-quickstart (matched)
* 	 server certificate expiration date OK
* 	 server certificate activation date OK
* 	 certificate public key: RSA
* 	 certificate version: #3
* 	 subject: O=rqlite:server,CN=rqlite-quickstart
* 	 start date: Tue, 26 Nov 2019 05:54:56 GMT
* 	 expire date: Wed, 25 Nov 2020 05:54:56 GMT
* 	 issuer: O=rqlite:ca,CN=ca
* 	 compression: NULL
[...]

PR #581, opened 2019-11-28
Fix redirect lost url path error

weenxin:
fix the [redirect address lost path segment](https://github.com/rqlite/rqlite/issues/580) issue


otoolep:
Master is being developed right now (though very slowly). It may be completely rewritten, so I may or may not merge this. I expect master has lots of bugs.

otoolep:

Thanks for the PR however, I appreciate the contribution.

PR #599, opened 2019-12-19
Move to Hashicorp Raft v1

otoolep:


otoolep:
See https://github.com/rqlite/rqlite/pull/600

PR #602, opened 2019-12-20
Add method to Store that returns leader ID

otoolep:
With the advent of Raft IDs, this distinction matters.


otoolep:
Originated with https://github.com/rqlite/rqlite/pull/431

PR #619, opened 2020-01-08
Typo fix

shane-kerr:


otoolep:
Thanks for fixing that, @shane-kerr

Issue #646, opened 2020-05-04
Fix race condition during system test


otoolep:
Seems to be a node being deprovisioned without graceful shutdown, during the non-voter test. Waiting for a graceful shutdown makes the tests stall however.

```
2020/05/04 12:42:01 [DEBUG] raft-net: 127.0.0.1:33473 accepted connection from: 127.0.0.1:56636
[http] 2020/05/04 12:42:01 HTTP service Serve() returned: accept tcp 127.0.0.1:43409: use of closed network connection
==================
WARNING: DATA RACE
Read at 0x00c0005e0e48 by goroutine 202:
  github.com/mattn/go-sqlite3.(*SQLiteConn).prepare.func2()
      /go/pkg/mod/github.com/mattn/go-sqlite3@v2.0.2+incompatible/sqlite3.go:1702 +0x42
  github.com/mattn/go-sqlite3.(*SQLiteConn).prepare()
      /go/pkg/mod/github.com/mattn/go-sqlite3@v2.0.2+incompatible/sqlite3.go:1702 +0x15c
  github.com/mattn/go-sqlite3.(*SQLiteConn).exec()
      /go/pkg/mod/github.com/mattn/go-sqlite3@v2.0.2+incompatible/sqlite3.go:799 +0xe5
  github.com/mattn/go-sqlite3.(*SQLiteConn).Exec()
      /go/pkg/mod/github.com/mattn/go-sqlite3@v2.0.2+incompatible/sqlite3.go:793 +0x209
  github.com/rqlite/rqlite/db.(*DB).Execute.func1()
      /go/src/github.com/rqlite/rqlite/db/db.go:242 +0x298
  github.com/rqlite/rqlite/db.(*DB).Execute()
      /go/src/github.com/rqlite/rqlite/db/db.go:277 +0xf4
  github.com/rqlite/rqlite/store.(*Store).Apply()
      /go/src/github.com/rqlite/rqlite/store/store.go:816 +0x8fc
  github.com/hashicorp/raft.(*Raft).runFSM.func1()
      /go/pkg/mod/github.com/hashicorp/raft@v1.1.1/fsm.go:67 +0x512
  github.com/hashicorp/raft.(*Raft).runFSM()
      /go/pkg/mod/github.com/hashicorp/raft@v1.1.1/fsm.go:136 +0x3de
  github.com/hashicorp/raft.(*Raft).runFSM-fm()
[...]

PR #647, opened 2020-05-04
Shut down gracefully during system testing

otoolep:
Confirm finding in https://github.com/rqlite/rqlite/issues/646


otoolep:
This recent failure looks (https://circleci.com/gh/rqlite/rqlite/3353) more like an issue with snapshotting, that has been reported by others (system stalling). I will not merge this yet, however, in case it is related to the change in this PR.

PR #659, opened 2020-07-22
Fix Snapshot test

otoolep:
If the snapshot count didn't change, the test didn't actually fail.


otoolep:
Well, the Go test framework would eventually time out, but this was a poor implementation.

Issue #678, opened 2020-09-28
Authentication issues in rqlite


geosoft1:
I have a small cluster with three virtual machines running rqlited v5.5.0 secured with different self signed certificates and basic auth started by hand as following:

```
$HOME/bin/rqlited -auth auth.json -http-addr 10.10.180.61:4001 -raft-addr 10.10.180.61:4002 -http-cert server.crt -http-key server.key -node-encrypt -node-cert server.crt -node-key server.key -http-no-verify ~/node.1

$HOME/bin/rqlited -auth auth.json -http-addr 10.10.180.62:4001 -raft-addr 10.10.180.62:4002  -http-cert server.crt -http-key server.key -node-encrypt -node-cert server.crt -node-key server.key -http-no-verify -join https://root:12...@10.10.180.61:4001 ~/node.2

$HOME/bin/rqlited -auth auth.json -http-addr 10.10.180.63:4001 -raft-addr 10.10.180.63:4002  -http-cert server.crt -http-key server.key -node-encrypt -node-cert server.crt -node-key server.key -http-no-verify -join https://root:12...@10.10.180.61:4001 ~/node.3

```

Here is `auth.json` file:

```
[
  {
    "username": "root",
    "password": "1234",
    "perms": ["all"]
  }
]

```
Command to generate the ceritificates:

`openssl req -x509 -nodes -days 365 -newkey rsa:2048 -keyout server.key -out server.crt
`

I observed some authentication issues in rqlite console, only on some commands like in the following example. Look like some commands works and some commands get the authentication error, but in the same session under the same authentication credentials. I've tested on these three nodes which are accessible and responds to queries.

```
rqlite -i -s https -H 10.10.180.31 -u root:1234
Welcome to the rqlite CLI. Enter ".help" for usage hints.
10.10.180.31:4001> create table foo (id int)
0 row affected
10.10.180.31:4001> insert into foo values (1)
1 row affected
10.10.180.31:4001> .tables
+------+
| name |
+------+
| foo  |
+------+
10.10.180.31:4001> select * from foo
+----+
| id |
+----+
| 1  |
+----+
10.10.180.31:4001> .status
ERR! unauthorized
10.10.180.31:4001> .dump foo.dump
SQL text file written successfully
10.10.180.31:4001> .backup foo.backup
backup file written successfully
10.10.180.31:4001> .restore foo.backup
ERR! unauthorized
10.10.180.31:4001> .restore foo.dump
ERR! unauthorized

```
[...]

PR #682, opened 2020-09-28
Explicitly handle "no leader" during backup

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/678

PR #705, opened 2020-12-21
Use Protobuf for encoding Raft Log commands

otoolep:
This PR changes Raft Log Entry encoding from JSON to Protobuf. Furthermore, larger Raft commands (which can result from batching SQL statements, or individually long SQL statements) are compressed before encoding.

This primary reason for this change is to reduce IO load since that is one of the largest performance bottlenecks. It will also reduce internode traffic.

Legacy JSON-encoded commands are still handled by this code, so this change is backwards-compatible with previous releases in the v5 series.


otoolep:
It seems in practice that compression only has a marginal impact on performance, reducing IO load by about 10%, and resulting in a 10% increase in performance. 

Disk usage goes down in proportion to compression however, so for large compressible requests, the disk savings can be very significant.

Issue #710, opened 2020-12-24
Leader dies, tries to join itself on restart?


mrkurt:
I have a three node rqlite running, if I kill the followers they come back fine. If I kill the leader, it won't come back.

This is using the `-join` flag, not service discovery. After it dies, what was the leader is restarted with this command (`-join` comes from our DNS and includes the healthy replicas):

```
rqlited -join http://[fdaa:0:1:a7b:7f:0:636:2]:4001,http://[fdaa:0:1:a7b:81:0:637:2]:4001 \
    -node-id 7b06352 \
    -http-addr :4001 \
    -http-adv-addr [fdaa:0:1:a7b:7b:0:635:2]:4001 \
    -raft-addr :4002 \
    -raft-adv-addr [fdaa:0:1:a7b:7b:0:635:2]:4002 \
    -raft-timeout 5s \
    -raft-election-timeout 60s \
    -join-attempts 100 \
    /data/db
```

It starts to boot then gives `failed to join cluster` errors repeatedly until it gives up:

```    [store] 2020/12/24 22:22:30 ensuring directory at /data/db exists
    [store] 2020/12/24 22:22:30 opening store with node ID 7b06352
    [store] 2020/12/24 22:22:30 SQLite in-memory database opened
    2020-12-24T22:22:30.133Z [INFO] raft: initial configuration: index=6 servers="[{Suffrage:Voter ID:7b06352 Address:[::]:4002} {Suffrage:Voter ID:7f06362 Address:[fdaa:0:1:a7b:7f:0:636:2]:4002} {Suffrage:Voter ID:8106372 Address:[fdaa:0:1:a7b:81:0:637:2]:4002}]"
    [store] 2020/12/24 22:22:30 no bootstrap needed
    [rqlited] 2020/12/24 22:22:30 join addresses are: [http://[fdaa:0:1:a7b:7f:0:636:2]:4001 http://[fdaa:0:1:a7b:81:0:637:2]:4001]
    2020-12-24T22:22:30.141Z [INFO] raft: entering follower state: follower="Node at [::]:4002 [Follower]" leader=
[...]

PR #711, opened 2020-12-27
Ignore join request if node already part of cluster

otoolep:
rqlite used to work like this, but suffered a regression due to a change in how Hashicorp Raft worked. The manner it changed in was not public, so relying on it was always fragile.

See https://www.hyrumslaw.com/


otoolep:
https://github.com/rqlite/rqlite/issues/710

Issue #516, opened 2018-06-07
http/2.0 support


carr123:
it seems http2 client is not supported in rqlite. http2 has a considerable performance boost over http1.1,
http.ListenAndServerTLS( )  function in go net/http supports http2 server by default.
so, i recommend you to use  ListenAndServerTLS( ) in rqlited implementation source code.

otoolep:
Thanks for your report.

How do you know rqlite doesn't support HTTP2? Can you be more precise?

otoolep:
From my reading of the docs, by using Go 1.6 or later, you pick up http2 support. This means rqlite supports it.

joonas-fi:
Yes, a recent Go version should bring support for H2 out of the box.

I think @carr123 meant that that there's an issue in how rqlite uses TLS. I don't have actual operational experience with H2 but from the high level details I understand that the only way a HTTP client upgrades to H2 if it's advertised in the TLS stack, specifically the ALPN frame: https://en.wikipedia.org/wiki/Application-Layer_Protocol_Negotiation

> It is needed by secure HTTP/2 connections

If we take a look at rqlite's source code here: https://github.com/rqlite/rqlite/blob/master/http/service.go#L219

We notice that `tls.Listen()` call has no knowledge that what we'll be hosting over the TLS socket is a HTTP application, and therefore the TLS stack wouldn't ever know to advertise for H2 support.

Therefore, I think what we'd need to do is what @carr123 suggested: use `http.ListenAndServeTLS()` to tie TLS and HTTP more tightly to enable H2 support.

I can appreciate the more composable approach that you @otoolep probably strived for in giving either a TCP or TLS listener as an argument to `server.Serve()`, so the server doesn't have to know about the type of the connection it's hosted on. It sounds cleaner and that's probably what I would have done as well - but unfortunately that loses H2 support.

I just now learned this as well myself while studying this ticket, thanks @carr123  :)

joonas-fi:
Another alternative could maybe be to manually configure `NextProtos` member for [tls.Config](https://godoc.org/crypto/tls#Config) to include [h2](https://godoc.org/golang.org/x/net/http2#pkg-constants). This would probably be a simpler code change for rqlite.

otoolep:
Fixed. Better late than never. :-)

Thanks for the suggestion and pointers @joonas-fi @carr123

PR #716, opened 2020-12-27
Enable http/2.0 support for TLS

otoolep:
As per https://blog.cloudflare.com/exposing-go-on-the-internet/

HTTP/2 is enabled automatically on any Go 1.6+ server if:

- the request is served over TLS/HTTPS
- Server.TLSNextProto is nil (setting it to an empty map is how you disable HTTP/2)
- Server.TLSConfig is set and ListenAndServeTLS is used or
- Serve is used and tls.Config.NextProtos includes "h2" (like `[]string{"h2", "http/1.1"}`, since Serve is called too late to auto-modify the TLS Config)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,8 @@
-## 5.7.1 (unreleased)
+## 5.8.0 (unreleased)
+
+### New features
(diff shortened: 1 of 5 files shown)


otoolep:
https://github.com/rqlite/rqlite/issues/516

PR #729, opened 2021-01-22
Use macOS instead of OSX

wilva:
If you use OSX instead of macOS in your doc, it suggests that the source does not work on recent versions of macOS.


otoolep:
Thank you @wilva

PR #730, opened 2021-01-22
minor typo fix

The-Alchemist:


otoolep:
Both ways are valid spelling, right?

otoolep:
I'm Irish, that's the way it's spelled in Britain and Ireland. :-)

otoolep:
Thanks, though.

PR #738, opened 2021-01-29
Don't use temp file when snapshotting database

otoolep:
This will save a disk read and write (the size of the entire SQLite database) during a database snapshot operation.


otoolep:
Related: was going to be too much work to get this in upstream: https://github.com/mattn/go-sqlite3/pull/911

PR #739, opened 2021-01-29
Don't use temp file when restoring to in-memory SQLite

otoolep:
Saves a read and write, the size of the SQLite database, when restoring an in-memory database.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -467,6 +467,15 @@ func (db *DB) Serialize() ([]byte, error) {
 	return b, nil
 }
 
+// Deserialize forces a disconnect from the current database and then re-open
+// as an in-memory database based on the contents of the byte slice.
+func (db *DB) Deserialize(b []byte) error {
+	if err := db.sqlite3conn.Deserialize(b, ""); err != nil {
+		return fmt.Errorf("failed to deserialize database: %s", err.Error())
+	}
+	return nil
+}
+
 // Dump writes a consistent snapshot of the database in SQL text format.
 func (db *DB) Dump(w io.Writer) error {
 	if _, err := w.Write([]byte("PRAGMA foreign_keys=OFF;\nBEGIN TRANSACTION;\n")); err != nil {
--- a/db/db.go
+++ b/db/db.go
@@ -115,6 +115,34 @@ func LoadInMemoryWithDSN(dbPath, dsn string) (*DB, error) {
 	return db, nil
 }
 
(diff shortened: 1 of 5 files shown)


otoolep:
When I actually run a node, have it do a few snapshots, and then restart it -- the CLI is getting "not a database".

otoolep:
Don't merge this, not until I understand the errors (and if they are real, why the test suite isn't catching them).

otoolep:
Seems fine, can't reproduce the issue via the CLI. I've also added end-to-end testing to ensure restoring to both in-mem and on-disk databases works fine.

otoolep:
I got the system into this state again.

```
127.0.0.1:4001> select count(*) from foo
ERR! file is not a database
127.0.0.1:4001> select count(*) from foo
ERR! file is not a database
127.0.0.1:4001> 
EOF (CTRL+D)
~/repos/rqlite/src/github.com/rqlite/rqlite/cmd/rqlite (no-disk-during-restore)$ ./rqlite 
Welcome to the rqlite CLI. Enter ".help" for usage hints.
Connected to version 5
127.0.0.1:4001> select count(*) from foo
ERR! file is not a database
127.0.0.1:4001> select count(*) from foo
ERR! file is not a database
127.0.0.1:4001> select count(*) from foo
ERR! file is not a database
127.0.0.1:4001> 
```

Not good.

otoolep:
The test case seems to be:

- start a node using in-mem database
- trigger snapshot
- restart node
- queries break, with "not a database".

otoolep:
Very strange that none of the current testing catches this, only the command line work.

otoolep:
I think I see the issue. This problem doesn't occur 100% of the time with the following script, but pretty close:

```bash
#!/bin/bash
DATA_DIR=`mktemp -d`

$RQLITED -raft-snap 13 -raft-snap-int=1s $DATA_DIR &
sleep 3

$RQBENCH -b 10 -m 1 -n 500 -o 'CREATE TABLE foo (id INTEGER NOT NULL PRIMARY KEY, name TEXT)' 'INSERT INTO foo(name) VALUES("fiona")'

killall rqlited
sleep 3

$RQLITED -raft-snap 50000 -raft-snap-int=500s $DATA_DIR
```

SQLite is using the memory passed in, and making changes to it -- that memory represents the database after all. But that memory is in the Go memory space. And I'm sure this is allowed, since Go is a garbage collected language.

Issue #731, opened 2021-01-22
Minimize time SQLite file is missing during restart


otoolep:
https://news.ycombinator.com/item?id=25873029

I'll need to think about this.

cc @mrkurt

mrkurt:
Oh nice! I have been hesitant to ask for these features because the problem we're tackling is not _all_ right for rqlite. But if it can do this, it gets us closer!

otoolep:
Maybe checksumming the file might help, but I'm not sure.

@mrkurt -- what's your write request rate?

otoolep:
Note to myself: opening up the log before fully initializing the store, might help understand how many entries are to be applioed.

otoolep:
If I do decide to make improvements in this area, which would help start-up times generally with an SQLite file on disk, https://github.com/rqlite/rqlite/pull/733 may be relevant.

mrkurt:
@otoolep we're doing about 700-1200 writes per second to Consul right now. Not all of those _should_ be going to consul (or rqlite). Only about 100/s of those make sense for Consul or rqlite.

The rest is state that should really be gossiped around.

otoolep:
I have a new idea that I think will work -- build the SQLite database in RAM first, which is very fast. Then, if in on-disk mode, write it over the disk just before the node completes coming up. I hope to release this functionality in the next release if I can make it work, without introducing risk into the code base.

otoolep:
@mrkurt -- I think I've made a big improvement here. I'm now seeing on-disk nodes start as quickly as in-memory nodes. The SQLite file will only be missing for a very short time.

See PR #750 for details. Can you build master, and try it out? I'd rather not release a new version until some other folks have tested the work.

otoolep:
If you use Windows, there is a build available already: 

https://ci.appveyor.com/api/buildjobs/mt7k2mm6139sut9x/artifacts/rqlite-latest-win64.zip

otoolep:
@mrkurt -- 5.10.0 is available now. In this release on-disk based systems restart almost as quickly as memory-based systems. Try it out.

PR #747, opened 2021-02-01
Time snapshot, restore, and startup times

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/731

PR #750, opened 2021-02-02
Switch to on-disk after initial logs applied

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/731

PR #776, opened 2021-03-05
feat: Support to specify the source IP

osxlinux:
Support to specify the source IP when you are in a complex network environment.


tcp/transport.go
```diff
@@ -52,7 +53,16 @@ func (t *Transport) Open(addr string) error {
 
 // Dial opens a network connection.
 func (t *Transport) Dial(addr string, timeout time.Duration) (net.Conn, error) {
-	dialer := &net.Dialer{Timeout: timeout}
+	//dialer := &net.Dialer{Timeout: timeout}
```
otoolep, line 56: Please remove line, don't comment out.

cmd/rqlited/main.go
```diff
@@ -132,6 +133,8 @@ func init() {
 	flag.IntVar(&compressionBatch, "compression-batch", 5, "Request batch threshold for compression attempt")
 	flag.StringVar(&cpuProfile, "cpu-profile", "", "Path to file for CPU profiling information")
 	flag.StringVar(&memProfile, "mem-profile", "", "Path to file for memory profiling information")
+	flag.StringVar(&sourceIp, "source-ip", "", "Specify a source ip address, when your node has multiple ip address segments")
```
otoolep, line 136: What do you mean by an "IP address segment"? Can you give me an example?

cluster/join_test.go
```diff
@@ -35,7 +35,7 @@ func Test_SingleJoinOK(t *testing.T) {
 
 	defer ts.Close()
 
-	j, err := Join([]string{ts.URL}, "id0", "127.0.0.1:9090", false, nil,
+	j, err := Join("", []string{ts.URL}, "id0", "127.0.0.1:9090", false, nil,
```
otoolep, line 38: Let's add a unit test where we actually pass in the local addr, and not just "".

cluster/join.go
```diff
@@ -50,11 +50,20 @@ func Join(joinAddr []string, id, addr string, voter bool, meta map[string]string
 	return "", ErrJoinFailed
 }
 
-func join(joinAddr, id, addr string, voter bool, meta map[string]string, tlsConfig *tls.Config, logger *log.Logger) (string, error) {
+func join(sourceIp string, joinAddr, id, addr string, voter bool, meta map[string]string, tlsConfig *tls.Config, logger *log.Logger) (string, error) {
```
otoolep, line 53: Go variable style states this should `sourceIP` -- and `srcIP` would be better (Go prefers short names). Also, `string` is not required as all types following are the same until `voter`.

cluster/join.go
```diff
@@ -50,11 +50,20 @@ func Join(joinAddr []string, id, addr string, voter bool, meta map[string]string
...
 	if id == "" {
 		return "", fmt.Errorf("node ID not set")
 	}
-
+	//The specified source IP is optional
```
otoolep, line 57: Add space between // and "The".

cmd/rqlited/main.go
otoolep: Variable should be named `sourceIP`.

otoolep (commented):
Thanks for your contribution. Some comments.

cmd/rqlited/main.go
osxlinux: "IP address segment"  When your host has multiple network segments, that is, multiple network cards, one of the address segments has a default route, but the other is the address that works as rqlite, that is, multiple address segments
```
eth0:  192.168.9.2/24  # Has a default route  
eth1:   172.16.9.2/24  #There is no default route, the policy is controlled by the "ip rule" command, rqlite works in this address segment
```

otoolep:
Actually, I have a question about this. Why is it needed? The join function uses a HTTP request. Does that HTTP request fail if you don't configure the local address during that request, with your system?

otoolep:
I know HTTP goes over TCP. Are you saying that without this, and with your setup, the HTTP (join) request fails? If so, why?

otoolep:
I'm not a network expert, but if you want the HTTP request to go out a specific interface, shouldn't that be some sort of routing config set in the OS?

Since this functionality is exposed in the Go libraries, it still seems reasonable to expose it here. I'd just like to understand better why this control is available.

otoolep:
OK, thanks for the extra information.

This functionality will be in 5.11.0 (with slightly changed command-line options). Seems like you can build your own source, so you don't need a release right now. If you do, let me know.

Issue #789, opened 2021-04-23
cannot run quries to cluster members - only the leader responds


Theog75:
after a while  (a couple of hours/days) our rqlite clusters (V5.11.1) reach a point where we can only query the leader:

for example running the query using any member other than the leader returns:
```
bash-4.4$ /rqlite -H rqlite-0.rqlite
Welcome to the rqlite CLI. Enter ".help" for usage hints.
Connected to version v5.11.1
rqlite-0.rqlite:4001> select * from rules
ERR! server responded with 503 Service Unavailable: not leader

rqlite-0.rqlite:4001> select * from  issues
ERR! server responded with 503 Service Unavailable: not leader

```

 the status response from one of the followers is :
```

curl http://rqlite-0.rqlite:4001/status?pretty


{
    "build": {
        "branch": "master",
        "build_time": "2021-04-13T10:24:42-0400",
        "commit": "927611c82c72056a99e20cde3279fac7fdf51484",
        "version": "v5.11.1"
    },
    "http": {
        "addr": "10.233.122.246:4001",
        "auth": "disabled",
        "redirect": ""
    },
    "node": {
        "start_time": "2021-04-22T19:32:14.932309505Z",
        "uptime": "16h38m28.56345708s"
    },
    "runtime": {
        "GOARCH": "amd64",
        "GOMAXPROCS": 8,
        "GOOS": "linux",
        "num_cpu": 8,
        "num_goroutine": 17,
        "version": "go1.15.7"
    },
    "store": {
        "addr": "10.233.122.246:4002",
        "apply_timeout": "10s",
        "db_conf": {
            "DSN": "",
            "Memory": true
        },
        "dir": "/node",
        "dir_size": 14560570,
        "election_timeout": "5s",
        "heartbeat_timeout": "4s",
        "leader": {
            "addr": "10.233.92.133:4002",
            "node_id": "rqlite-4"
        },
        "metadata": {
            "rqlite-0": {
                "api_addr": "rqlite-0.rqlite:4001",
                "api_proto": "http"
            },
            "rqlite-1": {
                "api_addr": "rqlite-1.rqlite:4001",
                "api_proto": "http"
            },
            "rqlite-2": {
                "api_addr": "rqlite-2.rqlite:4001",
                "api_proto": "http"
            }
        },
        "node_id": "rqlite-0",
        "nodes": [
            {
                "id": "rqlite-0",
                "addr": "10.233.122.246:4002"
            },
            {
[...]

PR #792, opened 2021-04-25
Fetch leader HTTP API address on demand

otoolep:
This release implements a significant design change. In the 5.0 series, Follower nodes learned the HTTP API address of a cluster Leader via information - known as Metadata - that each node wrote to the Raft log. This Metadata was then available to each node in the cluster, if it needed to redirect queries to the cluster Leader. However this design is somewhat complex, and requires the tracking of state, in addition to the SQLite database. It also meant that if the Metadata got out of sync with the Raft state, the cluster could be a significantly degraded state. In this new system, a node now queries the Leader as needed, when that node needs to learn the HTTP API address. And without any possibility of discrepancy between Metadata and Raft state, a whole class of potential bugs is removed.

The request for the Leader HTTP API address travels over the existing Raft TCP connection, so does not introduce any new failure modes. This multiplexing of the Raft connection is performed via the mux package.


otoolep:
Start work to revert to on-demand node "metadata" as opposed to sending it through consensus, which may be hard to get right for all node lifecycle operations. 

https://github.com/rqlite/rqlite/issues/789

PR #813, opened 2021-05-21
Add more startup logging

otoolep:


otoolep:
Help debug https://github.com/rqlite/rqlite/issues/811

Issue #818, opened 2021-06-16
Resuming instance from different network address with existing raft.db triggers unnecessary election and fails to join


brennantaylor:
**What version are you running?**
Tested 5.9.0 and 6.0.0 on OSX. Was also having the same issues with the `rqlite/rqlite:6.0.0` docker image.

**What did you do?**
Started 3 rqlited's locally on different ports to form a 3 proc cluster, say node-ids `rqlite-0`, `rqlite-1` and `rqlite-2`.

My intent was to remove rqlite-0 running on port 4001+4002 and re-add it on another port, say 4007+4008 to simulate a different network address like one may find when pods migrate around in Kubernetes.

**What did you expect to happen?**

When rejoining the cluster by either using the network addresses inside the existing raft log or by using the new `-join` params provided it would connect to the other members and catch its log up to the latest point and continue along its merry way at the new network addresses without needing to replicate the entire raft log.

**What happened instead?**

No matter what I tried it would try to trigger a new vote that the other nodes would reject because there was already a healthy leader.

The only way to get it to rejoin was by wiping out the `raft.db` for that process and having it join fresh.

**Please include the Status, Nodes, and Expvar output from each node (or at least the Leader!)**

Start 3 rqlited's. Have the 2nd and 3rd join the first to form the cluster.

```
$ rqlited -node-id=rqlite-0 -http-addr=localhost:4001 -raft-addr=localhost:4002 0
$ rqlited -node-id=rqlite-1 -http-addr=localhost:4003 -raft-addr=localhost:4004 -join localhost:4001 1
$ rqlited -node-id=rqlite-2 -http-addr=localhost:4005 -raft-addr=localhost:4006 -join localhost:4001 2
```

Check cluster state:

```
$ curl "http://localhost:4003/nodes?pretty"
{
    "rqlite-0": {
        "api_addr": "http://localhost:4001",
        "addr": "127.0.0.1:4002",
        "reachable": true,
        "leader": true
    },
    "rqlite-1": {
        "api_addr": "http://localhost:4003",
        "addr": "127.0.0.1:4004",
        "reachable": true,
        "leader": false
    },
    "rqlite-2": {
        "api_addr": "http://localhost:4005",
        "addr": "127.0.0.1:4006",
        "reachable": true,
        "leader": false
    }
}
```

Kill rqlite-0 and check to see that the leader has moved.

```
$ curl "http://localhost:4003/nodes?pretty"
{
    "rqlite-0": {
[...]

PR #822, opened 2021-06-26
Don't ignore -join even if previous state exists

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/818

PR #825, opened 2021-07-02
Migrate to Protocol Buffers at all layers

otoolep:
This will make it much easier to implementing transparent request forwarding.


otoolep:
Was reimplemented in https://github.com/rqlite/rqlite/pull/859

Issue #571, opened 2019-08-27
can we use rqlite as a libary?


the-oaks:
is it practically possible to use rqlite as a library that embedded in another Go application rather than a standalone cmd? If so, is it possible to provide a minimal working example (that replicate sqlite data)? or some hints on how shall we start on this minimal working example? thanks

the-oaks:
thanks a lots

narg95:
Hi, I know this issue is closed, but I am interested in this feature, I made some experiments in a fork and it is straightforward to get it embedded in a library. If you are open for this change I would be happy to open a PR with basically the following changes:

1. new package for the library e.g. `pkg/lib` 
2. A `Config` struct with all config settings e.g.:

```go
type Config struct {
	DataPath               string
	HTTPAddr               string
	HTTPAdv                string
	JoinSrcIP              string
	TLS1011                bool
	AuthFile               string
	X509CACert             string
	X509Cert               string
	X509Key                string
	NodeEncrypt            bool
	NodeX509CACert         string
	NodeX509Cert           string
	NodeX509Key            string
...
```
3. A new `Server` type which is able to `Start()` and `Stop()`
```go
type Server struct{
...
}

func (s *Server) Start(ctx context.Context, cfg *Config) error {
    // logic from main() using cfg instead of gloval vars
   // it starts the server async
}


func (s *Server) Stop() {
    // stops the server by signaling the stop
    // very likely canceling the context.Context
    // or reusing the `terminate` channel which is currently used with signal.Notify(terminate, os.Interrupt)
    // by sending an artificial os.Signal
}
```

5. [Optional] integrate the new `Config` and `Server` into the `cmd/rqlited/main.go` so there is only a centralized place for server bootstrapping 

What do you think?

narg95:
This proposal is just a first step in having a library, because it will still communicate using http. The next step is to extract the handlers logic inside the `http/service.go` into a library. Exposing the `Store`, so that the client create the commands, might be sufficient or perhaps a wrapper around it would be needed due to some extra checks or more complex validations, in any case your expertise @otoolep  will be required here =)

otoolep:
I'm not against this in principle, and am open to your changes. However I do have one practical concern -- I don't want to be limited in the future making changes to the structure of the rqlite implementation because folks have started relying on an API other than the HTTP API. This may not be an issue in practise (particularly with versioning and forking) but that's my concern.

[...]

PR #826, opened 2021-07-13
add server

narg95:
# Description

- Add the `server` package
- Add a `Server` struct to control the life cycle of a rqlite server with methods like `server.Start()` and `server.Stop()` server
- move server bootstrapping from `cmd/rqlited/main.go` to the new `Server` definition
- Introduce a `Config` struct to hold the server configuration settings

Issue #571 

# Questions

I was not able to find where the http server is gracefully shutdown, it looks like missing.


narg95:
Please take a look to added test and let me know if you find it valuable or if that behavior is already covered by system tests, because currently the race detector is complaining but `server.Start` is not meant to be called concurrently.

otoolep:
Thanks for your work -- this does make sense. I can see what you've got in mind.

I do agree that the `main` layer has got too big, and that this would be one way to refactor. But there may be other ways, I'd need to think about. There may be more than one concept wrapped up in `Server` for example. The config blob is very big, for example -- but that reflects all the composition done by `main` which this PR pushes into a `Server` object.

I'll have to think about this. It's not that it doesn't make sense, but any large change to the codebase in the past was made because *I* could really see it was needed, and it was a step towards where *I* wanted to take the database. At this time time I don't have a strong enough incentive to make such a big change, which I don't feel strongly about, to meet a need that is distinct from rqlite's core goal. Does this make sense?

So I'm happy to leave this open, but unless it makes it also significantly easier to work with the codebase, I'm not going to rush into merging this.

otoolep:
rqlite's core goal == lightweight, distibuted database built on SQLite.
Goal of this PR == make rqlite usable as a package by others.

These are different goals, that's all. And I only have cycles to work on the first goal. They aren't necessarily in conflict, but wrapping a higher-level abstraction around the current code is going to take me weeks to get comfortable with. But you've given me something to think about. :-)

otoolep:
Thanks for this contribution, I'm going to be making very significant changes to main.go now, so it was best to hold off.

PR #883, opened 2021-09-08
Add missing returns after HTTP errors

otoolep:


otoolep:
Deal with "superfluous writes" issue that occurred in https://github.com/rqlite/rqlite/issues/861#issuecomment-915023659.

PR #922, opened 2021-11-09
fix:solve the float64 problem during JSON deserialization

tiswong:
During the execution of SQL through the console, the result of numeric type will be deserialized to float64 and displayed as scientific notation


otoolep:
Thank you @tiswong -- I noticed this and meant to fix it, but wasn't sure how. This is great.

otoolep:
This is an important enough change that I released 6.8.0.

Issue #347, opened 2017-08-30
Single-node system always contacts disco service, if requested


otoolep:
When a node first comes up, and if it has been passed a disco-id, it contacts the disco service. This is as expected. However, on a restart, it contacts it again. Ideally it shouldn't do this, since it has already registered itself.

otoolep:
A second node, which joins a cluster through the disco service, does not suffer from this issue.

otoolep:
It should be noted that it's *ok* that the node does this, since the disco service is idempotent. This is about avoiding network calls unless necessary.

Issue #923, opened 2021-11-09
One node is unavailable in discovery mode causes a new node cannot join the cluster


wujushan:
I am developing by using the verison **6.7.0**. I find that, in discovery mode, if one node is unavailable will cause a new node cannot join the cluster. because the new node uses the IP of that unavailable node. 

I deployed the discovery service [rqlite-disco-express](https://github.com/rqlite/rqlite-disco-express) in my own PC. 

The log below is printed by new node. I find that the new node always uses the incorrect ip(**172.16.160.103:4001**) to join the cluster rather than trys the other one. 

**Please try a few more times if you cannot get the error log like the one below.**

```
rqlited] 2021/11/09 08:20:17 rqlited starting, version 6, commit unknown, branch unknown, compiler gc
[rqlited] 2021/11/09 08:20:17 go1.17.2, target architecture is arm, operating system target is linux
[rqlited] 2021/11/09 08:20:17 launch command: /bin/rqlited -http-addr 0.0.0.0:4001 -raft-addr 0.0.0.0:4002 -http-adv-addr 172.16.160.49:4001 -raft-adv-addr 172.16.160.49:4002 -disco-id e2d437c0-3d66-11ec-8442-8fa65a76e101 --disco-url http://172.16.160.49:10808 -on-disk /rqlite/file/data
[mux] 2021/11/09 08:20:17 mux serving on [::]:4002, advertising 172.16.160.49:4002
[rqlited] 2021/11/09 08:20:17 Raft TCP mux Listener registered with 1
[rqlited] 2021/11/09 08:20:17 preexisting node state detected in /rqlite/file/data
[rqlited] 2021/11/09 08:20:17 registering with Discovery Service at http://172.16.160.49:10808 with ID e2d437c0-3d66-11ec-8442-8fa65a76e101
[discovery] 2021/11/09 08:20:17 discovery client attempting registration of 172.16.160.49:4001 at http://172.16.160.49:10808/e2d437c0-3d66-11ec-8442-8fa65a76e101
[discovery] 2021/11/09 08:20:17 discovery client successfully registered 172.16.160.49:4001 at http://172.16.160.49:10808/e2d437c0-3d66-11ec-8442-8fa65a76e101
[rqlited] 2021/11/09 08:20:17 Discovery Service responded with nodes: [172.16.160.43:4001 172.16.160.49:4001 172.16.160.103:4001]
[rqlited] 2021/11/09 08:20:17 join addresses specified, node is not bootstrapping
[...]

PR #925, opened 2021-11-12
Ignore disco ID if there is preexisting state

otoolep:
Fixes https://github.com/rqlite/rqlite/issues/347


otoolep:
Related: https://github.com/rqlite/rqlite/issues/923

PR #931, opened 2021-11-26
Don't run heavy memory tests in parallel

otoolep:
Let's see if this helps these tests run faster, especially with -race, and prevent CircleCI timeouts.


otoolep:
CircleCI times seem better, at least not worse. The thinking here is that by ensuring the heavy memory tests all resources at the time they are run, they may run better and incur less context switching.

Issue #935, opened 2022-01-04
authentication improvements


nferch:
**Is your feature request related to a problem? Please describe.**
The current authentication functionality is great for applications and human users, but operating an rqlite cluster with authentication is a bit of a pain.

* Monitoring the cluster for a service discovery healthcheck (I use Nomad/Consul) requires authentication
* Credentials for joining a cluster are exposed via the command line.

**Describe the solution you'd like**
I have two ideas for improvements:

* Create a policy for unauthenticated access, so I don't have to create a user and embed credentials for service monitoring
* Ability to specify credentials for cluster join somewhere other than the command line.

**Describe alternatives you've considered**
* It may be possible to configure Nomad to query Vault for monitoring credentials, but I haven't investigated this.
* And of course, foregoing encryption and authentication is always an option.

**Additional context**
I'm running rqlite on Nomad, storing secrets in Vault. I've created a wrapper script that queries Consul for other cluster nodes to facilitate cluster joining.

otoolep:
Thanks for your suggestion, I'm interested in improving rqlite in the manner you suggest. However I need a little more detail.

>Monitoring the cluster for a service discovery healthcheck (I use Nomad/Consul) requires authentication

I presume you mean monitoring a rqlite cluster for health. How are you doing it today? Are you hitting the `readyz` endpoint?

https://github.com/rqlite/rqlite/blob/master/DOC/DIAGNOSTICS.md#readiness-checks 

>Ability to specify credentials for cluster join somewhere other than the command line.

Yes, this is a bit of a weakness as looking at the process table could reveal the details. I'm open to suggestions. Environment variables, perhaps?

nferch:
> Thanks for your suggestion, I'm interested in improving rqlite in the manner you suggest. However I need a little more detail.
> 
> > Monitoring the cluster for a service discovery healthcheck (I use Nomad/Consul) requires authentication
> 
> I presume you mean monitoring a rqlite cluster for health. How are you doing it today? Are you hitting the `readyz` endpoint?

There's two parts:

* A `service` declaration in my Nomad job that will poll the `/status` endpoint and publish a Consul service when it receives a 2xx response
* A wrapper startup script that queries Consul through DNS to fetch all of the live servers and populate the command line with a '-join` statement.

[...]

PR #940, opened 2022-01-06
Enhance Authentication and Authorization system

otoolep:
This change enhances the authentication and authorization system in two ways.

- Adds a new command-line option, `-join-as` which tells rqlite to read the username and password from the auth config file, when doing the join. This allows us to avoid putting passwords in the launch command.
- Adds support for a pseudouser in the auth config, indicated by `*`. This allows indicating that all requests -- even anonymous requests -- have any permissions assigned to `*`.


otoolep:
Actually, it would be better to do this as a new command-line param `-join-as`. Oh well.

otoolep:
https://github.com/rqlite/rqlite/issues/935

Issue #157, opened 2016-06-01
CLI should accept list of nodes


otoolep:
This would allow the CLI to try connecting to a different node is the first node goes down.

mainframe:
+1

otoolep:
A better approach would be to have the Go client library (which needs to be fully developed) implement this functionality. It should not be CLI-specific. Then the CLI would use the new client library.

snapo:
maybe dns in a records? or a hostlist json file? ... just some ideas :-)

otoolep:
Adding it the CLI is definitely an option though.

chermehdi:
Is somebody working on this? Seems like something that I can take on if nobody is actually doing it

otoolep:
Still needs to be done -- contributions welcome. Please read https://github.com/rqlite/rqlite/blob/master/CONTRIBUTING.md

chermehdi:
A couple of clarifications:

* By the CLI do we mean the `rqlite` or the `rqlited`? i.e do we need to specify the list of nodes to try to connect to as a client (rqlite), or specify the list of existing cluster nodes (rqlited)?

otoolep:
We mean the CLI, the code here: https://github.com/rqlite/rqlite/tree/master/cmd/rqlite

Before you do any significant coding, I suggest you show us what you plan to do in terms of a) the experience of using the CLI with this functionality in place, and what exactly it will do with those extra options. For instance, how will failure be detected, any retry, what node will be tried next, and when would the CLI give up (if at all).

chermehdi:
What I had in mind for this was:

Introduce a new struct to wrap the HTTP client calls which will be aware of extra things:

* Nodes that we wish to send queries to
* Health status of this node (checked and updated in the background)
* (optionally) A retry policy + a maximum timeout to respond before considering the node as unavailable).

- Instead of calling `http.Client.Do` directly we would call a proxy method on the client which will forward to one of the available hosts.

- If we are unable to connect to a node, we prompt to the user that the host is unavailable and we are trying another host, if that host is successful following requests will be sent to that host, a host change is only required if the active host is unavailable.
- A goroutine will continuously check host availabilities (every `X seconds`) on some `/healthz` API available on the node (not sure if that already provided by the nodes) and update the list of hosts that are unavailable.

otoolep:
[...]

PR #947, opened 2022-01-08
[Draft] Accept alternative servers to connect to in rqlite cli

chermehdi:
This PR adds a new flag `-alternatives` to supply a comma-separated list of hosts to fall back to if the primary host fails.

This is done by introducing a new client that handles host fallbacks and also (auth and redirects).

The implementation details have been discussed on #157 .


http/client.go
```diff
@@ -0,0 +1,186 @@
+package http
```
otoolep, line 1: Thanks for this file, I think it may be useful for some client library work I plan to do in the future. Factoring out this code may be useful to me. However, I prefer the code in the `http/` directory to be just for `rqlited`. So can you move this to the CLI directory instead? I'd the following path: `cmd/rqlite/http/client`.

If this client file turns out to work well for the CLI, I would actually make it a core part of a new client library -- which the CLI would import. Make sense?

cmd/rqlite/main.go
```diff
@@ -6,6 +6,7 @@ import (
 	"crypto/x509"
 	"encoding/json"
 	"fmt"
+	httpcl "github.com/rqlite/rqlite/http"
```
otoolep, line 9: I prefer non-standard-library imports to be group separately, and in a second block, is the style in this file already.

cmd/rqlite/main.go
```diff
@@ -93,6 +95,12 @@ func main() {
 		}
 		term.Close()
 
+		hosts := createHostList(argv)
+		recoveringClient := httpcl.NewClient(client, hosts,
```
otoolep, line 99: I know naming is hard :-) but I would just call it `client`.

otoolep: It's just a smarter client, not a "recovering client".

cmd/rqlite/query.go
```diff
@@ -2,13 +2,12 @@ package main
 
 import (
 	"fmt"
+	"github.com/mkideal/cli"
+	"github.com/mkideal/pkg/textutil"
```
otoolep, line 6: Same as above, this grouping is not consistent with the existing style.

otoolep:
Thanks, this looks promising. I think you're on the right track. Some comments.

http/client.go
chermehdi: Sure sounds reasonable. 

I am still not that familiar with the code base, so I put the client in the first directory that seemed logical, I didn't know the code for the cli is all grouped under `cmd/rqlite` :sweat_smile:

cmd/rqlite/main.go
chermehdi: client is already declared before in the same scope.

and you are right about the `recovering` part, let me find some better name for it :)

chermehdi:
I added the necessary changes as per the comments above.

otoolep:
Can you show me an example session, of how it looks when it failovers to another node?

cmd/rqlite/http/client_test.go
```diff
@@ -0,0 +1,162 @@
...
+	u1, _ := url.Parse(node1.URL)
+	u2, _ := url.Parse(node2.URL)
+	client := NewClient(httpClient, []string{u1.Host, u2.Host})
+	res, err := client.Query(url.URL{
+		Path: "/",
+	})
+	defer res.Body.Close()
+
+	if err == nil {
+		t.Errorf("expected error")
```
otoolep, line 66: Minor -- can you add a little more explanation to this string? Why was an error expected? This will help understand the test better, especially if it fails.

cmd/rqlite/execute.go
```diff
@@ -25,96 +26,70 @@ type executeResponse struct {
...
-		resp, err := client.Do(req)
-		if err != nil {
-			return err
-		}
-		response, err := ioutil.ReadAll(resp.Body)
-		if err != nil {
-			return err
-		}
-		resp.Body.Close()
+	resp, err := client.Execute(u, requestData)
```
otoolep, line 45: Nit: remove this blank line, so style is consistent.

cmd/rqlite/execute.go
```diff
@@ -25,96 +26,70 @@ type executeResponse struct {
...
 
-		if resp.StatusCode == http.StatusMovedPermanently {
-			nRedirect++
-			if nRedirect > maxRedirect {
-				return fmt.Errorf("maximum leader redirect limit exceeded")
-			}
-			urlStr = resp.Header["Location"][0]
-			continue
-		}
+	if resp.StatusCode == http.StatusUnauthorized {
```
otoolep, line 55: Do you think this check adds anything, versus the `resp.StatusCode != http.StatusOK` immediately afterwards? I don't think it does, and the code is cleaner without it.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+var ErrTooManyRedirects = fmt.Errorf("maximum leader redirect limit exceeded")
+
+// HostChangedError indicates that the underlying request was executed on a different host
+// different from the caller anticipated
+type HostChangedError struct {
+	NewHost string
+}
+
+func (he *HostChangedError) Error() string {
+	return fmt.Sprintf("HostChanedErr: new host is '%s'", he.NewHost)
```
otoolep, line 26: HostChanedErr -- looks like a typo.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+type ConfigFunc func(*Client)
+
+// Client is a wrapper around stock http.Client that adds "retry on another host" behaviour
+// based on the supplied configuration.
+//
+// Errors that occur while performing requests are returned as they are, and the predicate to determine whether to check
```
otoolep, line 34: Nit: shorten this line a bit so it's roughly as long as the line ending in "behaviour" above.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// WithScheme changes the default scheme used i.e "http"
+func WithScheme(scheme string) ConfigFunc {
+	return func(client *Client) {
+		client.scheme = scheme
+	}
+}
+
+func WithPrefix(prefix string) ConfigFunc {
```
otoolep, line 78: Needs a godoc string.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+}
+
+// WithBasicAuth adds basic authentication behaviour to the client's request
+func WithBasicAuth(creds string) ConfigFunc {
+	return func(client *Client) {
+		client.creds = creds
+	}
+}
+
+func (c *Client) Query(url url.URL) (*http.Response, error) {
```
otoolep, line 98: Needs a godoc string.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+	return c.execRequest(http.MethodGet, url, nil)
+}
+
+func (c *Client) Execute(url url.URL, body io.Reader) (*http.Response, error) {
```
otoolep, line 102: Needs a godoc string.

cmd/rqlite/http/client.go
```diff
@@ -0,0 +1,186 @@
...
+	for triedHosts < len(c.hosts) {
+		host := c.hosts[c.currentHost]
+		url.Scheme = c.scheme
+		url.Host = host
+		urlStr := url.String()
+		resp, err := c.requestFollowRedirect(method, urlStr, body)
+
+		// If the status code is anything beside "service unavailable"
+		// we should propagate the error back to the caller
+		if resp != nil && resp.StatusCode != http.StatusServiceUnavailable {
```
otoolep, line 117: I'm not sure about this. Why not try another node if this node returns StatusServiceUnavailable? What's special about StatusServiceUnavailable?

cmd/rqlite/main.go
```diff
@@ -93,6 +95,12 @@ func main() {
 		}
 		term.Close()
 
+		hosts := createHostList(argv)
+		client := httpcl.NewClient(httpClient, hosts,
+			httpcl.WithScheme(argv.Protocol),
+			httpcl.WithBasicAuth(argv.Credentials),
```
otoolep, line 101: With....do you write Java by any chance? :-)

cmd/rqlite/query.go
```diff
@@ -80,83 +80,56 @@ type queryResponse struct {
...
 
-		if resp.StatusCode == http.StatusMovedPermanently {
-			nRedirect++
-			if nRedirect > maxRedirect {
-				return fmt.Errorf("maximum leader redirect limit exceeded")
-			}
-			urlStr = resp.Header["Location"][0]
-			continue
-		}
+	if resp.StatusCode == http.StatusUnauthorized {
```
otoolep, line 105: Same as earlier, not sure this check adds anything versus the check immediately afterwards.

cmd/rqlite/execute.go
chermehdi: I left it there in case you want to add a specific handler for that specific case, otherwise, if we are just logging back errors, it does not add anything.

If you are not planning on dealing with 403 errors separately, I can remove the check.

cmd/rqlite/main.go
chermehdi: AFAIK this is a pretty common Go pattern for writing `Builder` like structures, where you don't want to write a constructor for every possible combination of arguments.

Since I assumed the client can be used elsewhere, I went with this to be more flexible.

chermehdi: And Yes, I write Java from time to time :-)

chermehdi:
@otoolep What sort of test do you want me to run locally (to see the output)?

Another question: Does the project have end-to-end tests (to test client behavior)?

cmd/rqlite/execute.go
otoolep: No plans right now to deal with them separately, I say remove.

cmd/rqlite/main.go
otoolep: I haven't come across in the Go code I work with, though I recognize it. I'm fine with here.

otoolep:
There are no end-to-end tests for the CLI.

otoolep:
Thanks for the updates, there are still some unresolved comments.

cmd/rqlite/http/client.go
chermehdi: We want to try another server only if the node is unreachable, if the request itself is not well-formatted (and caused the server to return an error say `500`), trying on another server wouldn't be any different.

otoolep: I see your point. However I'm not sure if StatusServiceUnavailable is that special. I would need to check the code.

In fact, the issue associated with this PR states that the "try another node logic" should kick in if the first node "goes down". I interpret that to mean if the node is not reachable. But receiving any HTTP status means the node *is reachable* and we shouldn't try another node (except in the case of an explicit redirect). Does that make sense? Or do you think of it differently? If so, what your definition of "goes down"?

chermehdi: So I retested my hypothesis, I initially assumed that if a node is down the client would wrap the "connection refused" into something like ServiceUnavailable, this is not actually the case, if a server is down (killed, oomed, evicted ...) a connection refused error is going to be returned with a response of nil.

So I guess the check for the specific ServiceUnavailable is not required.

chermehdi: Another question that came to mind:

* Do we want the client to have retry on error behaviour? a maximum wait time per request?

otoolep: Open to suggestions, I also suggest we hardcode a sensible policy in. You can also come back to it later, a follow-up PR.

chermehdi: Let's do that in another PR

cmd/rqlite/execute.go
chermehdi: Done!

otoolep:
@chermehdi -- are you ready for the (hopeful) final review?

Also, can you paste an example session here of what happens when a node fails over?

chermehdi:
@otoolep Yes a final review would be okay at this stage.

Also, I am getting this weird thing (not sure if it's expected).

So I start a 3 node cluster, with node1 as the leader, I ctrl-c node1, so I would expect some other node else to become the leader, there is nothing in the logs that showcases this

```
rqlited -node-id 1 ~/rqlite/node.1
rqlited -node-id 2 -http-addr localhost:4003 -raft-addr localhost:4004 -join http://localhost:4001 ~/rqlite/node.2
rqlited -node-id 3 -http-addr localhost:4005 -raft-addr localhost:4006 -join http://localhost:4001 ~/rqlite/node.3
```

After stopping node 1 with ctrl-c

Node 3 logs
```
2022-01-12T21:40:32.009Z [ERROR] raft: failed to heartbeat to: peer=127.0.0.1:4002 error="dial tcp 127.0.0.1:4002: connect: connection refused"
^C2022-01-12T21:40:36.605Z [INFO]  raft: aborting pipeline replication: peer="{Voter 2 127.0.0.1:4004}"
2022-01-12T21:40:36.605Z [ERROR] raft: failed to appendEntries to: peer="{Voter 1 127.0.0.1:4002}" error="dial tcp 127.0.0.1:4002: connect: connection refused
...
```

otoolep:
Those logs are normal, leader is telling you the node is down. With 1 node down in a 3-node cluster, cluster should still work fine however.

chermehdi:
I expected something more obvious in the logs, like mentioning that the server now transitioned into a candidate and became the leader or became a follower (because node 2 became leader now).

Another question: are these logs from the raft library or from within rqlite itself?

otoolep:
Are you sure node 3 is not the leader? They sure look like logs from a leader node. 

Anyway, those logs are from the Raft layer, not rqlite.

Is the CLI working as you expect? Can you show me your manual testing?

cmd/rqlite/main.go
```diff
@@ -18,20 +18,22 @@ import (
...
 	cli.Helper
-	Protocol    string `cli:"s,scheme" usage:"protocol scheme (http or https)" dft:"http"`
-	Host        string `cli:"H,host" usage:"rqlited host address" dft:"127.0.0.1"`
-	Port        uint16 `cli:"p,port" usage:"rqlited host port" dft:"4001"`
-	Prefix      string `cli:"P,prefix" usage:"rqlited HTTP URL prefix" dft:"/"`
-	Insecure    bool   `cli:"i,insecure" usage:"do not verify rqlited HTTPS certificate" dft:"false"`
-	CACert      string `cli:"c,ca-cert" usage:"path to trusted X.509 root CA certificate"`
-	Credentials string `cli:"u,user" usage:"set basic auth credentials in form username:password"`
-	Version     bool   `cli:"v,version" usage:"display CLI version"`
+	Alternatives string `cli:"a,alternatives" usage:"comma separated list of 'host:port' pairs to use as fallback"`
```
otoolep, line 28: So I've never really liked the `Host` and `Port` parameters the CLI currently takes, I much prefer the version you choose. But I wasn't going to change it until the next major version. As it happens, I'm going to release v7 soon (the PR is in progress at https://github.com/rqlite/rqlite/pull/957), so this gives us a chance to fix up the CLI.

What do you think about just reworking the CLI options so that there is a single `host` parameter which takes a comma-delimited list of scheme:host:port values? E.g. 

`-host=http://localhost:4001,http://localhost:4003` (or maybe `-hosts`?)

If the scheme wasn't supplied, HTTP would be assumed. We would remove the existing `Protocol` parameter too. All other parameters would remain. It might make your code clearer too.

WDYT? I would make your updates part of the upcoming 7.0 release.

chermehdi: I didn't want to remove the Host: Port initially for backward compatibility reasons, I would be down for changing that to use a list of hosts to connect, but I prefer to do that as part of a separate PR.

chermehdi:
I added correct propagation of `HostChangeError`, this is the behavior of the session:

rqlite_test

otoolep:
OK, thanks for your contribution -- merged.

chermehdi:
Can you create another bug for the -hosts (I can take care of it)

otoolep:
https://github.com/rqlite/rqlite/issues/960 -- you can start the PR, but I won't merge until about to release 7.0.

Issue #958, opened 2022-01-13
spurious 500 errors on queries


nferch:
Have been seeing some seemingly random 500 errors to the `/db/query` endpoint, e.g.

"""
curl -n 'https://10.69.44.157:4001/db/query?q=SELECT id from table'                                                                   
json: error calling MarshalJSON for type *http.DBResults: no DB results set
"""

I have seen this on both SELECT and CREATE TABLE statements.

Any ideas as to what might be causing this?

From what I can tell from the code, this response occurs when neither the `ExecuteResult` or the `QueryRows` are populated in the `DBResults` (https://github.com/rqlite/rqlite/blob/b9db6e4903b426a16e7a25af0d23e8396477ff2c/http/service.go#L118). Unfortunately this error state is preventing any potential error from the `Cluster.Execute()` call from being surfaced.

Perhaps a code change like https://gist.github.com/4cf758659710b1d626c0feb2222eb2bf would fix this. Would like to try this out on my own cluster but it seems a little less than trivial to build a Docker container with local modifications.

Can hack at this a bit more when I have a little more time, but curious if I'm missing anything obvious.

otoolep:
Can I see the schema for the table?

nferch:
here's the two statements that I'm positive have failed:

"""
                CREATE TABLE IF NOT EXISTS usps_mailpieces (
                    date TEXT NOT NULL,
                    mailpiece_id TEXT)
"""

"""
SELECT *
FROM usps_mailpieces
WHERE date="2022-01-10" AND mailpiece_id="399131667388%27"
"""

otoolep:
No issues against a brand new node (that I can see):
```bash
~ $ curl -XPOST 'localhost:4001/db/execute?pretty&timings' -H "Content-Type: application/json" -d '[
>     "CREATE TABLE IF NOT EXISTS usps_mailpieces (date TEXT NOT NULL,mailpiece_id TEXT)"
> ]'
{
    "results": [
        {
            "time": 0.000226702
        }
    ],
    "time": 0.017660336
}~ $ curl -G 'localhost:4001/db/query?pretty&timings' --data-urlencode 'q=SELECT * FROM usps_mailpieces  WHERE date="2022-01-10" AND mailpiece_id="399131667388%27"'
{
    "results": [
        {
            "columns": [
                "date",
                "mailpiece_id"
            ],
            "types": [
                "text",
                "text"
            ],
            "time": 0.000113823
        }
    ],
[...]

PR #959, opened 2022-01-13
Return a clearer response if no DB results

otoolep:
This shouldn't happen, but it might, this will allow the underlying error to be returned to the caller. The body of the response will now look something like:

{
    "results": [],
    "error": "some error",
    "time": 0.021976516
}


otoolep:
https://github.com/rqlite/rqlite/issues/958

Issue #960, opened 2022-01-13
Change how CLI accepts hosts, ports, and protocol


otoolep:
Modify the CLI as described here: https://github.com/rqlite/rqlite/pull/947#discussion_r783630664 and make part of the upcoming 7.0 release.

PR #962, opened 2022-01-17
Move to simpler CLI host command-line arguments

otoolep:
https://github.com/rqlite/rqlite/issues/960


otoolep:
This is work-in-progress, may not end up like this.

PR #969, opened 2022-01-21
Setting node ID is now required

otoolep:


otoolep:
Scrap this. It makes running single-node clusters untidy.

PR #992, opened 2022-02-03
Raft uses hostnames, not IP addresses

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/991

Issue #744, opened 2021-01-30
Cluster tried to verify peer by IP but not my specified Hostname, hence valid SSL cert delcared as invalid


Ne00n:
I tried to run a 2 node cluster for testing with lets encrypt certificates.
All I end up with is bad certificate errors, precisely:

"x509: cannot validate certificate for <IP> because it doesn't contain any IP SANs"

./rqlited -http-addr server01.domain.net:4003 -http-cert /etc/letsencrypt/live/server01.domain.net/fullchain.pem -http-key /etc/letsencrypt/live/server01.domain.net/privkey.pem -raft-addr server01.domain.net:4004  -node-encrypt -node-cert /etc/letsencrypt/live/server01.domain.net/fullchain.pem -node-key /etc/letsencrypt/live/server01.domain.net/privkey.pem dataDir

./rqlited  -http-addr server02.domain.net:4003 -http-cert /etc/letsencrypt/live/server02.domain.net/fullchain.pem -http-key /etc/letsencrypt/live/server02.domain.net/privkey.pem -raft-addr server02.domain.net:4004 -node-encrypt -node-cert /etc/letsencrypt/live/server02.domain.net/fullchain.pem -node-key /etc/letsencrypt/live/server02.domian.net/privkey.pem -join https://server01.domain.net:4003 dataDir

Any idea?

otoolep:
This might be similar to this issue: https://github.com/rqlite/rqlite/issues/695

The hashicorp library works with IP addresses, and performs name resolution at certain places. I (somewhat) followed that pattern. I'll need to look into it to see what is up, and if I can address.

PR #993, opened 2022-02-04
Don't resolve any addresses

otoolep:
Resolving makes working with systems like Kubernetes difficult, as those systems change IP addresses, even if the hostnames are stable. rqlite should probably never have worked liked this. If IP addresses are still perferred they can be passed in explicitly as command line arguments.


otoolep:
https://github.com/rqlite/rqlite/issues/695

otoolep:
https://github.com/rqlite/rqlite/issues/744

otoolep:
https://github.com/rqlite/rqlite/issues/991

PR #1013, opened 2022-03-11
docs: add missing column

n8rb:
Add a missing column to an example API call.


otoolep:
Thank you for your contributions.

Issue #1020, opened 2022-05-12
Write throughput when globally distributed 


jmordica:
Using this as a place to further discuss the idea of having an option to allow rqlite cache write transactions for ~100ms in order to decrease the amount of round trips per transaction (which could significantly improve write throughout).

Are there any other thoughts since the last discussion about this?

If we come up with a solution I’d be willing to fund the development.

otoolep:
I'd like to understand the use case better. What data would you like to store such that you can tolerate the potential for data loss (node crashes before it sends the batch the Raft log)? And if you knew a write request that had been previously ack'ed by rqlite was not actually committed, how do you expect to recover?

jmordica:
Good questions:

1. Fast changing hot rows like user status/state. These types of transactions also doesn’t care as much about read after write consistency and can possibly suffer missing some transactions based on the likely hood that a new state/status change will happen again very soon (overwriting the previously missed state). One would normally use a kv store for something like this but the current distributed kv implementations are complex and require yet more infrastructure.
2. I’m not sure how to overcome the issue where an ackd request might not actually be written, however these types of transactions are somewhat understood to not need guaranteed consistency because they will likely be overwritten very soon after some state/status changes again.

I would just implement my own bulk requests using the API, but in a distributed environment where writes can happen from multiple clients, it seems like it would be better to handle the caching server side (rqlite), unless there are other workarounds I’m not thinking of.

otoolep:
OK, thanks, I recall us talking about this. I'll have to think about it, it would be a different API and behaviour to what rqlite currently exposes. On the surface it mightn't be too difficult, but it would have to be maintained going forward. Long term, that can be the higher cost. So I can't promise I'll add this to rqlite at the present time, but I'm happy to leave the issue open for discussion.

otoolep:
If you can tolerate data loss, have you looked at https://litestream.io/ ?

otoolep:
Another question -- still considering this. What do you want to happen if some of the requests failed, but others go through? If a bulk request, created locally by rqlite, has more than 1 SQL INSERT, some may fail, some may success. So you could even have a partial success, which again would be difficult to track. Do you care? Probably not.

otoolep:
Slept on this feature request, more appealing to me now. :-)

I'd just think about semantics a little more. I presume a new API would be OK for this? Something like this:

`localhost:4001/db/bulk`

[...]

PR #1024, opened 2022-05-16
Support queued writes

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1020

PR #1056, opened 2022-07-21
WIP creds forwarding

otoolep:


otoolep:
Oh, and proper handling of failed authorization from the remote node, so that the local node returns 401 Unauthorized.

Issue #1051, opened 2022-07-15
Authorization not enforced for inter-node comms


otoolep:
rqlite security is controlled as described [here](https://github.com/rqlite/rqlite/blob/master/DOC/SECURITY.md).

Security can be configured such that a node requires credentials to join a cluster (since joining uses the HTTP API). However all further requests by the node which joined to any other node will use the inter-node comms link, which does not enforce authorization. This means changes could be made to the cluster via node that has joined, but which doesn't have a local Security config file.

The fix, in general terms, is to enforce authorization, even for the inter-node comms channel.

otoolep:
This should be fairly straightforward to do:

- Add a new `Credentials` protobuf object to https://github.com/rqlite/rqlite/blob/master/cluster/message.proto. The credentials will be "username" and "password", both strings.
- Add the `Credentials` object to `Command` (see https://github.com/rqlite/rqlite/blob/master/cluster/message.proto#L12)
- Then copy the credentials from a HTTP request to the proto object, when a request is forwarded to another node in the cluster. Presumably this will require changes to https://github.com/rqlite/rqlite/blob/master/cluster/client.go
- Adding permission checking in the Cluster Service code (at https://github.com/rqlite/rqlite/blob/master/cluster/message.proto#L12) similar to how it's done currently in the HTTP layer (see https://github.com/rqlite/rqlite/blob/master/http/service.go#L1247). This means a CredentialStore object will need to be passed into the Cluster Service object, like what happens with HTTP today.
- Come up with a way to signal the error back, so the HTTP layer in the first node can respond with a 403 if the other node returns "no authorization".

New unit tests needed to be added at each layer.

ngharrington:
Thanks @otoolep happy to take a look a this -- these notes are great and align with what I had in mind. Hopefully will spend some time this evening (EST), but certainly will look this week.

otoolep:
There is one tricky part to this -- queued writes. Because queued writes happen asynchronously, the context of which user sent the original HTTP request is not passed to the queue. So that part is going to need some thinking.

You can see the queued write implementation in the HTTP layer.

otoolep:
Queued writes are called "queued executes" in the code.

otoolep:
[...]

PR #1057, opened 2022-07-24
WIP TCP service auth

ngharrington:


cluster/service.go
```diff
@@ -167,6 +177,23 @@ func (s *Service) serve() error {
 	}
 }
 
+func (s *Service) checkCommandPerm(c *Command, perm string) bool {
+	if s.credentialStore == nil {
+		return true
+	}
+
+	var username string
```
otoolep, line 185: You could remove a block in your `if` statement by simply setting username and password to "" to begin with. I.e.
```Go
username := ""
password := ""
if c.Credentials != nil {
                 username = c.Credentials.GetUsername()
		password = c.Credentials.GetPassword()
}
```

otoolep:
Looks like you're on the right track.

ngharrington:
@otoolep I feel pretty good about the functionality here and the new tests. The last major thing I want to do is remove the username and password fields from the Query and Execute methods on the Client in favor of an internal Credentials field in the Client struct. Definitely interested in any additional feedback you may have as well at this point.

ngharrington:
On second thought, not going to make the credentials change I described above. Since the client instance is part of the http service struct, it's not ideal to change the credentials in that instance for each request.

Pending further review I'm comfortable with the state this is in now.

cluster/service.go
```diff
@@ -211,9 +235,10 @@ func (s *Service) handleConn(conn net.Conn) {
 			stats.Add(numExecuteRequest, 1)
 
 			resp := &CommandExecuteResponse{}
-
 			er := c.GetExecuteRequest()
-			if er == nil {
+			if !s.checkCommandPerm(c, auth.PermExecute) {
+				resp.Error = "Unauthorized"
+			} else if er == nil {
```
otoolep, line 241: Style thing, I realise, but I think the `er == nil` check should go first.

cluster/service.go
```diff
@@ -243,7 +268,9 @@ func (s *Service) handleConn(conn net.Conn) {
 			resp := &CommandQueryResponse{}
 
 			qr := c.GetQueryRequest()
-			if qr == nil {
+			if !s.checkCommandPerm(c, auth.PermQuery) {
+				resp.Error = "Unauthorized"
+			} else if qr == nil {
```
otoolep, line 273: Ditto for `qr`.

cluster/service_test.go
```diff
@@ -170,6 +171,84 @@ func Test_NewServiceSetGetNodeAPIAddrTLS(t *testing.T) {
...
 }
 
+func Test_NewServiceTestExecuteQueryAuthNoCredentials(t *testing.T) {
+	ml := mustNewMockTransport()
+	db := mustNewMockDatabase()
+
+	// Test that for a cluster with no credential store configed
+	// all users are authed for both operations
+	var c CredentialStore = nil
+	t.Log(c)
```
otoolep, line 181: Why are you logging credentials during testing? Not a big deal, just wondering if you have a specific reason.

http/service.go
```diff
@@ -1091,7 +1096,17 @@ func (s *Service) execute(w http.ResponseWriter, r *http.Request) {
...
 		}
-		results, resultsErr = s.cluster.Execute(er, addr, timeout)
+
+		username, password, ok := r.BasicAuth()
+		if !ok {
+			username = ""
+		}
+
+		results, resultsErr = s.cluster.Execute(er, addr, username, password, timeout)
+		if resultsErr != nil && resultsErr.Error() == "Unauthorized" {
```
otoolep, line 1106: Checking if `Error` is equal to a specific string is obviously brittle. I wonder if we should add a dedicated flag to the `resultsErr` struct, indicating "no authorization". WDYT?

http/service.go
```diff
@@ -1180,7 +1195,15 @@ func (s *Service) handleQuery(w http.ResponseWriter, r *http.Request) {
 			http.Error(w, ErrLeaderNotFound.Error(), http.StatusServiceUnavailable)
 			return
 		}
-		results, resultsErr = s.cluster.Query(qr, addr, timeout)
+		username, password, ok := r.BasicAuth()
+		if !ok {
+			username = ""
+		}
+		results, resultsErr = s.cluster.Query(qr, addr, username, password, timeout)
+		if resultsErr != nil && resultsErr.Error() == "Unauthorized" {
```
otoolep, line 1203: Ditto.

otoolep (commented):
Looks very close to what I would do -- thanks very much. I've got some minor comments.

I'm also traveling until early August, so may be slow responding.

cluster/service_test.go
ngharrington: Just a holdover from some debugging.  Will clean up.

http/service.go
ngharrington: Agreed, sort of anticipated this feedback :).

cluster/service.go
ngharrington: I'm definitely fine with changing this and will (not dogmatic about it in the least) but I am curious about why you prefer that choice (even stylistically). My thinking here is basically return auth failures as soon as you can, but admittedly I didn't actually do completely that because I wanted to put this check in the existing if/else block rather than do the check before we even extract the ExecuteRequest.

otoolep: Two reasons.

- I like to check fundamental assumptions first. If the `er` is nil, something has gone fundamentally wrong with the call.
- Secondly, this code reminds me of Go error checking, and the `if err == nil` pattern. So to my eyes, it looks more Go-like to check a nil case first.

Again, just style.

http/service.go
ngharrington: Added a lot more structure around the errors returned from the TCP service.  Let me know what you think/if you may prefer to take a different approach.

cluster/message.proto
```diff
@@ -22,14 +27,26 @@ message Command {
...
+        RESULTS_ERROR_UNAUTHORIZED = 1;
+        RESULTS_ERROR_NIL_QUERY_REQUEST =2;
+    }
+    Type type = 1;
+    string error = 2;
 }
 
 message CommandExecuteResponse {
-	string error = 1;
+	ResultsError error = 1;
```
otoolep, line 45: You can't do this, it's not backwards-compatible -- it would prevent this software running with, say, nodes running 7.5. You need to add a new field. I suggest a simple Boolean field to `CommandExecuteResponse` and `CommandQueryResponse` , indicating the "unauthorized" condition. 

It means you'll need to check two conditions -- `Error != ""` or "Unauthorized`. If either is true, return an error.

cluster/service_test.go
```diff
@@ -170,6 +171,82 @@ func Test_NewServiceSetGetNodeAPIAddrTLS(t *testing.T) {
...
+	if err != nil {
+		t.Fatal(err)
+	}
+}
+
+func Test_NewServiceTestExecuteQueryAuth(t *testing.T) {
+	ml := mustNewMockTransport()
+	db := mustNewMockDatabase()
+
+	// Test that for a cluster with a credential store configed
```
otoolep, line 211: configed -> configured

otoolep (changes requested):
This looks good. There is one final change that must be made -- once done, I'll merge.

otoolep:
Actually, I will merge this and make some follow-up changes.

Thanks for your contribution @ngharrington

otoolep:
Most of the work towards fixing: https://github.com/rqlite/rqlite/issues/1051

PR #1059, opened 2022-08-06
Remote auth follow-on

otoolep:
The proto was not well designed in the first place, but it's better to deal with it as-is, and wait until another major release to fix it.

Follow up to https://github.com/rqlite/rqlite/pull/1057


otoolep:
https://github.com/rqlite/rqlite/issues/1051

PR #1060, opened 2022-08-09
fix minor unreachable code caused by t.Fatal

Abirdcfly:
https://pkg.go.dev/testing#T.Fatalf
> Fatalf is equivalent to Logf followed by FailNow.

https://pkg.go.dev/testing#T.FailNow
> FailNow marks the function as having failed and stops its execution by calling runtime.Goexit (which then runs all deferred calls in the current goroutine).


otoolep:

PR #1071, opened 2022-10-06
fixed typo: flag is -node-no-verify, not -no-node-verify

fabiant7t:
There is a typo in the security documentation. The flag `-no-node-verify` is misspelled and should be corrected to `-node-no-verify`.

See
```
cmd/rqlited/flags.go:	flag.BoolVar(&config.NoNodeVerify, "node-no-verify", false, "Skip verification of a remote node cert")
```


otoolep:
Thanks for the correction.

Issue #1073, opened 2022-10-13
API should support nil as parameter


otoolep:
https://rqlite.slack.com/archives/C0324J4LC91/p1665626647783139

otoolep:
It supports it fine, actually. Looks like a Go client issue.

otoolep:
https://rqlite.slack.com/archives/C0324J4LC91/p1665629560821889

otoolep:
Actually, may not quite be right.

otoolep:
```bash
$ curl -XPOST 'http://kotsadm:password@10.43.141.224:4001/db/execute?pretty&timings' -H "Content-Type: application/json" -d '[
    ["INSERT INTO test(name, value) VALUES(?, ?)", "fiona", null]
```
]'
unsupported type

sgalsaleh:
Note: this is only an issue with parameter binding. Title can be changed to something like:

"API should support null values when using parameter binding"

PR #1074, opened 2022-10-13
Support nil as a paramterized value

otoolep:


otoolep:
Fix for https://github.com/rqlite/rqlite/issues/1073

PR #1093, opened 2022-10-26
test: use `T.TempDir` to create temporary test directory

Juneezee:
A testing cleanup. 

This pull request replaces `ioutil.TempDir` with `t.TempDir`. We can use the `T.TempDir` function from the `testing` package to create temporary directory. The directory created by `T.TempDir` is automatically removed when the test and all its subtests complete. 

Reference: https://pkg.go.dev/testing#T.TempDir

```go
func TestFoo(t *testing.T) {
	// before
	tmpDir, err := ioutil.TempDir("", "")
	if err != nil {
		panic(err)
	}
	defer os.RemoveAll(tmpDir)

	// now
	tmpDir := t.TempDir()
}
```

### DIFF
--- a/db/db_test.go
+++ b/db/db_test.go
@@ -21,8 +21,7 @@ import (
  */
 
 func Test_DbFileCreation(t *testing.T) {
-	dir, err := ioutil.TempDir("", "rqlite-test-")
-	defer os.RemoveAll(dir)
+	dir := t.TempDir()
 	dbPath := path.Join(dir, "test_db")
 
 	db, err := Open(dbPath, false)
--- a/http/service_test.go
+++ b/http/service_test.go
@@ -1139,7 +1139,7 @@ func Test_TLSServce(t *testing.T) {
 	m := &MockStore{}
 	c := &mockClusterService{}
 	var s *Service
-	tempDir := mustTempDir()
+	tempDir := t.TempDir()
 
 	s = New("127.0.0.1:0", m, c, nil)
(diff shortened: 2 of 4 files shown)


otoolep:
Thank you -- this looks like a nice change. Any idea why the test is failing on Windows however? Perhaps the OS has nothing to do with it, and it's just racy.

https://ci.appveyor.com/project/otoolep/rqlite/builds/45189990#L1528

Juneezee:
> Thank you -- this looks like a nice change. Any idea why the test is failing on Windows however? Perhaps the OS has nothing to do with it, and it's just racy.
> 
> https://ci.appveyor.com/project/otoolep/rqlite/builds/45189990#L1528

@otoolep This PR manages to uncover some hidden issues of the tests on Windows :smile: Windows does not allow removing an open file (https://github.com/golang/go/issues/50510#issuecomment-1008535852). If we don't close the `os.File` properly, `t.TempDir()` will fail to do the cleanup and cause the test to fail

```go
func TestFoo(t *testing.T) {
	tmpDir := t.TempDir()

	file, err := os.CreateTemp(tmpDir, "dumpwaltest*.txt")
	if err != nil {
		t.Fatal(err)
	}

	// On Windows, the test will fail if `file` is not closed because
	// t.TempDir() will not be able to cleanup the temporary directory.
	defer file.Close()
}
```

otoolep:

Issue #1103, opened 2022-11-01
Random database disk image is malformed errors


jvarruda:
Can you help with this case where randomly some queries are ruturning with the `{"error": "database disk image is malformed"}`  error?

**What version are you running?**
v7.6.1

**Are you using Docker or Kubernetes to run your system?**
No

**Are you running a single node or a cluster?**
Cluster with 3 nodes.

**What did you do?**
A python application using sqlalchemy-rqlite and pyrqlite performing simple SELECT query in just one table.

**What did you expect to happen?**
Get the result back from the query.

**What happened instead?**
Randomly get the following error: {"error": "database disk image is malformed"}

**Please include the Status, Nodes, and Expvar output from each node (or at least the Leader!)**

Status:
```
curl 192.168.100.116:4001/status?pretty
{
    "build": {
        "branch": "master",
        "build_time": "2022-08-07T09:23:22-0400",
        "commit": "58568a4577940bf3adf8794c9e6e977a36e9e5db",
        "compiler": "gc",
        "version": "v7.6.1"
    },
    "cluster": {
        "addr": "192.168.100.116:4002",
        "api_addr": "192.168.100.116:4001",
        "https": "false"
    },
    "http": {
        "auth": "disabled",
        "bind_addr": "192.168.100.116:4001",
        "cluster": {
            "conn_pool_stats": {
                "192.168.100.114:4002": {
                    "idle": 4,
                    "max_open_connections": 64,
                    "open_connections": 4
                },
                "192.168.100.189:4002": {
                    "idle": 3,
                    "max_open_connections": 64,
                    "open_connections": 4
                }
            },
            "local_node_addr": "192.168.100.116:4002",
            "timeout": 30000000000
        },
        "queue": {
            "_default": {
                "batch_size": 16,
                "max_size": 128,
                "sequence_number": 0,
                "timeout": 50000000
            }
        }
    },
    "last_backup_time": "2022-11-01T03:00:04.12518951Z",
    "node": {
        "current_time": "2022-11-01T16:41:43.449966112Z",
        "start_time": "2022-10-28T17:26:47.632592627Z",
        "uptime": "95h14m55.817374557s"
    },
    "os": {
        "executable": "/usr/bin/rqlited",
        "hostname": "nodew-2",
        "page_size": 4096,
        "pid": 14912,
        "ppid": 14911
    },
    "runtime": {
        "GOARCH": "amd64",
        "GOMAXPROCS": 32,
        "GOOS": "linux",
[...]

PR #1110, opened 2022-11-09
Disk image malformed

otoolep:


otoolep:
Unit test to reproduce https://github.com/rqlite/rqlite/issues/1103

PR #1135, opened 2022-12-02
Ignore explicit join request if there is preexisting state

otoolep:


otoolep:
Test fails because there is a need to join if ID or address changes. Perhaps the bug is to ignore self-joins. But is that an error? Scenario may be: 3-node cluster. Kill leader. Node 2 becomes leader, then restart it with same command line params. It will try to join itself, it it comes back up as leader.

Issue #687, opened 2020-10-17
Accessing rqlite HTTP via jQuery, got CORS error


brailateo:
Access to XMLHttpRequest at 'http://10.0.0.1:4001/db/execute?timings' from origin 'http://localhost' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource.

It would be nice to be able to access the HTTP API via jQuery directly, from JavaScript in a browser, avoiding any other nodejs or php servlets.
I was looking for some command-line parameters that would fix that but did not find any!

Teo

otoolep:
I am not super-knowlegeable in this area (though I know what CORS is about). What is the recommended way to fix this?

brailateo:
I think that it would be simpler to use: https://github.com/rs/cors

Some examples are here: https://flaviocopes.com/golang-enable-cors/

Teo

patdx:
I don't think that code will work in the browser, while `mode: 'no-cors',` is no-op in `node-fetch`, in the actual browser it prevents the Javascript code from reading `response.json()` so you can't do anything with the response.

Ultimately CORS is a security mechanism implemented by browsers, so there is not some easy loophole to circumvent it in the browser. The solution needs to come from the server side.

That said, I'm not sure exactly what the correct solution should look like.

I've just run into this issue as I want to debug some responses using browser dev tools locally, but not currently planning to expose an rqlite to the public internet and consume the response from browser application, so don't really need so many fancy settings, but would appreciate a quick way to let me connect to rqlite HTTP API from a local browser.

mikroskeem:
rqlite simply needs to pass on CORS headers with its responses. I'm personally using rqlite for toy frontend applications & small workshops, being able to throw Caddy out from the deployment would make rqlite usage a lot more convenient.

That being said, are there any plans to implement CORS headers support?

otoolep:
Would it be sufficient to simply add a command-line option to `rqlited`, allowing you to specify the value to be set for the `Access-Control-Allow-Origin` HTTP header?

E.g. you would pass, say, this when launching `rqlited`: `-http-allow-origin="https://developer.mozilla.org"` and then all HTTP responses would include the following HTTP header in the response:

`Access-Control-Allow-Origin: https://developer.mozilla.org`

otoolep:
In action:
```bash
rqlited -http-allow-origin=https://rqlite.io data
```

Then we get this:
```bash
[...]

PR #1168, opened 2023-02-26
Simple support for Access-Control-Allow-Origin

otoolep:
Address https://github.com/rqlite/rqlite/issues/687 in a basic manner.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 7.14.0 (unreleased)
 ### New features
-- [PR #1168](https://github.com/rqlite/rqlite/pull/1168): Support for setting Access-Control-Allow-Origin. Fixes [issue #687](https://github.com/rqlite/rqlite/issues/687).
+- [PR #1168](https://github.com/rqlite/rqlite/pull/1168): Support for setting Access-Control-Allow-Origin header in HTTP responses. Fixes [issue #687](https://github.com/rqlite/rqlite/issues/687).
 
 ## 7.13.2 (February 23rd 2023)
 ### Implementation changes and bug fixes
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## 7.14.0 (unreleased)
 ### New features
-- [PR #1168](https://github.com/rqlite/rqlite/pull/1168): Support for setting Access-Control-Allow-Origin header in HTTP responses. Fixes [issue #687](https://github.com/rqlite/rqlite/issues/687).
+- [PR #1168](https://github.com/rqlite/rqlite/pull/1168): Support for setting `Access-Control-Allow-Origin` header in HTTP responses. Fixes [issue #687](https://github.com/rqlite/rqlite/issues/687).
 
 ## 7.13.2 (February 23rd 2023)
 ### Implementation changes and bug fixes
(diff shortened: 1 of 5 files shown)


mikroskeem:
I believe it'd be useful to set `Access-Control-Allow-Methods: GET, POST` & `Access-Control-Allow-Headers: Content-Type` ([see why Content-Type should be set explicitly](https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header)) because we can send queries using both GET & POST, and `Access-Control-Allow-Credentials: true` if basic authentication is enabled.

In addition you'd need to add support for `OPTIONS` method on any of the endpoints - that's for [preflight request](https://developer.mozilla.org/en-US/docs/Glossary/Preflight_request). Replying with status 204 & CORS headers is sufficient.

otoolep:
@mikroskeem -- thanks for the pointers. Just to be clear, even if ensure rqlite sets all the right headers such that it "supports CORS", it seems like `Access-Control-Allow-Origin` is the only one that needs to be user-configurable, right? Or perhaps the user might want to also control which methods can be invoked by the "origin"?

mikroskeem:
Being able to configure origin will do just fine for now.

I believe that if anyone wants to configure anything more complex like what you described, they'd go for reverse proxy.

patdx:
I made a little sandbox to test out different variations of CORS here:

- https://github.com/patdx/rqlite-sandbox
- https://patdx.github.io/rqlite-sandbox/rqlite-sandbox.html

It's a standalone HTML file so you can download it and double click it to run locally from your computer too. If both the HTTP GET and HTTP POST actions can work with authentication and without errors then I think that should be reasonable.

> it seems like Access-Control-Allow-Origin is the only one that needs to be user-configurable, right?

I agree this would fit the kinds of light uses I had in mind too.

-----

As far as I can understand the `Access-Control-Allow-Credentials: true` is for using the browser's credential store, kind of similar to how the browser can remember cookies. It seems reasonable to allow it, but I'm not sure if there is a good way to make use of it with rqlite at the moment. I think the server needs to challenge the user for credentials via a [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) header but I don't think rqlite does this? It just returns a plain 401 Unauthorized from what I can see.

image

More info: https://www.sjoerdlangkemper.nl/2018/09/12/authorization-header-and-cors/

For credentials provided by setting the `Authorization` header directly in JavaScript, whether or not I set `Access-Control-Allow-Credentials: true`, It was still necessary add Authorization to the Access-Control-Allow-Headers to get it to work.

For **Access-Control-Allow-Methods**, I didn't find this necessary. It seems like GET, HEAD, POST may be allowed by default? https://stackoverflow.com/a/44385327 But based on the documentation it does seem like they would prefer you to set it explicitly.

----

Conclusion:

In my test I ended up needing the following headers to get it to work: https://github.com/patdx/rqlite-sandbox/blob/main/Caddyfile

```
Access-Control-Allow-Credentials: true
Access-Control-Allow-Origin: {{user input here}}
Access-Control-Allow-Headers: authorization, content-type
Access-Control-Allow-Methods: GET, POST
```

Preflight OPTIONS request support is also necessary. In particular for `/db/execute` and `/db/query` but also likely for other URLs.

Issue #1167, opened 2023-02-26
[Question] Clarification on securing inter-node raft communication


otto-dev:
I've reviewed the [Security guide](https://rqlite.io/docs/guides/security/) and would like to confirm my understanding. By default, I enable mutual TLS on all services as a best practice for blocking malicious actors within the same data-center. However, since rqlite doesn't currently support mutual TLS, I need to reconsider my security approach.

Specifically, while I can secure HTTPS traffic via Baisc Auth, I'm concerned about securing inter-node raft communication. If an attacker within the same data-center connects to the unsecured raft ports with a rogue version of rqlite, how much access can they gain? Is it possible (and necessary) to prevent such connections by setting `-node-ca-cert` to a CA that only I have access to?

otoolep:
I am interested in implementing mutual TLS, rqlite security is important to me. The current TLS support in rqlite is mostly about preventing eavesdropping of internode traffic. To prevent rogue nodes connecting to the Raft port I suggested network-based protections, as you probably saw in the guide.

A rogue node could probably cause some damage by connecting to the Raft port -- in fact any program could send TCP traffic to the port. That's why I suggest network-based security (only allowing connections from specific IP addresses, for example).

However I think your request is valid -- and presumably it's easy enough to add to rqlite. Let me look into it.

otoolep:
After some quick study of the Go TLS docs, I think I can add mutual-TLS support fairly easily.

@otto-dev -- can you build rqlite yourself and deploy, if I push changes to the `master` branch? Otherwise I can add the functionality I think, and make a candidate release which you can test. Or are you running Docker images?

otoolep:
I would add the support for both the HTTP server (port 4001) *and* inter-node communications (port 4002).

otoolep:
For my own reference: https://venilnoronha.io/a-step-by-step-guide-to-mtls-in-go

otto-dev:
Thank you so much for taking the time to investigate this and for considering my request. I also wanted to share some high-level thoughts on the implementation.

Regarding mutual TLS for raft communication, it should be possible to add it transparently. What I mean is that making it the new default when `-node-cert` is provided shouldn't cause any issues for existing users, since -node-cert is already trusted by either -node-ca-cert or system defaults. By using the certificates and CA in both directions, nothing should break.

[...]

PR #1171, opened 2023-03-05
Initial TLS utilities

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1167

otto-dev:
```diff
3,4d2
< -http-cert-client: when a node joins a cluster (which is done via HTTP), it will present this x509 cert, if set
< -http-key-client: when a node joins a cluster via HTTPS, it will use this private x509 key
5a4
> -http-verify-client: Enables mutual TLS for HTTPS, requiring clients to present a certificate signed by http-ca-cert. Nodes will present -http-ca-cert when joining a cluster (done via HTTP).
10,11d8
< -node-cert-client: when a node connects to other nodes it will present this x509 cert, if set
< -node-key-client: when a node connects to other nodes it will use this private x509 key
12a10
> -node-verify-client: Enables mutual TLS for node-to-node communication using -node-ca-cert and -node-cert
```

Couple initial thoughts before testing: No problem with how it is, but it may be possible to simplify further. Since the nodes are uniform in their roles, and all nodes need to be able to connect to each other in either direction, I think a distinction between client/server certificate among nodes is not required. `-http-verify-client` and `-node-verify-client` would have a dual effect each:
- presenting the certificate in client connections
- validating the certificate in server connections

otoolep:
@otto-dev -- I see what you mean now, just allowing the CA to be set for each node (both for client and server verification) and just supporting one cert/key pair per node (at least for internode comms) does make sense to me, now that I see it. I'll simplify as you suggest.

otoolep:
@otto-dev -- this is what I'm proposing now:

`-http-cert`: path to HTTPS server x509 certificate. This is the cert that is presented to clients that connect, as well as being the cert this node will present when authenticating as a client to other nodes' HTTP API.
`-http-key`: path to HTTPS private x509 key
`-http-ca-cert`: CA cert that a node will use to verify all HTTPS certificates it receives during TLS handshake
`-http-no-verify`: node will not verify any HTTPS certificate
`-node-encrypt`: must be set to true to enable node-to-node encryption.
`-node-cert`: path to node's x509 certificate. This is the cert that is presented to nodes that connect on the Raft interface, as well as being the cert this node will present when authenticating as a client to other nodes' Raft network interface.
`-node-key` path to node's private x509 key
`-node-ca-cert`: CA cert that a node will use to verify inter-node (Raft) communications it receives during TLS handshake
`-node-no-verify`: node will not verify any nodes presented during inter-node communications.

WDYT?

otto-dev:
Thank you.

**HTTP**: For HTTP, I suggest adding a flag like `-http-verify-client` to enable HTTP mTLS (off by default), which would maintain backwards compatibility,  allow for Basic Auth + TLS (server-only) setups, and make it easier to play with rqlite during development and with third-party tooling.

**raft**: Looks good. I think enabling mTLS by default for `raft` is a bold but good idea. I can't think of a scenario where doing so breaks backwards compatibility for existing configurations, again for the reason of my previous comment, that directionality of the TLS connection between nodes "does not matter" . If you wanted to be super safe, you could add an enable flag here as well.

Sorry I didn't have the chance to test your good work in detail yet, but I will do so as soon as I have the opportunity.

otto-dev:
Great work! Thank you so much for this.

I have tested the most recent commit (11ed5731d699084f5b5c976dd6c520d51dd08a49) and see mTLS working properly. However, I found thefollowing minor things that need fixing:
- [this fix](https://github.com/rqlite/rqlite/pull/1171#issuecomment-1459149855)
- `-node-no-verify` and `-http-no-verify` no longer work. I see `x509: certificate signed by unknown authority`
- There is no way to disable HTTP mTLS if `-http-cert`/`-http-key` are set. I think a flag such as `-http-verify-client` to enable mTLS (off by default) should be added. Unless you are happy to force mTLS for all HTTP clients, of course.

otoolep:
Thanks for the feedback, @otto-dev -- let look into all this.

otoolep:
Thanks for the `ForceAttemptHTTP2` fix. I reproduced the issue for joining, but not for notifying, so didn't add it to `bootstrap.go`. I added unit tests to show all this: https://github.com/rqlite/rqlite/pull/1171/commits/89c3dfaa2e9d3df661a994d022ac3ae47cd735d6

Without adding ``ForceAttemptHTTP2` to `join.go` `Test_SingleJoinHTTPSOK` will fail. However `Test_BootstrapperBootSingleNotifyHTTPS` passes without changing the bootstrapping code.

otoolep:
@otto-dev -- I think everything is good now. Want to give it a test?

otto-dev:
Wow, amazing work! Thanks a ton for this. I only found one minor issue that can be fixed with a one-liner.

Manual tests:
- Cluster is reported up in stdout
- Can execute `SELECT 1` via HTTP API

Raft
- :heavy_check_mark:  Can connect without TLS
- :heavy_check_mark:  Can connect with mTLS
- :heavy_check_mark:  Rejects untrusted CA of peer node
- :heavy_check_mark:  Allows untrusted CA of peer node with -node-no-verify

HTTP
- :heavy_check_mark:  Can connect without TLS
- :red_square: Can connect with TLS. Not working, see error and fix below. This is on the latest commit ff31decd7958bfd1e0e2c805f17ae736f23c4f8f. I'm not sure why the test passes but at runtime there is a failure. 
- :heavy_check_mark:  Does not request client certificate without http-verify-client
- :heavy_check_mark:  Can connect with mTLS _(only with the fix below)_
- :heavy_check_mark:  Requests client certificate with http-verify-client
- :heavy_check_mark:  Rejects untrusted CA of peer node
- :heavy_check_mark:  Allows untrusted CA of peer node with -http-no-verify


HTTP TLS error:
```
[cluster-bootstrap] 2023/03/09 10:51:35 failed to notify all targets: [http://localhost:4001 http://localhost:4003] (failed to post notification to node at https://localhost:4001/notify: Post "https://localhost:4001/notify": net/http: HTTP/1.x transport connection broken: malformed HTTP response "\x00\x00\x18\x04\x00\x00\x00\x00\x00\x00\x05\x00\x10\x00\x00\x00\x03\x00\x00\x00\xfa\x00\x06\x00\x10\x01@\x00\x04\x00\x10\x00\x00", will retry)
```

HTTP TLS error fix:
```diff
diff --git a/cluster/bootstrap.go b/cluster/bootstrap.go
index 79a4e09..47b3ea4 100644
--- a/cluster/bootstrap.go
+++ b/cluster/bootstrap.go
@@ -135,6 +135,7 @@ func (b *Bootstrapper) notify(targets []string, id, raftAddr string) error {
        // Create and configure the client to connect to the other node.
        tr := &http.Transport{
                TLSClientConfig: b.tlsConfig,
+               ForceAttemptHTTP2: true,
        }
        client := &http.Client{Transport: tr}
```

Then everything works as expected. Great work!

otoolep:
@otto-dev -- one final change. I decided to disable mutual TLS for node-to-node by default. This maximise backwards compatibility, just in case any current users are relying on the existing behaviour. To enable mutual TLS for node-to-node communication, pass `-node-verify-client` to `rqlited`.

otoolep:
That is odd, I can't repro it with my unit test. The unit test must be faulty, let me see what I can do.

otoolep:
@otto-dev -- found the issue, faulty test. It was properly replicating what happens in production code. Everything should be good now.

otoolep:
I'm going to merge this, since it has the fix you requested. Would be great if you could test a build from the master branch, just to be sure everything is OK.

otto-dev:
Yes, everything works on master. Great work!

Raft
- :heavy_check_mark:  Can connect without TLS
- :heavy_check_mark:  Can connect with TLS
- :heavy_check_mark:  Rejects untrusted CA of peer node
- :heavy_check_mark:  Allows untrusted CA of peer node with -node-no-verify
- :heavy_check_mark:  Rejects untrusted CA of peer node if -node-verify-client
- :heavy_check_mark:  Can connect with mTLS
- :heavy_check_mark:  Requests client certificate with node-verify-client

HTTP
- :heavy_check_mark:  Can connect without TLS
- :heavy_check_mark:  Can connect with TLS 
- :heavy_check_mark:  Does not request client certificate without http-verify-client
- :heavy_check_mark:  Can connect with mTLS
- :heavy_check_mark:  Requests client certificate with http-verify-client
- :heavy_check_mark:  Rejects untrusted CA of peer node
- :heavy_check_mark:  Allows untrusted CA of peer node with -http-no-verify

otoolep:
@otto-dev -- I've released 7.14.0, with full support for mutual TLS. 

Can you tell me more about what you're using rqlite for? I really like to hear about use cases. Feel free to drop me an email (address in my profile) if you prefer.

- https://github.com/otoolep
- https://www.philipotoole.com/contact/

PR #1224, opened 2023-04-24
Fixed credentials load error checking

phmx:


otoolep:
You're right, thanks for the fix @phmx

Issue #1157, opened 2023-02-03
Support for RETURNING in api response


alexgreendesign:
Thanks so much for your great work on RQLite - it's fantastic.

Would it be possible to add support for RETURNING in the api response?

For example, a query like:
`DELETE FROM table WHERE column = 'value' RETURNING other_column;`
Should return other_column. 

It does return last_insert_id, which in the case of inserts can be followed by a SELECT to get the data as a workaround, but this doesn't work for deletes.

otoolep:
Thanks for your report.

What do you mean it isn't supported? Can you show me using actual API calls what happens, and what you expect? To address this issue as fast as possible, a full reproduction case would be best. I.e. create your table, do your query, show me what you get, and explain what you actually want. I would need to see all of this in, say, the form of `curl` requests (or at least as actual rqlite shell session).

https://rqlite.io/docs/api/api/ for examples on using `curl`.

alexgreendesign:
Sure:

```
curl -XPOST 'localhost:4001/db/execute?pretty' -H "Content-Type: application/json" -d '[
    "CREATE TABLE _test (id INTEGER NOT NULL PRIMARY KEY, name TEXT)"
]'

curl -XPOST 'localhost:4001/db/execute?pretty' -H "Content-Type: application/json" -d '[
    "INSERT INTO _test ( name ) VALUES( '1' ) RETURNING id;"
]'
```

Returns:
```
{
    "results": [
        {
            "last_insert_id": 1
        }
    ]
}
```

Expected something like:
```
{
    "results": [
        {
            "last_insert_id": 1,
            "types": {
                "id": "integer"
            },
            "rows": [
                {
                    "id": 1
                }
            ]
        }
    ]
}
```

Also for example:
```
curl -XPOST 'localhost:4001/db/execute?pretty' -H "Content-Type: application/json" -d '[
    "INSERT INTO _test ( name ) VALUES( '2' ) RETURNING name;"
]'
```

Returns:
```
{
    "results": [
        {
            "last_insert_id": 2,
            "rows_affected": 1
        }
    ]
}
```

Expected something like:
```
{
    "results": [
        {
            "last_insert_id": 2,
            "rows_affected": 1,
            "types": {
                "name": "text"
            },
            "rows": [
                {
                    "name": "2"
                }
            ]
        }
    ]
}
```

And:
```
curl -XPOST 'localhost:4001/db/execute?pretty' -H "Content-Type: application/json" -d '[
[...]

PR #1312, opened 2023-06-20
add support for sqlite RETURNING clause

elv-gilles:
This PR adds support for [sqlite RETURNING clause](https://www.sqlite.org/lang_returning.html).

Requests involving the returning clause must include a boolean field before the SQL statement (see `ParseRequest` in `http/request_parser.go`).


otoolep:
Thanks for your contribution, however this is not the way to solve this. Instead the SQL parser that rqlite uses should be updated to detect when a SQL statement contains `RETURNING` and then rqlite should automatically do the right thing. That is my plan anyway.

The SQL parser is here: https://github.com/rqlite/sql

It would that code that needs support added for detecting `RETURNING`.

otoolep:
Unless I am missing something that will prevent `RETURNING` support being added to the SQL parser, I do not plan to merge this PR.

otoolep:
Thanks again for your contribution. If you'd instead like to add support to the SQL parser Go code, I would be interested in a such a PR. Basically some way that the parser could tell you `RETURNING` was in a given statement, allowing new logic to kick in in rqlite.

otoolep:
I would suggest you propose some of your ideas first though (usually as a GitHub issue), as I  wouldn't like folks to spend a long time coding something, only to find out later that it didn't make sense when created as a PR.

elv-gilles:
OK, no problem: I missed the SQL parser. 

I agree this is a better solution and am looking into the parser.

elv-gilles:
I implemented support in the sql parser (https://github.com/rqlite/sql/pull/9), which unfortunately seems to have an issue (https://github.com/rqlite/sql/issues/10).

For the time being, I'll use what I proposed in this PR (from fork)

elv-gilles:
Notes:
* relates to #1157
* I tried several golang parsers (including `rqlite/sql`) and found [github.com/auxten/postgresql-parser](https://github.com/auxten/postgresql-parser) supporting complex statements **and** the returning clause.

PR #1317, opened 2023-06-22
Ensure WAL replay works

otoolep:


db/db.go
```diff
@@ -174,6 +210,54 @@ func RemoveFiles(path string) error {
...
+	}
+
+	if deleteMode {
+		db, err := Open(path, false, false)
+		if err != nil {
+			return err
+		}
+		if db.Close(); err != nil {
+			return err
+		}
```
elv-gilles, line 256: seems the intent was:

```
		if err = db.Close(); err != nil {
			return err
		}
```

PR #1350, opened 2023-07-18
Update old comments with correct function name

bjornpagen:
Just updating old comments


otoolep:
Thanks for your contribution -- I'll have to look into that test failure, but it's unrelated to your change.

PR #1405, opened 2023-10-03
provide a default timeout on http client

jtarchie:
When using the `rqlite` CLI, I could not determine an issue of connection. It turns out that the arguments I had provided were not working. The issue was not present until I added a timeout to the HTTP client.

### DIFF
--- a/cmd/rqlite/main.go
+++ b/cmd/rqlite/main.go
@@ -18,6 +18,7 @@ import (
 
 	"github.com/Bowery/prompt"
 	"github.com/mkideal/cli"
+	clix "github.com/mkideal/cli/ext"
 	"github.com/rqlite/rqlite/cmd"
 	"github.com/rqlite/rqlite/cmd/rqlite/history"
 	httpcl "github.com/rqlite/rqlite/cmd/rqlite/http"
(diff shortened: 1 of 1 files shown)


otoolep:
Thanks for your contribution. Can you show me what you passed to `rqlite`? Normally it exits immediately if it can't connect.

I do think your change makes sense.

cmd/rqlite/main.go
```diff
@@ -394,10 +395,13 @@ func getHTTPClient(argv *argT) (*http.Client, error) {
...
-	client := http.Client{Transport: &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: argv.Insecure, RootCAs: rootCAs},
-		Proxy:           http.ProxyFromEnvironment,
-	}}
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: argv.Insecure, RootCAs: rootCAs},
+			Proxy:           http.ProxyFromEnvironment,
+		},
+		Timeout: 10 * time.Second,
```
otoolep, line 403: Let's make this a const, like what we have at line 26. E.g. 

`const httpTimeout = 10 * time.Second`

jtarchie: Can do.

jtarchie:
> Thanks for your contribution. Can you show me what you passed to `rqlite`? Normally it exits immediately if it can't connect.
> 
> I do think your change makes sense.

I assumed when I passed `-s https` that it would switch to the correct port 443. When I added the timeout, I finally saw the error message of the `scheme://hostname:port` and realized what was missing. I understand why the port wasn't changed now.

otoolep:
@jtarchie -- were you running an `rqlited` process at the time? If so, what host and port was it listening on? When I try to reproduce this issue I get:

```bash
rqlite -s https
ERR! Get "https://127.0.0.1:4001//status": http: server gave HTTP response to HTTPS client
```

and if nothing is running I get:
```bash
rqlite -s https 
ERR! Unable to connect to rqlited at https://127.0.0.1:4001 - is it running?
```

otoolep:
Ack, thanks, I'll merge this (or something very similar) soon.

jtarchie:
I can make changes if you want. Please let me know.

cmd/rqlite/main.go
```diff
@@ -394,10 +396,13 @@ func getHTTPClient(argv *argT) (*http.Client, error) {
...
-	client := http.Client{Transport: &http.Transport{
-		TLSClientConfig: &tls.Config{InsecureSkipVerify: argv.Insecure, RootCAs: rootCAs},
-		Proxy:           http.ProxyFromEnvironment,
-	}}
+	client := http.Client{
+		Transport: &http.Transport{
+			TLSClientConfig: &tls.Config{InsecureSkipVerify: argv.Insecure, RootCAs: rootCAs},
+			Proxy:           http.ProxyFromEnvironment,
+		},
+		Timeout: httpTimeout,
```
adelowo, line 404: Could the timeout be configurable like the `insecure` option too? e.g `30s` or something that can be parsed with `time.ParseDuration`

otoolep:
I thought that might be overkill, but the Go docs say this timeout affects all HTTP requests. So if retrieving backup took longer than the timeout (quite possible) it might break that.

So yeah, let's add a timeout option. It would be:

```bash
-t 10s
```

Default to 30s?

jtarchie:
I used the [built-in functionality](https://github.com/mkideal/cli#example-14-time-and-duration) of the CLI library. It can parse durations from the command line arguments.

otoolep:
Thanks!

PR #1417, opened 2023-11-04
release: add missing dash in linux quickstart instructions

arturhoo:
There is a missing dash that causes the `cd` command to fail, if one is following/automating the instructions blindly.


otoolep:
Thanks!

Issue #1500, opened 2023-12-19
Cluster bootstrap fails ("too many colons in address") when IPv6 addresses are returned by DNS discovery


jtackaberry:
**What version are you running?**

v8.12.1

**Are you using Docker or Kubernetes to run your system?**

Yes, Microk8s configured with IPv6 support (dual stack)

**Are you running a single node or a cluster?**

3 node cluster

**What did you do?**

In developing the Helm chart, I tested out enabling dual stack mode on the K8s service that rqlite uses for service discovery.  So rqlite will receive IPv6 addresses from DNS in discovering its peers.  With both `dns` and `dns-srv`, I observed the cluster failing to bootstrap with the following error:

> [cluster-bootstrap] 2023/12/19 03:16:36 failed to notify all targets: [2001:470:b0d6:fd:dfe:afeb:1d8a:32bb:4001 10.3.50.187:4001 2001:470:b0d6:fd:ce70:bd4e:b908:3b6d:4001 10.3.59.109:4001 2001:470:b0d6:fd:8a42:53cc:a14a:5a26:4001 10.3.90.38:4001] (failed to notify node at 2001:470:b0d6:fd:dfe:afeb:1d8a:32bb:4001: factory is not able to fill the pool: dial tcp: address 2001:470:b0d6:fd:dfe:afeb:1d8a:32bb:4001: too many colons in address, will retry)

I haven't looked at the code, but my complete guess is that rqlite is handling the DNS lookup, receiving IPv6 addresses, appending `:<port>` to each address, and passing that up to the raft library.  I suspect the raft library requires IPv6 addresses to be represented in the canonical form `[<addr>]:<port>`, for example `[2001:470:b0d6:fd:dfe:afeb:1d8a:32bb]:4001`, and it's rqlite's responsibility to format it appropriately?

A bit of a tangent here, but note port 4001 above.  As you can tell from the diagnostic info below, I started rqlited with `-disco-mode=dns-srv -disco-config={"name":"rqlite-headless","service":"http"}`.  Even once I reverted the `rqlite-headless` service to IPv4-only, I still had problems with `dns-srv`. I specified `http` as the service name because [the docs](https://rqlite.io/docs/clustering/automatic-clustering/#dns-srv) references the HTTP port.  But in practice this seems to require the raft port (4002), not http.  Changing disco-config to `{"name":"rqlite-headless","service":"raft"}` which gets back port 4002 fixed that additional issue.

[...]

PR #1505, opened 2023-12-19
Correct handling of IPv6 addresses in rqlite shell

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1500

PR #1509, opened 2023-12-21
Add command line flag -node-verify-server-name

otoolep:

PR #1513, opened 2023-12-21
Fix server name override with inter-node TLS

jtackaberry:
With #1509 I was still unable to get the cluster bootstrap properly.

> [rqlited] 2023/12/21 21:45:32 launch command: /bin/rqlited -node-id rqlite-0 -http-addr 0.0.0.0:4001 -http-adv-addr rqlite-0.rqlite-headless.data.svc.cluster.local:4001 -raft-addr 0.0.0.0:4002 -raft-adv-addr rqlite-0.rqlite-headless.data.svc.cluster.local:4002 -auth=/config/sensitive/users.json -join-as=_system_rqlite -node-cert=/config/sensitive/node.crt -node-key=/config/sensitive/node.key -node-verify-server-name=rqlite.data.svc.cluster.local -node-ca-cert=/config/sensitive/node-ca.crt -http-cert=/config/client-tls/tls.crt -http-key=/config/client-tls/tls.key -http-ca-cert=/config/sensitive/client-ca.crt -disco-mode=dns-srv -disco-config={"name":"rqlite-headless","service":"raft"} -bootstrap-expect=3 -join-interval=1s -join-attempts=120 -raft-shutdown-stepdown -fk=true /rqlite
> [...]
> 2023-12-21T21:45:35.989Z [ERROR] raft: failed to make requestVote RPC: target="{Voter rqlite-2 rqlite-2.rqlite-headless.data.svc.cluster.local:4002}" error="tls: failed to verify certificate: x509: certificate is valid for rqlite.data.svc.cluster.local, not rqlite-2.rqlite-headless.data.svc.cluster.local" term=2

It turns out if I pass cfg.NodeVerifyServerName through on the mux side as well, it works.  I tested it both with mTLS disabled and enabled.

The problem is I don't really understand *why* it works (which is just to say I don't understand why it was failing to begin with).   The logs above come from rqlite-0, and the raft logs there are reporting that rqlite-2 is presenting an invalid certificate.  Is the raft library somehow (whether directly or indirectly) using the tls.Config created for the listener for outgoing connections?

Possibly you have some insight there.  But what I can at least say is that it works. :)

The PR also contains a couple other minor fixes (as separate commits).

### DIFF
--- a/cluster/service_mux_test.go
+++ b/cluster/service_mux_test.go
@@ -106,7 +106,7 @@ func mustNewTLSMux() (net.Listener, *tcp.Mux) {
 	key := x509.KeyExampleDotComFile("")
 	defer os.Remove(key)
 
-	mux, err := tcp.NewTLSMux(ln, nil, cert, key, "", true, false)
+	mux, err := tcp.NewTLSMux(ln, nil, cert, key, "", "", true, false)
 	if err != nil {
 		panic(fmt.Sprintf("failed to create TLS mux: %s", err))
 	}
--- a/cmd/rqlited/main.go
+++ b/cmd/rqlited/main.go
@@ -405,7 +405,7 @@ func startNodeMux(cfg *Config, ln net.Listener) (*tcp.Mux, error) {
 		}
 		log.Println(b.String())
 		mux, err = tcp.NewTLSMux(ln, adv, cfg.NodeX509Cert, cfg.NodeX509Key, cfg.NodeX509CACert,
-			cfg.NoNodeVerify, cfg.NodeVerifyClient)
+			cfg.NodeVerifyServerName, cfg.NoNodeVerify, cfg.NodeVerifyClient)
 	} else {
 		mux, err = tcp.NewMux(ln, adv)
 	}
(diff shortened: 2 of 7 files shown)


otoolep:
All good, see my work in https://github.com/rqlite/rqlite/pull/1514 - I've added an end-to-end test now.

I'm not sure I understand either, because clients could be coming from lots of different hostnames. It doesn't make sense TBH. I think I need to study this a bit more -- but my tests will not pass unless ServerName is set on both client and server.

jtackaberry:
> All good, see my work in https://github.com/rqlite/rqlite/pull/1514 - I've added an end-to-end test now.

Nice, I also fixed unit tests presumably in parallel with what you were doing in #1514.  But I also added a specific ServerName test in config_test.go which you may want to copy as well.

I'll close this PR then, in favor of #1514.

otoolep:
OK, thanks.

Something is not right here however. I have a sense it's passing, but for the wrong reason. I'm going to dig into this a bit more before merging the other PR.

otoolep:
I'll take a look tomorrow -- sounds like you can build a version that unblocks your own work, but clearly the ServerName functionality it broken on master.

jtackaberry:
> I have a sense it's passing, but for the wrong reason. 

Yes, I have this same sense. The fact that it works but neither of us know why is a red flag.  To your point, it fixes master which is in a broken state, so it might be worth merging sooner than later for that reason alone, but I agree it seems like the story shouldn't end there.

> sounds like you can build a version that unblocks your own work, but clearly the ServerName functionality is broken on master.

Yep, I'm unblocked.   Continuing on with the chart work.

PR #1520, opened 2023-12-23
End-to-end testing of setting ServerName

otoolep:


otoolep:
Follow up on https://github.com/rqlite/rqlite/pull/1514

Issue #1560, opened 2024-01-04
Support other S3-like targets for auto backups


jtackaberry:
Currently rqlite is relatively opinionated about the target cloud storage for auto backups.  It precludes using S3-like storage backends that are S3 API compatible but don't exactly implement all of S3's behavior.

In my own personal case, this includes Minio and Wasabi.  In both cases, the endpoint doesn't include the bucket name, but rqlite currently (based on how it uses the AWS SDK) always prepends the bucket name to the endpoint, as is AWS's way.

So this FR is just to support S3-compatible APIs but be less opinionated about the endpoint.

I could submit a PR for this one, since I have the ability to test with at least Minio and Wasabi.

otoolep:
Go for it, happy to review a PR.

jtackaberry:
On the to-do list. I may also add some logging as well. Currently uploads are entirely silent unless they fail. :)

otoolep:
Fixed on master.

PR #1563, opened 2024-01-04
Allow configurable S3 path style for auto backups

jtackaberry:
AWS has deprecated path-style URLs but it's required for some S3-like alternatives such as Minio.

This is a PR for #1560, which allows the user to force the S3 path style to work with Minio.  Marked as draft as updates to tests are still needed.

I will pair it with a docs PR against [rqlite/rqlite.io](https://github.com/rqlite/rqlite.io) later tonight, to show examples of how to configure rqlite with Wasabi (which it turns out doesn't need forced path style) and Minio (which does).

### DIFF
--- a/auto/backup/uploader.go
+++ b/auto/backup/uploader.go
@@ -173,6 +173,7 @@ func (u *Uploader) upload(ctx context.Context) error {
 		stats.Get(lastUploadBytes).(*expvar.Int).Set(cr.Count())
 		u.lastUploadTime = time.Now()
 		u.lastUploadDuration = time.Since(startTime)
+		u.logger.Printf("completed auto upload to %s in %s", u.storageClient, u.lastUploadDuration)
 	}
 	return err
 }
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"io"
+	"strings"
 
 	"github.com/aws/aws-sdk-go/aws"
 	"github.com/aws/aws-sdk-go/aws/credentials"
@@ -53,7 +54,16 @@ func NewS3Client(endpoint, region, accessKey, secretKey, bucket, key string, for
 
 // String returns a string representation of the S3Client.
 func (s *S3Client) String() string {
-	return fmt.Sprintf("s3://%s/%s", s.bucket, s.key)
+	if s.endpoint == "" || strings.HasSuffix(s.endpoint, "amazonaws.com") {
+		// Native Amazon S3, use AWS's S3 URL format
+		return fmt.Sprintf("s3://%s/%s", s.bucket, s.key)
+	} else if !s.forcePathStyle {
+		// Endpoint specified but not using path style (e.g. Wasabi)
+		return fmt.Sprintf("s3://%s.%s/%s", s.bucket, s.endpoint, s.key)
+	} else {
+		// Endpoint specified and using path style (e.g. MinIO)
+		return fmt.Sprintf("s3://%s/%s/%s", s.endpoint, s.bucket, s.key)
+	}
 }
 
 // Upload uploads data to S3.
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -20,16 +20,18 @@ type S3Config struct {
 	SecretAccessKey string `json:"secret_access_key"`
 	Bucket          string `json:"bucket"`
 	Path            string `json:"path"`
+	ForcePathStyle  bool   `json:"force_path_style"`
 }
 
 // S3Client is a client for uploading data to S3.
 type S3Client struct {
-	endpoint  string
-	region    string
-	accessKey string
-	secretKey string
-	bucket    string
-	key       string
+	endpoint       string
+	region         string
+	accessKey      string
+	secretKey      string
+	bucket         string
+	key            string
+	forcePathStyle bool
 
 	// These fields are used for testing via dependency injection.
 	uploader   uploader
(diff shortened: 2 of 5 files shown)


jtackaberry:
I'll submit a separate PR to add logging when backups occur.

jtackaberry:
This PR is ready for review.

Doc updates over at https://github.com/rqlite/rqlite.io/pull/7

auto/backup/uploader.go
```diff
@@ -173,6 +173,7 @@ func (u *Uploader) upload(ctx context.Context) error {
 		stats.Get(lastUploadBytes).(*expvar.Int).Set(cr.Count())
 		u.lastUploadTime = time.Now()
 		u.lastUploadDuration = time.Since(startTime)
+		u.logger.Printf("completed auto upload to %s in %s", u.storageClient, u.lastUploadDuration)
```
otoolep, line 176: I'm concerned this log message is too noisy since it's periodic, and doesn't add much value. I try to avoid periodic repeating log messages if possible. The "last upload duration" is actually available from /status I believe (and if it's not, it should be). Or added to the /debug/vars output.

I'm open to convincing, but this data is available already I think.

aws/s3.go
```diff
@@ -20,37 +21,49 @@ type S3Config struct {
...
-	secretKey string
-	bucket    string
-	key       string
+	endpoint       string
+	region         string
+	accessKey      string
+	secretKey      string
+	bucket         string
+	key            string
+	forcePathStyle bool
```
otoolep, line 35: From reading this: https://docs.aws.amazon.com/sdk-for-go/api/aws/#Config

whether S3ForcePathStyle is set or not, uploading will work, correct? I'm reading the AWS SDK docs, but it's not clear what the implications are of setting this variable. This is distinct from this PR, does it matter if this variable is set or not?

auto/backup/uploader.go
jtackaberry: Ok, let me take a stab at convincing you:

Logs are used to signal important events taken by software. If I'm exporting, compressing, and shipping a copy of the database out to cloud storage, this for me qualifies as pretty important. While it's true it can be seen in the metrics (although getting those metrics into a system like Prometheus is currently challenging), logging this kind of event is very useful diagnostically.  For example if I see a spike in iowait% or cpu I might be wondering why, and the log line quickly, conveniently, and fully explains it.

As a operator, I would never complain about this kind of log event. I work with software that logs all kinds of useless crap not even the devs care about when I press them (sometimes even at WARN level!), but one log line every few hours or even minutes to signal an event that can have significant bearing on system resources is definitely not in that camp. :)

aws/s3.go
jtackaberry: Amazon S3 works fine when it's set to true, yeah, but Amazon has deprecated this mode and is phasing it out, so for Anazon-native S3 I would want to enable it even if it does currently work.

https://aws.amazon.com/blogs/aws/amazon-s3-path-deprecation-plan-the-rest-of-the-story/

auto/backup/uploader.go
jtackaberry: I hasten to add it's terribly uncharacteristic of me to argue in favor of *more* logging than less, but suffice it to say if you remain unconvinced I can drop it from the PR.

otoolep: All good, let's leave it so.

otoolep:
Looks good -- thanks.

otoolep:
I'm guessing we should do something similar for auto-restore from these systems?

jtackaberry:
> I'm guessing we should do something similar for auto-restore from these systems?

Unfortunately I forgot to test it (I will tomorrow) but that should be covered by the PR.  downloader.go was updated.

otoolep:
Ah yes, I see that now, missed it when reviewing the PR.

otoolep:
Is there any suggestions for testing? Could Minio be launched in CircleCI, the way I launch currently launch Consul and etcd for some end-to-end testing?

https://github.com/rqlite/rqlite/blob/master/.circleci/config.yml#L237

https://github.com/rqlite/rqlite/blob/master/system_test/e2e/auto_state.py

The idea would be we would launch Minio, and run an end-to-end upload and restore using a local Minio endpoint.

jtackaberry:
> Unfortunately I forgot to test it (I will tomorrow) but that should be covered by the PR.

Restores tested, worked fine.

> Could Minio be launched in CircleCI, the way I launch currently launch Consul and etcd for some end-to-end testing?

In principle, yes.  I'm not sure about the container execution environment with CircleCI and how you access services like etcd (or minio in this case) from rqlite, i.e. how you know the IP it's on.  I'm guessing all the containers share the same virtual interface so you just connect to localhost?

Using docker directly -- I'll let you translate to CircleCI-speak :) -- you can start a test instance of Minio like this:

```bash
$ docker run --name minio-test -p 9000:9000 -e MINIO_ROOT_USER=test -e MINIO_ROOT_PASSWORD=testpass quay.io/minio/minio server /data
```

In order to create a test bucket, you need to use Minio's `mc` tool.  This binary is available in the minio container image, so if you can exec into it after the container is running that'd be one option, or alternatively mc is available as a standalone container image at `quay.io/minio/mc`.  Assuming minio is available over localhost, you can create the bucket like so:

```bash
mc alias set test http://localhost:9000 test testpass
mc mb test/rqlite
```

This will create a bucket called `rqlite`.

rqlite can then point to it using the root user.  (Obviously you wouldn't do this in production -- you'd create a separate policy that grants access to the bucket, create a user, and attach the policy to that user -- but this is good enough for testing.)

```json
{
  "version": 1
  "type": "s3",
  "interval": "5s",
  "sub": {
    "access_key_id": "test",
    "secret_access_key": "testpass",
    "bucket": "rqlite",
    "endpoint": "http://localhost:9000",
    "path": "backups/citest.sqlite.gz",
    "region": "us-east-1",
    "force_path_style": true
  }
}
```

The endpoint needs to be prefixed with `http://` here.  Without an explicit scheme it assumes https.  But now that I run through this recipe, I see the stringified form of S3Client is looking a little janky: `s3://http://localhost:9000/rqlite/backups/citest.sqlite.gz`.   

I could submit another PR to strip the scheme from the endpoint before constructing the string (which would lose the fact of http or https in the string), or just drop the `s3://` prefix altogether for the non-Amazon-native case so it's prefixed with `http://` or `https://`.  Do you have a preference?  AFAICT the stringified form of S3Client is just used for display/logging purposes.  `s3://` isn't really a defined standard, so once you're outside of Amazon land all bets are off.

PR #1574, opened 2024-01-06
Switch to "BestSpeed" for all internode traffic

otoolep:


jtackaberry:
@otoolep  if you *really* want the best balance of speed and compression ratio, LZ4 is hard to beat.

PR #1578, opened 2024-01-06
Don't require static credentials for S3 access

jtackaberry:
Static AWS credentials make sense when you're running outside of AWS, but when you're running *inside* AWS, it's an operational smell.  Almost always the right thing to do is to use the role assigned to the environment where the workload is running, for example an EC2 role.  Or, for workloads deployed in AWS EKS, an [IRSA role](https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html).

The AWS SDK automatically does the right thing here when static credentials aren't specified.  If it's running in Kubernetes, it will use the K8s secret from the Service Account to authenticate itself to get an STS token.  Otherwise, it will try the EC2 IMDS (Instance Metadata Service) to fetch a token for the role assigned to the EC2 instance.

Or, even outside of AWS, it allows falling back to the `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` environment variables, or a shared credentials file (whose path can be overridden by `AWS_SHARED_CREDENTIALS_FILE`).

This is all very idiomatic AWS stuff.  All that's required is to forego specifying static credentials when creating the Session object.

This PR does just that when both the access key id and key secret are empty.

### DIFF
--- a/aws/s3.go
+++ b/aws/s3.go
@@ -120,12 +120,18 @@ func (s *S3Client) Download(ctx context.Context, writer io.WriterAt) error {
 }
 
 func (s *S3Client) createSession() (*session.Session, error) {
-	sess, err := session.NewSession(&aws.Config{
+	cfg := aws.Config{
 		Endpoint:         aws.String(s.endpoint),
 		Region:           aws.String(s.region),
-		Credentials:      credentials.NewStaticCredentials(s.accessKey, s.secretKey, ""),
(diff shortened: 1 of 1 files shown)


otoolep:
Thanks, looks good. Would you mind generating a PR for the rqlite.io docs/

jtackaberry:
> Thanks, looks good. Would you mind generating a PR for the rqlite.io docs?

Will do.

Issue #1581, opened 2024-01-09
Getting database disk image is malformed after rqlite interruption


dwco-z:
**What version are you running?**
v8.15.2

**Are you using Docker or Kubernetes to run your system?**
No

**Are you running a single node or a cluster?**
Cluster

**Issue Description**
I'm facing an issue where rqlite might fall into "failed to open store: failed to create snapshot store: check failed: journal mode to WAL: database disk image is malformed" after being previously interrupted (killed).

**Cluster information** (2-nodes cluster):
	- Node ID: 192.168.15.111, Suffrage=Voter
	- Node ID: 192.168.15.44, Suffrage=NonVoter
	- Node Interrumpted: 192.168.15.111, Suffrage=Voter
	

**Launch commands**:
	192.168.15.111(Voter):  	   rqlited.exe -raft-log-level=DEBUG -node-id 192.168.15.111 -http-addr 192.168.15.111:8906 -raft-addr 192.168.15.111:8904 -join 192.168.15.111:8904 -bootstrap-expect 1 -raft-non-voter=false -node-ca-cert DatabaseCertificateAuthority.cer -node-cert DatabasePublicCertificate.cer -node-verify-client -node-key DatabasePrivateCertificate.pem -http-ca-cert DatabaseCertificateAuthority.cer -http-cert DatabasePublicCertificate.cer -http-verify-client -http-key DatabasePrivateCertificate.pem Local
	192.168.15.44 (NonVoter): rqlited.exe -raft-log-level=DEBUG -node-id 192.168.15.44 -http-addr 192.168.15.44:8906 -raft-addr 192.168.15.44:8904 -join 192.168.15.111:8904 -raft-non-voter=true -node-ca-cert DatabaseCertificateAuthority.cer -node-cert DatabasePublicCertificate.cer -node-verify-client -node-key DatabasePrivateCertificate.pem -http-ca-cert DatabaseCertificateAuthority.cer -http-cert DatabasePublicCertificate.cer -http-verify-client -http-key DatabasePrivateCertificate.pem Local

After interrumption, if I start the cluster that's what happens in the voter machine:

```
             _ _ _
            | (_) |
   _ __ __ _| |_| |_ ___
  | '__/ _  | | | __/ _ \   The lightweight, distributed
  | | | (_| | | | ||  __/   relational database.
  |_|  \__, |_|_|\__\___|
          | |               www.rqlite.io
          |_|
 
[...]

PR #1582, opened 2024-01-09
Explicitly convert migrated snapshot DB to WAL mode

otoolep:


otoolep:
See if https://github.com/rqlite/rqlite/issues/1581 happens sooner in the upgrade process. This is a reasonable change regardless.

PR #1591, opened 2024-01-12
Add GitHub action on release to push multi arch image to dockerhub

NerdyShawn:


otoolep:
Cool.

NerdyShawn:
I'm still sorting through validating on my side to a sample dockerhub registry. Will ping again when make more progress on producing a usable image. I think I mainly need to get the tagging passed to the dockerfile and some of the action. I think this way can keep most of your circli bits in place but then we'll look for probably like a semVer tag released to trigger the image publication to dockerhub.

NerdyShawn:
Closing, fixed by 
- https://github.com/rqlite/rqlite/releases/tag/v8.24.9
Thanks Philip! :muscle:

Issue #1554, opened 2024-01-02
Support dynamic reloading of TLS files for raft and http protocols


jtackaberry:
Certificate rotations are a standard operational requirement.  Today, certificate rotations require a rolling restart of the cluster.

It's increasingly common for certificates to be issued through automated means with relatively short validity periods to mitigate certificate compromises (because revocation in PKI is notoriously broken).  For example, in Kubernetes, one can use [cert-manager](https://cert-manager.io/) to manage generation and rotation of certificates.  An operator may choose to use, for example, Let's Encrypt for the the client-facing HTTP port, which currently issues certificates with 90 day lifetimes (and [has said they may consider ratcheting down the lifetime in future](https://letsencrypt.org/2015/11/09/why-90-days)).  Inter-node certificates could be generated using, say, GCP CAS or AWS ACM PCA, or even cert-manager's native CA issuer.

In all these cases, the certificate rotation is fully automated, and updated PEM files are automatically swapped into the running rqlite pods (typically by means of a mounted Kubernetes secret).

This FR is a small quality-of-life enhancement to offload certificate management to outside orchestration such as cert-manager, for both the HTTP API and internode communication.

As with #1553, this could be done using [fsnotify](https://github.com/fsnotify/fsnotify) or polling mtime every few seconds, and should follow symlinks.

jtackaberry:
I was thinking about this one and I think a simpler implementation would be to replace population of the tls.Config.Certificates field with GetCertificate and GetClientCertificate callbacks, which point to a function that reads the cert/key from disk, caching it for a period of time (30 seconds or so).

This means the cert loading happens in-band of the connection, and this should work equally well for both egress and ingress TLS connections.

I'll do up a PR.

otoolep:
What is the standard approach, isn't responding to `SIGHUP` the way these systems traditionally work? I'd rather implement support for that, caching files for N seconds leads to racy, hard-to-test code in practice.

otoolep:
All manner of things are harder test when systems do async things like expiring caches. For example, what if one of the refreshes from disk fails? How do we inform the user? Continually log every 30 seconds? It gets messy.

jtackaberry:
It tends to be a mix between dynamic reloading and SIGHUP.  I see a couple options here:

1. I can use fsnotify or basic os.Stat polling to detect changes instead of caching as in my original suggestion
2. I can add a SIGHUP handler to reload certificates

[...]

PR #1595, opened 2024-01-13
Dynamically reload cert/keys if changed on disk

jtackaberry:
Sketching out the ideas for #1554.  Works for server certs, *should* work for client certs but I haven't tested it yet. CA certs aren't dynamically reloaded, but I don't think that's too important (and I don't think gotls makes that easy).

As mentioned in #1554, this implementation causes the node cert/key to be monitored multiple times (because they're passed multiple times to `CreateClientConfig()` and `CreateServerConfig()`), and so changes will result in logging several instances of "reloading certificate".  I plan to fix this by keeping a global cache of `monitoredCertificate` objects keyed by cert/key and returning the same object when the same cert/key is passed, but if you have a better idea let me know.

`monitorFiles()` is the stat poller.  Pretty simple.  I left it under rtls for now and we can decide where to put it later.

No tests yet, I can look at that once the design is finalized.


jtackaberry:
Also on reflection I'll probably rename "monitor" to "watch" (and variants thereof), just because it's a touch more idiomatic and "monitored" has other connotations.

otoolep:
You want feedback on this now? Or should I wait?

jtackaberry:
> You want feedback on this now? Or should I wait?

rtls/monitor.go
```diff
@@ -0,0 +1,83 @@
...
+			}
+			if !mtimes[n].IsZero() && info.ModTime() != mtimes[n] {
+				changed = append(changed, file)
+			}
+			mtimes[n] = info.ModTime()
+		}
+		if changed != nil {
+			cb(changed)
+		}
+		time.Sleep(2 * time.Second)
```
otoolep, line 31: This is not idiomatic Go. You should use the ticker pattern.

https://gobyexample.com/tickers

otoolep (commented):
This makes sense to me, but needs to be approached differently. For example services like this usually return a channel, allowing them to be shutdown cleanly. See https://github.com/rqlite/rqlite/blob/master/queue/queue.go#L198 for an example in rqlite. I would expect to see use of the "done" channel pattern.

Also, this could could be quite racy, since I don't see any read-write synchronization being done.

rtls/monitor.go
jtackaberry: Sure, I can adapt it easily enough. It's not clear to me what value using a ticker and channel will provide over callback in this case.  A ticker makes sense when you want to select on multiple things (most notably a context), but in this case it was literally just a "do a thing every interval T."

> Also, this could could be quite racy, since I don't see any read-write synchronization being done.

It uses an atomic.Value to hold a scalar (a pointer to tls.Certificate).

jtackaberry: >  I would expect to see use of the "done" channel pattern.

What is it that will signal doneness?

In this case there's no cleanup activity for the goroutine to do, and it doesn't impede rqlited's ability to terminate.

jtackaberry: To be clear, I'm not objecting to making the changes, just asking to be administered the cluebat a little bit more. :)

otoolep:
Yeah, following Go patterns is very important to me. It promotes quality, ease of testing, and makes it clearer for other programmers what your actual intention is. In this case using a Ticker makes it crystal clear what you're doing, in a way Sleep does not.

Not being shut down the monitoring service makes unit-testing more difficult. For example, unit testing takes place within the context of a single binary. Without clean shutdown, if I force a stack trace of the test binary (perhaps I'm debugging a test deadlock), I've extra Goroutines running that are nothing to do with the test failure.

otoolep:
>Not being able to shut down the monitoring service makes unit-testing more difficult. For example, unit testing takes place within the context of a single binary. 

Multiple unit tests run inside a single binary. Multiple different unit tests of the monitoring service would leave multiple goroutines running that make debug harder. Every time a given unit test finishes within the binary, I want all resources cleaned up so runtime is fully reset of the next unit test.

otoolep:
More generally, the monitoring code should be a struct, one the code can instantiate. You can see this pattern in the fsnotify system -- `NewWatcher`.

jtackaberry:
> Not being able to shut down the monitoring service makes unit-testing more difficult. For example, unit testing takes place within the context of a single binary. Without clean shutdown, if I force a stack trace of the test binary (perhaps I'm debugging a test deadlock), I've extra Goroutines running that are nothing to do with the test failure.

Right, good point. I'm with you from a testing perspective.

Outside of the testing context, I'm not sure how this should work.  The usual pattern is to propagate a `context.Context` throughout which is used to signal clean termination. rqlite isn't doing that, and with the goroutine being spawned from within `rtls.CreateServerConfig()` and `rtls.CreateClientConfig()`, that solution would end up being fairly invasive (needing an update to a large number of functions).

Alternatively, the file monitor goroutine could be spawned (and subsequently signaled to terminate) once from `rqlited/main.go` as a single goroutine (managed by a singleton struct) and rtls.Create{Client,Server}Config could register against that.

jtackaberry:
> More generally, the monitoring code should be a struct, one the code can instantiate. You can see this pattern in the fsnotify system -- NewWatcher.

Similar to what I concluded my last comment with.  The only nuance with rqlite is if the watcher's lifecycle is to be maintained from `rqlited/main.go` it would AFAICT need to be a global singleton of that struct, otherwise, like with Context, a large number of functions would need to be updated to propagate the Watcher object.

otoolep:
The use of Context within rqlite is weak, it should be better (rqlite development started before Context became a real thing, and that pattern has stuck unfortunately). If there is a better, idiomatic, solution using Context that gives the code complete control over the Monitoring object's lifecycle, that works better.

jtackaberry:
> If there is a better, idiomatic, solution using Context that gives the code complete control over the Monitoring object's lifecycle, that works better.

I think there is, similar to what you're doing with auto backups in `rqlited/main.go` -- creating a cancellable context and passing it to a function that spawns a goroutine -- the only thing is AFAICT it would require use of some form of global state.  This is common enough in Go (even in rqlite I see globals are used for at least expvar.Map) but some folks strongly object to globals. (Presumably because they *can* indicate a code smell, though IMO they don't *always*.)

In fact, I actually started down these lines but back-pedaled to a single function when I thought I could get away with it. :)  But it sounds like something more sophisticated is in order.

Two approaches come to mind:
 1. Make the file watcher itself a single global goroutine where rtls (and possibly elsewhere in future) registers a file watch (slice of filenames with a channel to be poked) via a `Watch()` function.  The registered files would be global state (map plus mutex) in the file watcher package (whatever that is).  `rqlited/main.go` then calls the `Start()` function in that package, which takes a Context and spawns a goroutine to poll the registered files, and which terminates when the context is cancelled.  `Watch()` then takes care of updating the global map which is checked by the single goroutine.
 2. Or, `NewFileWatcher()` returns an instance of a `FileWatcher` struct, which keeps said map/mutex within the struct, and where each instance spawns its own goroutine similar to what's in my branch now.  Except here the key difference is that the package also offers a `SetDefaultContext()` function which is called by `rqlited/main.go` to set a single context that's used to cancel all goroutines spawned by `NewFileWatcher`.  Then the `rtls` package creates one global instance of `FileWatcher` for its use -- global mainly to minimize the duplication of watching the same file (namely the node cert/key, since IIRC `CreateClientConfig()` is called for each internode connection).

So in the first case the global state is a map (plus corresponding mutex) that tracks the files registered for watch, while in the second case the state is a shared context set by main() and a singleton instance of `FileWatcher` used by rtls.

jtackaberry:
> global mainly to minimize the duplication of watching the same file (namely the node cert/key, since IIRC CreateClientConfig() is called for each internode connection).

You know, maybe this is an premature/unnecessary optimization.  Even if we do spawn a separate goroutine per node connection, so what?  We aren't talking about clusters with thousands or even hundreds of nodes.  Ok, we stat the same files multiple times per interval, and this will look a little wonky to anyone paying attention to strace, but it'll simplify things in rtls which uses the file monitor.

rtls/monitor.go
```diff
@@ -0,0 +1,83 @@
...
+		} else {
+			log.Printf("reloaded certificate %s", certFile)
+		}
+	})
+	return mc, nil
+}
+
+// load reads the certificate and key files from disk and caches the result for Get()
+func (mc *monitoredCertificate) load() error {
+	cert, err := tls.LoadX509KeyPair(mc.certFile, mc.keyFile)
```
LuKks, line 69: Question, does this safely fails if one file changed but the other still didn't?

For example, what happens if the cert file changes and there is some random slowness until the key file gets also written? Before loading them it could be wise to have a delay, unless `LoadX509KeyPair` fails if there is any kind of mismatch

jtackaberry: > unless LoadX509KeyPair fails if there is any kind of mismatch

Exactly. If the key doesn't match the cert, `LoadX509KeyPair()` returns an error and the active cert/key in memory won't be replaced.  When the other file gets updated (or whatever operational issue is resolved), then this will get re-executed and at that time.

LuKks: I did this similar feature in a lib

https://github.com/LuKks/simple-hosting/blob/main/index.js#L228-L255
Plus https://github.com/LuKks/simple-hosting/blob/main/lib/watch-file.js

otoolep:
@jtackaberry -- got a second request for this feature, so revisiting. Thanks for your initial work, looking at it now.

PR #1596, opened 2024-01-13
Track Raft logs which change the database

otoolep:


store/store.go
```diff
@@ -255,8 +251,13 @@ type Store struct {
 
 	// Latest log entry index actually reflected by the FSM. Due to Raft code
 	// this value is not updated after a Snapshot-restore.
-	fsmIndex   uint64
-	fsmIndexMu sync.RWMutex
+	fsmIdx   uint64
+	fsmIdxMu sync.RWMutex
```
jtackaberry, line 255: If you're just trying to ensure safety for concurrent get/set on a value (like fsmIdx), atomics are usually quite a bit faster.

Issue #820, opened 2021-06-23
rqlited Raft log control option doesn't control Snapshot logging


Dorsosiv:
**Is your feature request related to a problem? Please describe.**
I would like to decrease the amount of logs created by rqlite 

I would like the log-level to show logs that are more critical that the level I selected as a client.

for example, if I set the log-level to warning, then Ill see in the logs the warning and errors and fatals that the rqlite will have

**Describe the solution you'd like**
having an option for log level 

**Describe alternatives you've considered**
there is no alternative

otoolep:
In my experience the Raft code generates the most logs. Are you aware of the `-raft-log-level` option? 

E.g.

`./rqlited -raft-log-level=ERROR data`

Is that sufficient?

Dorsosiv:
@otoolep thanks for the answer, Ill check if it helped and notify you

Dorsosiv:
looks like it helped. Thanks

Dorsosiv:
Hi, I saw this weird behavior : 

notice that in the second line in the launch command the raft-log-level is ERROR and we still get INFO from the rqlited.



```
[rqlited] 2021/07/06 21:31:27 launch command: /rqlited -node-id rqlite-0 -raft-snap 12288 -raft-timeout 6s -raft-election-timeout 8s -join-attempts 5 -http-addr rqlite-0.rqlite:4001 -raft-addr rqlite-0.rqlite:4002 -raft-log-level ERROR -join http://rqlite-4.rqlite:4001 /node
[mux] 2021/07/06 21:31:27 received handler registration request for header 1
[mux] 2021/07/06 21:31:27 received handler registration request for header 2
[cluster] 2021/07/06 21:31:27 service listening on 10.254.11.183:4002
[mux] 2021/07/06 21:31:27 mux serving on 10.254.11.183:4002, advertising 10.254.11.183:4002
[rqlited] 2021/07/06 21:31:27 no preexisting node state detected in /node, node may be bootstrapping
[rqlited] 2021/07/06 21:31:27 join addresses specified, node is not bootstrapping
[store] 2021/07/06 21:31:27 opening store with node ID rqlite-0
[store] 2021/07/06 21:31:27 ensuring directory at /node exists
[store] 2021/07/06 21:31:27 0 preexisting snapshots present
[store] 2021/07/06 21:31:27 first log index: 0, last log index: 0, last command log index: 0:
[store] 2021/07/06 21:31:27 no cluster bootstrap requested
[...]

PR #1611, opened 2024-01-18
Only log full Raft snapshots

otoolep:


otoolep:
This effectively addresses #820

Issue #1609, opened 2024-01-17
Support automatic VACUUM in rqlite


otoolep:
It's not clear to me if VACUUM will break how Snapshotting works in rqlite 8.x. I need to look into a) if it does, and b) how we can support VACUUM in the 8.x series

The root cause for this is because rqlite Raft snapshotting assumes that each WAL file that is copied to the `rsnapshots` directory follows directly in sequence from the previous WAL file. Executing a VACUUM invalidates this assumption, and would mean WAL-based snapshotting would no longer work. This means rqlite would cease to function correctly.

Finally, executing a VACUUM comes with all the usual caveats as documented by the SQLite documentation, most notable that PRIMARY KEYs may be changed by the process. rqlite will do nothing to change this fact.

_Update: seems like it will break rqlite, since it moves pages around. That means full rqlite snapshot will be required post VACUUM._

otoolep:
The proposal is to support the periodic VACUUM of the SQLite database managed by rqlite. The feature would work be controlled follows:
- Two new command line flags would be added to `rqlited`, which would allow operators to control the automatic and periodic VACUUM of the SQLite database managed by rqlite:
  - First command line flag would be `-auto-vacuum-int` and would accept a Go-format time duration string, such as `7d`. This would set the interval between VACUUM attempts. If not set, then no automatic VACUUMs are executed on that node.
  - The second command line flag would be `-auto-vacuum-size` which would specify the minimum size (in bytes) that the SQLite database file must be before a VACUUM is executed. 0 is a valid value to set, and means a VACUUM will be executed every cycle.

The VACUUM flags should be set on every node on the cluster, including read-only nodes, for which VACUUM is required. It will not be **necessary** to have identical VACUUM settings on each node however.

Each node will persistently store the last time a VACUUM successfully executed, so even if the node is restarted, it will next attempt a VACUUM on schedule.

VACUUM is an IO-intensive operation, and will also require a fair amount of free disk space. Operators should ensure that there is at least 2 times the amount of free disk space as the size of the SQLite database about to be VACUUMed.

jmordica:
LGTM!

otoolep:
[...]

PR #1615, opened 2024-01-19
Add more involved WAL checkpoint unit test

otoolep:
Add extensive testing of the interaction between VACUUM, WAL checkpoint, and WAL copying.

https://github.com/rqlite/rqlite/issues/1609


otoolep:
Actually, test was inaccurate, and missing a WAL copy. This shows that a full snapshot is not required after a VACUUM as long as every WAL file in the sequence is captured.

PR #1616, opened 2024-01-19
Add time and checksum based change-detection functions to the database level

otoolep:


otoolep:
More testing for https://github.com/rqlite/rqlite/issues/1609

PR #1617, opened 2024-01-19
Add VacuumInto to database layer

otoolep:


otoolep:
More work for https://github.com/rqlite/rqlite/issues/1609

PR #1618, opened 2024-01-19
Add generic KV store GET and SET to Log

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1609

PR #1624, opened 2024-01-21
Fix cluster Client to avoid retrying commands when churning through connections pool

aromanovich:
During a load testing I've discovered that non-idempotent transactions sometimes are seemingly being retried.

I think the cluster client's `retry` method is one of the places where the retrying can happen. When a node redirects a query to a remote leader, it retries any error happened in attempt to `writeCommand`, including errors where we sent the entire command, but haven't got it ACKed.

This PR fixes `retry` to never retry cases when a command might have been sent.

### DIFF
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -222,7 +222,7 @@ func (c *Client) Backup(br *command.BackupRequest, nodeAddr string, creds *proto
 		Credentials: creds,
 	}
 
-	if err := writeCommand(conn, command, timeout); err != nil {
+	if err, _ := writeCommand(conn, command, timeout); err != nil {
 		handleConnError(conn)
 		return err
 	}
@@ -303,7 +303,7 @@ func (c *Client) RemoveNode(rn *command.RemoveNodeRequest, nodeAddr string, cred
 		},
 		Credentials: creds,
 	}
-	if err := writeCommand(conn, command, timeout); err != nil {
+	if err, _ := writeCommand(conn, command, timeout); err != nil {
 		handleConnError(conn)
 		return err
 	}
@@ -342,7 +342,7 @@ func (c *Client) Notify(nr *command.NotifyRequest, nodeAddr string, creds *proto
 		},
 		Credentials: creds,
 	}
-	if err := writeCommand(conn, command, timeout); err != nil {
+	if err, _ := writeCommand(conn, command, timeout); err != nil {
 		handleConnError(conn)
 		return err
 	}
@@ -383,7 +383,7 @@ func (c *Client) Join(jr *command.JoinRequest, nodeAddr string, creds *proto.Cre
 			Credentials: creds,
 		}
 
-		if err := writeCommand(conn, command, timeout); err != nil {
+		if err, _ := writeCommand(conn, command, timeout); err != nil {
 			handleConnError(conn)
 			return err
 		}
@@ -487,28 +487,33 @@ func (c *Client) retry(command *proto.Command, nodeAddr string, timeout time.Dur
 	var p []byte
 	var errOuter error
 	var nRetries int
+	var retriableOuter bool
 	for {
-		p, errOuter = func() ([]byte, error) {
+		p, errOuter, retriableOuter = func() ([]byte, error, bool) {
+			var retriableInner bool
 			conn, errInner := c.dial(nodeAddr, c.timeout)
 			if errInner != nil {
-				return nil, errInner
+				return nil, errInner, true
 			}
 			defer conn.Close()
 
-			if errInner = writeCommand(conn, command, timeout); errInner != nil {
+			if errInner, retriableInner = writeCommand(conn, command, timeout); errInner != nil {
 				handleConnError(conn)
-				return nil, errInner
+				return nil, errInner, retriableInner
 			}
 
 			b, errInner := readResponse(conn, timeout)
 			if errInner != nil {
 				handleConnError(conn)
-				return nil, errInner
+				return nil, errInner, false
 			}
-			return b, nil
+			return b, nil, false
 		}()
 		if errOuter == nil {
 			break
+		}
+		if !retriableOuter {
+			return nil, errOuter
 		}
 		nRetries++
 		stats.Add(numClientRetries, 1)
@@ -519,31 +524,34 @@ func (c *Client) retry(command *proto.Command, nodeAddr string, timeout time.Dur
 	return p, nil
 }
 
-func writeCommand(conn net.Conn, c *proto.Command, timeout time.Duration) error {
+func writeCommand(conn net.Conn, c *proto.Command, timeout time.Duration) (err error, retriable bool) {
 	p, err := pb.Marshal(c)
 	if err != nil {
-		return fmt.Errorf("command marshal: %w", err)
+		return fmt.Errorf("command marshal: %w", err), false
 	}
 
 	// Write length of Protobuf
 	if err := conn.SetDeadline(time.Now().Add(timeout)); err != nil {
-		return err
+		// This error is retriable, as we haven't written the command yet
+		return err, true
 	}
 	b := make([]byte, protoBufferLengthSize)
 	binary.LittleEndian.PutUint64(b[0:], uint64(len(p)))
 	_, err = conn.Write(b)
 	if err != nil {
-		return fmt.Errorf("write length: %w", err)
+		// This error is retriable, as we haven't written the command yet
+		return fmt.Errorf("write length: %w", err), true
 	}
-	// Write actual protobuf.
+	// Write actual protobuf
 	if err := conn.SetDeadline(time.Now().Add(timeout)); err != nil {
-		return err
+		return err, true
 	}
 	_, err = conn.Write(p)
 	if err != nil {
-		return fmt.Errorf("write protobuf bytes: %w", err)
+		// This error is not retriable, as we might have sent the entire command
+		return fmt.Errorf("write protobuf bytes: %w", err), false
 	}
-	return nil
+	return nil, false
 }
 
 func readResponse(conn net.Conn, timeout time.Duration) (buf []byte, retErr error) {


otoolep:
Thanks for this, I'm keen to make the system more correct and reliable.

Specifically which kinds of operations did you see being resent? Any in particular? What caused `writeCommand` to fail? Network issues, or just load on the remote node causing timeouts?

otoolep:
I do think you may be on to something here, the "just retry any failures" may not be the right thing to do.

aromanovich:
I have an rqlite cluster consisting of 3 nodes. Each node has enough compute and network resources. Ping time between them is under 1ms.

Here's my testing tool: https://gist.github.com/aromanovich/3eaf0ca8b53c9b9f6933df440422176e.
I've constructed it to test both throughput and consistency guarantees, thus the complicated logic.
It also uses `RequestParameterizedContext` that I've suggested in https://github.com/rqlite/gorqlite/pull/37.

I execute it three times using command like `N=100 RQLITE_URL='...' ./loadtest` (each time connecting to a different node) and leave it running for some time. Eventually all of them exit with an error like `xs[i]-xs[i-1] != 1: 36577 36577` (so this is not a read anomaly).

Further inspection shows the following:
```
localhost:4000> SELECT * from xs where id = "9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613"
+------------------------------------------------------------------------------+----------------------+-------+
| id                                                                           | r                    | x     |
+------------------------------------------------------------------------------+----------------------+-------+
| 9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613 | -7776685022141003624 | 36577 |
+------------------------------------------------------------------------------+----------------------+-------+
| 9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613 | -7776685022141003624 | 36577 |
+------------------------------------------------------------------------------+----------------------+-------+
| 9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613 | 3599219768392510623  | 36617 |
+------------------------------------------------------------------------------+----------------------+-------+

localhost:4000> SELECT * from cs where id = '9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613';
+------------------------------------------------------------------------------+----------------------+-------+
| id                                                                           | r                    | x     |
+------------------------------------------------------------------------------+----------------------+-------+
| 9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613 | -7776685022141003624 | 36576 |
+------------------------------------------------------------------------------+----------------------+-------+
| 9a08d9e5-164a-4a70-a268-44937d0d561a:node-2-fqdn:18613 | 3599219768392510623  | 36616 |
+------------------------------------------------------------------------------+----------------------+-------+
```

It worth noticing that:
1. all rows have the same `id` , meaning the test tool called a single `RequestParameterizedContext` exactly once;
2. the `RANDOM()` results stored in `r` differ, suggesting that retrying happened above rewriting logic.

In this particular case it seems that [handleRequest](https://github.com/rqlite/rqlite/blob/231a7ec171e832af4484084df6199dfbeafdc5b3/http/service.go#L1251) was called and succeeded twice. One of the obvious places where the retry could've happened is gorqlite's [rqliteApiCall](https://github.com/rqlite/gorqlite/blob/4e4ea5aa6d88d465c100f34afe2a53721ece46c1/api.go#L46).

I'm quite sure I've seen duplicates with the same `r` value (though much less often), suggesting retries happening somewhere down both the client and `handleRequest` method. I think this PR fixes one of such places.

otoolep:
How long does this take to happen? Have you baselined your test setup by seeing if this issue ever happens with a single node?

Before assuming it's a clustering issue (which it could well be, of course) have you proved to your satisfaction that this is not something external to the rqlite system (like your test program, or the rqlite client), by first load testing a single node?

cluster/client.go
```diff
@@ -519,31 +524,34 @@ func (c *Client) retry(command *proto.Command, nodeAddr string, timeout time.Dur
...
+		return err, true
 	}
 	_, err = conn.Write(p)
 	if err != nil {
-		return fmt.Errorf("write protobuf bytes: %w", err)
+		// This error is not retriable, as we might have sent the entire command
+		return fmt.Errorf("write protobuf bytes: %w", err), false
 	}
-	return nil
+	return nil, false
```
otoolep, line 554: The issue I have with this is that we don't know if it's retryable. Perhaps the network completely failed, not a single packet made it to the other side, and we could retry the write.

Before we make this change, I think it would be better to instrument the code better to count if writeCommand is ever failing (and in which place). If we detect no errors during your test run, and yet we still see results we can't account for, then this change won't help.

I do think this function is a bit naive however, for the reasons you are touching on here.

otoolep:
Actually, we already have some instrumentation that can help us understand this issue better. See this metric: https://github.com/rqlite/rqlite/blob/v8.16.8/cluster/client.go#L514C13-L514C29

Each node exposes this counter on /debug/vars. The metric name is `num_client_retries`. Is it non-zero on any node? If it remains at zero then there have been zero retries. Granted it's not specific to a type operation (that can be fixed) but if it's zero on all your nodes when you get your system into this state, then we can be sure it's not an issue with retries, at least not at the level your PR changes.

otoolep:
https://github.com/rqlite/rqlite/pull/1624 is a better place to start debugging this. I'll merge this instrumentation into master, so it's available in the next release. Perhaps you would like to build master yourself?

otoolep:
I just released 8.17.0, which has extra instrumentation built it. I'd be interested in what the metrics read when you hit this issue.

aromanovich:
> How long does this take to happen?
With `M=50 N=100` it takes around 10 minutes to fail (suspiciously close to 10 minutes, actually :)), and the first duplicate `x` value lies within [20000, 40000] range. Usually it is a short run of duplicates containing 10-20 values.

> if this issue ever happens with a single node?
Never happened on a single-node cluster.

>  The metric name is `num_client_retries`. Is it non-zero on any node?
Yes. Just ran a test on a freshly restarted cluster — 2 of 3 nodes have num_client_retries = 8. I can also see a couple of TCP retransmits on my container metrics.

> Perhaps you would like to build master yourself?
I'll do that and reproduce the issue with your extended stats later this week.

> If we detect no errors during your test run, and yet we still see results we can't account for, then this change won't help.
https://github.com/rqlite/gorqlite/blob/4e4ea5aa6d88d465c100f34afe2a53721ece46c1/api.go#L46 has to be another (and I guess much bigger) source of retries in the face of network issues.

otoolep:
>Yes. Just ran a test on a freshly restarted cluster — 2 of 3 nodes have num_client_retries = 8. I can also see a couple of TCP retransmits on my container metrics.

OK, well there you go. That's pretty clear evidence the system is retrying. I'd like to see the detailed metrics I just added, just so we can confirm it the Execute/Request code, and then I'll look into a change (perhaps something along the line you suggested).

That said, let's say we change the code so that the failure is sent all the way back to the client. That will then force the client to decide what to do. Resend, or give up, so the choice becomes the same. The only way to make this truly better is some sort of sequence numbers on each side of the connection. It does seem remarkable that the system is failing in just the right way that  a) the cluster client detects a failure, and b) the node on the other end still got a perfectly acceptable request. 

Another way to handle this is to change your SQL statements so that INSERTs are idempotent. This would require you to use stuff like https://sqlite.org/lang_conflict.html and specify primary keys.

aromanovich:
> That will then force the client to decide what to do. Resend, or give up, so the choice becomes the same.

At-least-once vs at-most-once is a choice, but I would argue that the client usually has more context for the right decision.

If the database chooses at-least-once, clients will see inconsistent results in some cases.
For example, if a client does `INSERT username INTO users` and the database first succeeds and then retries, the result will be surprising in either case:
* if `username` column has an unique constraint, the table content will be valid, but the client will get an error stating constraint violation;
* if `username` column is not unique, the table will end up with a duplicate row, but the client won't get any error.

So the client can't trust the `INSERT` outcome in any case. If it has to be sure, it will have to assert `SELECT COUNT(username) FROM users` equals to 1.

But if the database goes with at-most-once, clients will need to do that only in case of errors, which are supposed to be rare. In my experience this is the approach most SQL databases and their drivers take.

otoolep:
Yep, I agree. The client needs more control, so my initial plan would be to expose the errors to the client as you suggest.

That said, I'm keen to make sure we rule out *bugs*. To that end, I still think giving the client more visibility would help flush out any bugs. I am still surprised you're seeing issues, assuming you have good hardware. I remain concerned about a bug somewhere in rqlite. I do have some extensive cluster load tests as part of the automatic test suite, haven't seen anything like this before. If anything, I've only debugged missing writes, not extra writes.

I'll look into enhancing the API shortly, and pushing out a new release. It's days away however, perhaps you can keep gathering info (and, of course, running any modified version of rqlite you think is helpful).

otoolep:
Thanks again for raising this. I'm going to close it in preference to the simpler model where the node will not retry and return any error it gets. Let's see how that goes.

otoolep:
If we need more complex logic, we can try it.

aromanovich:
I ran tests on the new rqlited with extended stats. I've also instrumented gorqlite to print any retries within [rqliteApiCall](https://github.com/rqlite/gorqlite/blob/4e4ea5aa6d88d465c100f34afe2a53721ece46c1/api.go#L46).

rqlited's stats show some `num_client_retries` and no `num_client_request_retries`. All the duplicates happened due to retries on gorqlite's side. It's probably worth changing the default there as well.

aromanovich:
Thanks for taking on this issue!

otoolep:
>rqlited's stats show some num_client_retries

Probably because rqlite still retries requests between nodes that should be idempotent e.g. queries. I didn't see any harm leaving those retries in place. You might see `num_client_query_retries` or `num_client_query_retries` be non-zero.

otoolep:
>All the duplicates happened due to retries on gorqlite's side. It's probably worth changing the default there as well.

A PR perhaps changing the default would be more than welcome.

PR #1625, opened 2024-01-22
Count Cluster Client retries at operation level

otoolep:


otoolep:
Triggered by https://github.com/rqlite/rqlite/pull/1624

PR #1639, opened 2024-01-28
aws: reuse session and S3 manager

mauri870:
The AWS session and s3 manager are concurrent safe, they should be reused whenever possible:

    Sessions are safe to use concurrently as long as the Session is not
    being modified. Sessions should be cached when possible, because
    creating a new Session will load all configuration values from the
    environment, and config files each time the Session is created.

See https://pkg.go.dev/github.com/aws/aws-sdk-go/aws/session

Currently, an aws session and s3 client/manager are created every time a call to Upload, CurrentID or Download is made. I changed it so it creates one session and S3 manager during app startup and reuse it afterwards.


otoolep:
Thanks for your contribution, looks good.

PR #1653, opened 2024-01-31
fix busy_timeout stat for read-only db

mauri870:


mauri870:
@otoolep Would you like me to add a test that covers this? Perhaps inside https://github.com/rqlite/rqlite/blob/c5876e6/db/db_common_test.go#L15?

otoolep:
Well, thank you for this copy-n-paste bug.

Sure, more testing is always welcome -- go for it.

db/db_common_test.go
```diff
@@ -13,30 +12,20 @@ import (
 )
 
 func testBusyTimeout(t *testing.T, db *DB) {
-	rbt := random.Intn(10000)
-	_, err := db.ExecuteStringStmt(fmt.Sprintf("PRAGMA busy_timeout=%d", rbt))
+	wantTimeout := random.Intn(10000)
```
otoolep, line 15: The best way to test this is to generate two different random numbers, one for each timeout, and then read them back. Make sure they line up. Agreed?

otoolep:
Actually, I think if you modify the test I as suggested inline that's sufficient test coverage. No need to add anything in `db_common_test.go`.

db/db_common_test.go
mauri870: Yes, I just pushed that. Thanks!

otoolep:
Looks good, thanks. I'll merge on green.

I'm planning to release a version tonight, so this fix will in it.

mauri870:
Awesome, thanks @otoolep 🙏

otoolep:
How did you spot it?

mauri870:
I've been reviewing the changes you've made since my last commit. (https://github.com/rqlite/rqlite/compare/ec9934d167df831f86162a6a9ffd2a82c8d24dee...master)

PR #1659, opened 2024-01-31
proto: generate proto files with go generate

mauri870:
During codebase changes, I had to modify proto files. Noticed a lack of instructions for compiling them.

This commit adds support for running `go generate ./...` to generate the pb.go files.


otoolep:
I do have it documented, but here: https://rqlite.io/docs/install-rqlite/building-from-source/#protobuf-code-generation

How do I make this part of the build process generally?

mauri870:
> I do have it documented, but here: https://rqlite.io/docs/install-rqlite/building-from-source/#protobuf-code-generation
> 
> How do I make this part of the build process generally?

Thanks, I should've looked up the docs 😅 

Hmmm, I wonder if it makes sense to mention this whole $GOPATH section, I think it makes it too complicated. If someone is changing the protobuf files it is probably a developer that already has the source code somehow, probably a git clone. 

Maybe we can mention just:

```
// install protoc and protoc-gen-go...
// cd into rqlite...
go generate ./...
```

Wdyt?

mauri870:
> How do I make this part of the build process generally?

I usually include 'go generate' in a makefile target, I don't believe there is a way to run it automatically with 'go build.' So you can either incorporate it into a build script or makefile, or simply execute it manually when necessary. Running it manually whenever there are proto changes should be fine as well, it does not happen all the time.

otoolep:
Yes, I've always disliked the awkwardness of the current build instructions, this looks much better -- thanks.

Issue #1657, opened 2024-01-31
Support queries automatically timing out after X seconds


otoolep:
Offer this as a query parameter. I think the following C call could work:

https://sqlite.org/c3ref/interrupt.html

Start a timer in Go, and if it fires, call this C function on the connection.

otoolep:
Ha, the underlying SQLite Go library already supports setting a timeout, and uses this method.

https://github.com/mattn/go-sqlite3/blob/4702d9b5d640f42488752f5cf70a195b748ffe96/sqlite3.go#L2071

Could be worth trying this out.

mauri870:
I tried working on this and failed miserably. Maybe I'm trying to approach this the wrong way.

I implemented it in such a way that I can invoke [QueryContext](https://github.com/mauri870/rqlite/commit/b2d46800bbf041cf51fd03089acb8ad5ac36e248#diff-30c25f2a7d8ae335d2aada4b0a6013ad9d159595bb6a2a14b1d0022451b6753fR776) and [ExecContext](https://github.com/mauri870/rqlite/commit/b2d46800bbf041cf51fd03089acb8ad5ac36e248#diff-30c25f2a7d8ae335d2aada4b0a6013ad9d159595bb6a2a14b1d0022451b6753fR641) with a context.WithTimeout with the duration specified by the `sql_timeout` query parameter. Even tho the context reached the deadline those functions fail to return an error for it.

Any clues on what should be the proper way to implement this?

otoolep:
Can you show me your code?

mauri870:
Sorry, I forgot to add the link to my reply, here it is

https://github.com/rqlite/rqlite/compare/master...mauri870:rqlite:feature/sql-timeout

otoolep:
>Even tho the context reached the deadline those functions fail to return an error for it.

You mean the deadline doesn't expire? If I was you I would focus on writing a db-level unit test first, to prove to yourself it works/doesn't work and debug that. You are on the right track, as I too believe it'll be a new field in the `command.Request` proto.

But focus on getting it working at the DB level solely via unit tests. And perhaps just focus on `db.Query` to start. Show me that failing.

mauri870:
Thanks, I'll try that

mauri870:
[...]

PR #1666, opened 2024-02-02
add support for sql query timeout

mauri870:
For #1657


otoolep:
So, what was the secret?

DOC/DATA_API.md
```diff
@@ -199,6 +199,14 @@ curl -XPOST 'localhost:4001/db/execute?timeout=2m' -H "Content-Type: application
 ]'
 ```
 
+### SQL Query Timeout
```
otoolep, line 202: I don't maintain these docs any longer. Instead we will update the docs on https://github.com/rqlite/rqlite.io/blob/master/content/en/docs/API/api/_index.md

db/db_test.go
```diff
@@ -986,6 +988,117 @@ func Test_ParallelOperationsInMemory(t *testing.T) {
 	exWg.Wait()
 }
 
+func mustSetupDbForTimeoutTests(t *testing.T, n int) (*DB, string) {
```
otoolep, line 991: Nit: Go style says to use "DB", not "Db".

db/db_test.go
```diff
@@ -986,6 +988,117 @@ func Test_ParallelOperationsInMemory(t *testing.T) {
...
+func Test_ExecShouldTimeout(t *testing.T) {
+	db, path := mustSetupDbForTimeoutTests(t, 1000)
+	defer db.Close()
+	defer os.Remove(path)
+
+	q := `
+INSERT INTO test_table (key1, key_id, key2, key3, key4, key5, key6, data)
+SELECT t1.key1 || t2.key1, t1.key_id || t2.key_id, t1.key2 || t2.key2, t1.key3 || t2.key3, t1.key4 || t2.key4, t1.key5 || t2.key5, t1.key6 || t2.key6, t1.data || t2.data
+FROM test_table t1 LEFT OUTER JOIN test_table t2`
+	mustTimeoutExecute(db, q, 1*time.Millisecond)
```
otoolep, line 1047: Is this a test? Are you testing that it does timeout? Then you should check the error code returned by the underlying Execute or Timeout function.  Implicitly testing by not panic'ing isn't the right approach.

The "must" functions are for setup purposes, for doing stuff in tests that is not meant to be tested, and must execute for the actual testing to commence. `mustSetupDbForTimeoutTests` is a good example of such a function.

db/db_test.go
```diff
@@ -986,6 +988,117 @@ func Test_ParallelOperationsInMemory(t *testing.T) {
...
+
+func Test_QueryShouldTimeout(t *testing.T) {
+	db, path := mustSetupDbForTimeoutTests(t, 1000)
+	defer db.Close()
+	defer os.Remove(path)
+
+	q := `SELECT key1, key_id, key2, key3, key4, key5, key6, data
+	FROM test_table
+	ORDER BY key2 ASC`
+	mustTimeoutQuery(db, q, 1*time.Microsecond)
```
otoolep, line 1066: Ditto, this is not the right way to test this. Do the query, and test that the error that comes back looks good.

http/query_params.go
```diff
@@ -123,6 +123,16 @@ func (qp QueryParams) Key() string {
 	return qp["key"]
 }
 
+// SqlTimeout returns the value of the key named "sql_timeout_seconds".
+func (qp QueryParams) SqlTimeout(def time.Duration) time.Duration {
```
otoolep, line 127: Go Style says to call this "SQLTimeout"

otoolep: Also, add a check at line 29 for the validity of this variable, if exists, then you can parse directly here.

mauri870:
Basically what you said, the query was executing too fast. I hope the tests are not flaky, I ran with -count=1000 to make sure and it seems to have consistent results.

http/query_params.go
otoolep: I'm also torn about the use of "sql". I think "db_timeout" is better.

otoolep (commented):
Thank you, this looks generally like what I would do.

http/query_params.go
mauri870: Thank you; I was quite unsure about the naming here. I wanted something that conveys that this is a timeout per query, not per transaction or per database connection.

DOC/DATA_API.md
mauri870: Thanks, I moved the docs to https://github.com/rqlite/rqlite.io/pull/10

db/db_test.go
mauri870: Agreed, updated the tests :)

otoolep:
Thanks again, looks good, will make a couple of follow-up changes.

PR #1669, opened 2024-02-03
System-level testing of Query Timeouts

otoolep:


otoolep:
@mauri870 -- FYI, this system-level test is failing, not sure why at this time. Need to debug.

A system-level test is one where an almost-complete node is spun up and queries are executed. It's a type of integration test. While the DB-level unit tests for timeout do pass, this integration test is failing.

otoolep:
Ha, I believe it will work now. I had the timeout too long, and was checking for the error incorrectly.

Issue #1672, opened 2024-02-06
readyz/freshness do not consider raft index


aderouineau:
**What version are you running?**
8.19.0

**Are you using Docker or Kubernetes to run your system?**
Neither: running  rqlited directly.

**Are you running a single node or a cluster?**
Single voter + read-only

**What did you do?**
1.  Launched a voter node: `./rqlite-v8.19.0-linux-amd64/rqlited -node-id 1 data1`
2. Imported a large (~10GB) SQLite database with `.boot` in rqlite CLI
3. Launched a read-only node: `./rqlite-v8.19.0-linux-amd64/rqlited -raft-non-voter=true -http-addr=localhost:4011 -raft-addr=localhost:4012 -join=localhost:4002 data2`
4. Queried `curl -G 'localhost:4011/readyz?noleader'` which returned `[+]node ok`
5. Queried `curl -G 'localhost:4011/readyz'` which returned `[+]node ok [+]leader ok [+]store ok`
6. Queried `curl -G 'localhost:4011/db/query?level=none&freshness=1s' --data-urlencode 'q=SELECT * FROM table LIMIT 5'` which returned `{"results":[{"error":"no such table: table"}]}`

Logs of interest on the read-only node:
```
[rqlited] 2024/02/06 00:25:38 node HTTP API available at http://localhost:4011
[rqlited] 2024/02/06 00:25:38 connect using the command-line tool via 'rqlite -H localhost -p 4011'
[raft] 2024/02/06 00:25:39 [WARN]  failed to get previous log: previous-index=10 last-index=0 error="log not found"
[store] 2024/02/06 00:28:27 initiating node restore on node ID localhost:4012
[store] 2024/02/06 00:28:58 successfully opened database at /home/derouin/rqlite/data2/db.sqlite due to restore
[store] 2024/02/06 00:28:58 node restored in 31.508336751s
```

* I expected `/readyz` to be able to consider the staleness of its raft log, either by default or with a parameter. In this case, it should not say "node ok" until the data was available locally.
* I expected the SELECT query with `level=none&freshness=1s` to return an error due to staleness. It seems like `freshness` is based on the last time the node contacted the leader, but I think the test should also look at the raft index.

otoolep:
Thanks for the feedback.

>I expected /readyz to be able to consider the staleness of its raft log, either by default or with a parameter. In this case, it should not say "node ok" until the data was available locally.

[...]

PR #1673, opened 2024-02-06
Record latest Leader commit index

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1672

store/store.go
```diff
@@ -1806,7 +1807,8 @@ func (s *Store) isStaleRead(freshness int64) bool {
 	if freshness == 0 || s.raft.State() == raft.Leader {
 		return false
 	}
-	return time.Since(s.raft.LastContact()).Nanoseconds() > freshness
+	return time.Since(s.raft.LastContact()).Nanoseconds() > freshness &&
+		s.raft.CommitIndex() == s.DBAppliedIndex()
```
aderouineau: Having `s.raft.CommitIndex() == s.DBAppliedIndex()` makes the condition too strict, especially for large updates that can still finish within the allowed freshness.

Could we instead add a`Time` field in `Store` that gets updated to `LastContact` when there is no new index, and when there is, to whenever the index gets applied?

aderouineau: I noticed `s.DBAppliedIndex()` is protected by a mutex. Would it make sense to replace it with a https://pkg.go.dev/sync/atomic#Uint64?

store/store.go
```diff
@@ -1806,7 +1807,8 @@ func (s *Store) isStaleRead(freshness int64) bool {
 	if freshness == 0 || s.raft.State() == raft.Leader {
 		return false
 	}
-	return time.Since(s.raft.LastContact()).Nanoseconds() > freshness
+	return time.Since(s.raft.LastContact()).Nanoseconds() > freshness ||
+		s.raft.CommitIndex() != s.DBAppliedIndex()
```
aderouineau, line 1811: Having `s.raft.CommitIndex() != s.DBAppliedIndex()` makes the condition too strict, especially during a large update that can still finish within the allowed staleness.

Also, I noticed `s.DBAppliedIndex()` is protected by a mutex. What do you think about instead using a https://pkg.go.dev/sync/atomic#Uint64?

store/store.go
otoolep: Yeah, you're the second person to suggest this. @jtackaberry did too.

Makes sense, but I'll do that in a separate PR.

PR #1675, opened 2024-02-06
Use atomic.Uint64 instead of mutexes for Store indexes

otoolep:
The code is simpler and easier to read.

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -19,6 +19,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"sync/atomic"
 	"time"
 
 	"github.com/hashicorp/go-hclog"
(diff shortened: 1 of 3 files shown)


otoolep:
cc @jtackaberry @aderouineau

aderouineau (approved):

store/store.go
```diff
@@ -915,16 +914,11 @@ func (s *Store) WaitForFSMIndex(idx uint64, timeout time.Duration) (uint64, erro
...
 	for {
 		select {
 		case <-tck.C:
-			s.fsmIdxMu.RLock()
-			fsmIdx = s.fsmIdx
-			s.fsmIdxMu.RUnlock()
-			if fsmIdx >= idx {
-				return fsmIdx, nil
+			if s.fsmIdx.Load() >= idx {
+				return s.fsmIdx.Load(), nil
```
jtackaberry, line 921: The first and second Load() isn't synchronized. I assume in this context it's ok if the returned value is different than the one compared?  The original mutex-based code this same behavior too so I imagine it's fine, but I figure it can't hurt to double check.

otoolep: Yeah, it always goes up so it doesn't matter, but actually there is a neater way to fix. Thanks for flagging.

otoolep: https://github.com/rqlite/rqlite/commit/248c66f251cbea10094c2074c6623a7955deb63f

PR #1676, opened 2024-02-06
Proof of concept for OTel tracing

punya:
Screenshot 2024-02-06 2 57 31 PM

This only covers non-queued execute queries. If the direction makes sense, we could add support for other query types, improve code organization, add tests and documentation.


command/proto/command.proto
```diff
@@ -24,6 +24,7 @@ message Request {
 	bool transaction = 1;
 	repeated Statement statements = 2;
 	int64 dbTimeout = 3;
+	map<string, string> metadata = 4;
```
otoolep, line 27: So why do this? I am wondering if I had use context objects properly and throughout rqlite (I started coding rqlite long before contexts were introduced) would this be necessary?

cmd/rqlited/flags.go
```diff
@@ -206,6 +206,9 @@ type Config struct {
 
 	// MemProfile enables memory profiling.
 	MemProfile string
+
+	// OTLPDest sets the OTLP endpoint
+	OTLPDest string
```
otoolep, line 211: Just so we're clear, what did you set this to at launch time?

otoolep (commented):
The Trace you showed me looks about right. Clustered systems running on local host do about 10-100 INSERTs per second (because the disks saturate for all the nodes). rqlite won't win any prizes for speed.

cmd/rqlited/flags.go
punya:
```shellsession
$ cat << EOF > config.yml
receivers:
  otlp:
    protocols:
      grpc:
exporters:
  googlecloud:
    project: some-project-name
service:
  pipelines:
    traces:
      receivers: [otlp]
      exporters: [googlecloud]
EOF

$ otelcol-contrib \
  --config ./config.yml &

$ OTEL_SERVICE_NAME=rqlited.1 go run ./cmd/rqlited \
  --otlp-dest localhost:4317 \
  ~/node1 &

$ OTEL_SERVICE_NAME=rqlited.2 go run ./cmd/rqlited \
  --http-addr localhost:14001 \
  --raft-addr localhost:14002 \
  --otlp-dest localhost:4317 \
  --join localhost:4002 \
  ~/node2 &

$ OTEL_SERVICE_NAME=rqlite go run ./cmd/rqlite \
  --otlp-dest localhost:4317
```

punya: I should mention that this command-line argument is short and explicit, but doesn't expose the full feature set of the OTLP exporter. Before this PR is merged, I'd want to switch over to the [autoexport package](https://pkg.go.dev/go.opentelemetry.io/contrib/exporters/autoexport) which is configured using [these environment variables](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/).

PR #1681, opened 2024-02-10
Support checking Stale Reads relative to Leader clock

otoolep:


store/store.go
```diff
@@ -274,7 +274,9 @@ type Store struct {
 
 	// Latest log entry index actually reflected by the FSM. Due to Raft code
 	// this value is not updated after a Snapshot-restore.
-	fsmIdx *atomic.Uint64
+	fsmIdx         *atomic.Uint64
+	fsmUpdateTime  *AtomicTime
```
aderouineau, line 278: Add a comment to specify that this is the node-local time?

store/store.go
```diff
@@ -274,7 +274,9 @@ type Store struct {
...
+	appendedAtTime *AtomicTime
```
aderouineau, line 279: Add a comment to specify that this is the leader-local time (which may not be the current leader)?

command/proto/command.proto
```diff
@@ -36,6 +36,7 @@ message QueryRequest {
 	}
 	Level level = 3;
 	int64 freshness = 4;
+	int64 max_stale = 5;
```
aderouineau, line 39: What do you think about re-using `Level`, introducing a `QUERY_REQUEST_LEVEL_FRESH`? Or as an alternative, have a `strict_fresh` boolean.

aderouineau (changes requested):
Should `/readyz` be updated to allow freshness check? Strictly speaking I don't think it's necessary as an application could send a `SELECT 1` with the strict freshness requirement. But it may be a nice convenience?

command/proto/command.proto
otoolep: What is the goal? I don't see what you're getting at. What meaning are you trying to communicate with `stresh_fresh`?

store/store.go
otoolep: OK

store/store.go
otoolep: OK

otoolep:
>Should /readyz be updated to allow freshness check? Strictly speaking I don't think it's necessary as an application could send a SELECT 1 with the strict freshness requirement. But it may be a nice convenience?

Maybe, but one change at a time. :-) Let's see if we can agree on this proposed change, and gain confidence that it *is* correct.

command/proto/command.proto
otoolep: (Corrected review comment)

aderouineau: `QUERY_REQUEST_LEVEL_FRESH` / `strict_fresh` means that you not only want `time.Since(LeaderLastContact).Nanoseconds() > Freshness`, but also want `FSMIndex < CommitIndex && LastFSMUpdateTime.Sub(LastAppendedAtTime).Nanoseconds() > MaxStale`

otoolep: I see. Let me think about it, it might be simpler to understand.  I don't want to add a new consistency level however, because today consistency levels have nothing to do with time -- it's an orthogonal concept. 

But perhaps a simple `strict` flag, as an optional flag to include with `freshness` might be better. That way the user doesn't have to think about specifying two durations. Or I just add a new, standalone duration parameter, called `freshness_strict` (or something).

aderouineau: > **I don't want to add a new consistency level however, because today consistency levels have nothing to do with time -- it's an orthogonal concept.**

I don't think there's a definition that says they do not have to abide by time rules. Quoting https://en.wikipedia.org/wiki/Consistency_(database_systems):
> consistency (or correctness) refers to the requirement that any given [database transaction](https://en.wikipedia.org/wiki/Database_transaction) must change affected data only in allowed ways. Any data written to the database must be valid according to all defined rules, including [constraints](https://en.wikipedia.org/wiki/Integrity_constraints), [cascades](https://en.wikipedia.org/wiki/Cascading_rollback), [triggers](https://en.wikipedia.org/wiki/Database_trigger), and any combination thereof.

And I believe from the point of view of a user, `QUERY_REQUEST_LEVEL_FRESH` makes sense because it means they want a higher degree of trust (consistency). Granted, the one thing that makes it strange is that it's only applied if `freshness` is specified in the query.

However, I am not against having a boolean query parameter that strengthens the freshness check. I am just trying to justify why a new consistency level could make sense. I believe I could even argue that "freshness" could have been a parameterized level of consistency instead of a query parameter, as it itself is a level of consistency above "NONE".

> **perhaps a simple strict flag**

I figured `strict_fresh` because `strict` is pretty generic and there may be uses for that in the future.

store/store.go
```diff
@@ -1786,11 +1793,18 @@ func (s *Store) updateAppliedIndex() chan struct{} {
...
 }
 
-func (s *Store) isStaleRead(freshness int64) bool {
-	if freshness == 0 || s.raft.State() == raft.Leader {
+func (s *Store) isStaleRead(freshness, maxStale int64) bool {
+	if s.raft.State() == raft.Leader {
 		return false
 	}
-	return time.Since(s.raft.LastContact()).Nanoseconds() > freshness
+	return IsStaleRead(
```
aderouineau, line 1800: What is the advantage of splitting into two functions?

store/state.go
```diff
@@ -15,6 +16,31 @@ import (
 	"github.com/rqlite/rqlite/v8/snapshot"
 )
 
+// IsStaleRead returns whether a read is stale.
+func IsStaleRead(
+	LeaderLastContact time.Time,
```
aderouineau, line 21: It is strange to see PascalCase for local variables.

command/proto/command.proto
otoolep: >I don't think there's a definition that says they do not have to abide by time rules. 

As the designer of rqlite, I have a certain discretion to decide what the term _consistency_ involves for my database. :-)

>I figured strict_fresh because strict is pretty generic and there may be uses for that in the future.

I would go with `freshness_strict`, which makes it clear that's it's related, but we're starting to bikeshed now. I recognize that `freshness` is fundamentally not a great term, but it's more disruptive (and a breaking change) to try to change it.

otoolep: Updated comment.

store/store.go
otoolep: Testability.

store/state.go
otoolep: Yeah, Copilot did it. :-) Will fix now, you're right.

otoolep:
Thanks again @aderouineau -- I think we're getting close here. Letting it sit for a few days, while we think about the consequences of this change will definitely help. Ongoing feedback welcome.

otoolep:
Try a simple, optional, boolean flag `freshness_strict`. An example query would be:

```
curl 'localhost:4001/db/query?level=none&freshness=1s&freshness_strict' --data-urlencode 'q=SELECT * FROM foo'
```

otoolep:
Available in 8.20.0.

Thanks again, @aderouineau

PR #1682, opened 2024-02-12
Move Linux development and release to Go 1.22

otoolep:


otoolep:
Does like building RISC:

https://app.circleci.com/pipelines/github/rqlite/rqlite/4682/workflows/a91e71e3-73ae-4c4a-bfaf-920b330dfcdd/jobs/43161

mauri870:
Known riscv64 issue, see https://github.com/golang/go/issues/65646

otoolep:
Thanks for the cross-ref @mauri870

mauri870:
Quick update, this issue is now fixed and will be back ported to 1.22, in the next release. I already tested it myself and the error is now gone.

otoolep:
Still failing, even with Go 1.22.1.

https://app.circleci.com/pipelines/github/rqlite/rqlite/4841/workflows/49971640-6473-4367-a79a-4446eb79015f/jobs/45680

mauri870:
@otoolep I don't think this was included in 1.22.1, unfortunately. It was fixed too close to the patch release that it could not be included. We'll need to wait for 1.22.2.

It can be worked around by using internal linking for riscv64, but tbh that is not worth the effort.

Issue #1691, opened 2024-02-19
Do not convert BLOB value to base64


wyhaya:
**Is your feature request related to a problem? Please describe.**

When querying via the HTTP API, BLOB are always encoded as base64 and represented as a String in JSON, making it impossible for us to differentiate whether it is a BLOB or a TEXT.

```json5
"values": [
    [
        1,
        1.234,
        "Hello world",
        // BLOB or TEXT?
        "/+4i"       
    ]
]
```

**Describe the solution you'd like**

There should be a way to distinguish between BLOB and TEXT types.

**Describe alternatives you've considered**

1. We can use a binary protocol to replace JSON, which can avoid this situation.
2. Convert the BLOB to a JSON Array. For example: [0, 255, 255, 128].

otoolep:
I agree that this part of rqlite could use improvement. I'm trying to decide the best way to enhance the API to do so.

However can't you use the `types` column of the returned response to check it? Can I see the full request and response?

wyhaya:
> However can't you use the types column of the returned response to check it?

SQLite has affinity type, so even if the column type is 'BLOB', it can still store 'TEXT'.

##### HTTP Request Body
```json
[
  "SELECT * FROM all_types where id >= 12 LIMIT 2;"
]
```

##### HTTP Response Body
```json
{
  "results": [
    {
      "columns": [
        "id",
        "real",
        "text",
        "blob"
      ],
      "types": [
        "integer",
        "real",
        "text",
        "blob"
      ],
      "values": [
        [
          12,
          1.234,
          "Hello world",
          "/+4i"
        ],
        [
          13,
          1.234,
          "Hello world",
          "text"
        ]
      ]
    }
  ]
}
```

otoolep:
Can I ask what your use case is that, such that you could end up writing text to a blob field? If you control the application writing to rqlite, don't you control what goes into the columns? Also, are you aware of the `STRICT` keyword?

Since SQLite supports storing any value in any column, I want to make rqlite easy to use in this mode. But if I can learn more about your use case, it will help me provide a better solution.

otoolep:
This is a related issue: https://github.com/rqlite/rqlite/issues/1345

wyhaya:
I am writing a database manager which supports connection to rqlite.

When rqlite return data, I can't tell at all if it's BLOB or TEXT because it's always JSON(String).
[...]

PR #1704, opened 2024-02-26
Add rqlite shell support for BLOB data as arrays

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1691

Issue #1717, opened 2024-03-05
Recovery process can result in old snapshot getting sent to node


otoolep:
Started with https://github.com/rqlite/rqlite/issues/1711

This may be another race between nodes coming up during the Recovery process and pushing various snapshots to each other, with no guarantees when they arrive and are installed -- and the reap removing older snapshots.

otoolep:
@dwco-z - this may not be solvable through code, and may require some changes to the Recovery process steps.

dwco-z:
@otoolep do you mean I should adjust the way I'm doing the recovery? Maybe adding some random timeout during each node's startup to reduce the chances of the race condition?

otoolep:
@dwco-z -- maybe you can run the latest code, which will be available once this build completes: https://ci.appveyor.com/project/otoolep/rqlite/builds/49333481

I couldn't reproduce the issue locally, so your help would be good. Re-run your test, but with `-raft-log-level=INFO`, get the issue to occur, and then attach the logs to this bug.

dwco-z:
Sure, will do

dwco-z:
Here it is, I changed from-raft-log-level=DEBUG to -raft-log-level=INFO and captured the issue again:
[node2.log](https://github.com/rqlite/rqlite/files/14497271/node2.log)
[node3.log](https://github.com/rqlite/rqlite/files/14497272/node3.log)
[node1.log](https://github.com/rqlite/rqlite/files/14497270/node1.log)

otoolep:
OK, as I suspect, more snapshot reaping going on.

It's not immediately clear to me how to solve this issue. I'll need to think about it.

otoolep:
Couple of ideas, for my own reference:
- dig into why the Leader is sending an old snapshot to the follower
- consider reaping the snapshot in the future, if a node gets an older snapshot. Need to be cautious, as this could be a sign that something has gone wrong.

dwco-z:
> OK, as I suspect, more snapshot reaping going on.
> 
> It's not immediately clear to me how to solve this issue. I'll need to think about it.

Ok, anytime you need me to test changes just let me know. I believe you could not reproduce the bug because of some timing difference as you mentioned before.

dwco-z:
Just in case you need logs from the last created PR:
[node2.log](https://github.com/rqlite/rqlite/files/14509277/node2.log)
[node3.log](https://github.com/rqlite/rqlite/files/14509279/node3.log)
[...]

PR #1718, opened 2024-03-05
Support INFO-level logging of snapshot reaping

otoolep:


otoolep:
Help debug https://github.com/rqlite/rqlite/issues/1717

PR #1740, opened 2024-04-01
Insert BLOBs as byte slices

otoolep:


otoolep:
This is now supported.

Issue #1741, opened 2024-04-03
rqlited + DNS client 100% CPU usage after network disconnecting (windows)


mmolder:
**What version are you running?**
rqlite v8.23.0 on Windows 10/11

**Are you using Docker or Kubernetes to run your system?**
No

**Are you running a single node or a cluster?**
2+ node cluster

**What did you do?**
- Start rqlited on one server with the following command line:
    - `rqlited -node-id A -http-addr 0.0.0.0:4001 -http-adv-addr x.x.x.x:4001 -raft-addr 0.0.0.0:4002 -raft-adv-addr x.x.x.x:4002 -raft-cluster-remove-shutdown=true -raft-reap-node-timeout=15m C:\Path\To\db`
- Start rqlited on one or two more servers joining the first one to form a cluster
    - `rqlited -node-id B -http-addr 0.0.0.0:4001 -http-adv-addr y.y.y.y:4001 -raft-addr 0.0.0.0:4002 -raft-adv-addr y.y.y.y:4002 -raft-cluster-remove-shutdown=true -raft-reap-node-timeout=15m -join x.x.x.x:4002 C:\Path\To\db` 
- Pull the ethernet cable on one of the servers to lose connection
- Plug the cable back in and have it rejoin the cluster

After this, whenever rqlite is started on the server the CPU usage of rqlited and the windows DNS Client maxes out the CPU and slows it down to be unusable until rqlite is stopped. Flushing the DNS solves it temporarily until the above is repeated.

**What happened instead?**
![image](https://github.com/rqlite/rqlite/assets/11852114/000945c0-4cb3-45d6-87cc-cf28fcaa2d3c)

otoolep:
Thanks for the report.

I want to understand the sequence. You're saying that pulling the cable interferes somehow with the operation/behaviour of the server, such that the next time rqlited starts, it consumes 100% CPU? Is creating a cluster required to bring out this issue?

Basically this seems like something you could reproduce more simply. Is that the case?

mmolder:
The servers are connected together on a local network through a switch. I am testing some fault handling if a server were to be disconnected when running a cluster by joining and leaving it nicely and not so nicely. Pulling the ethernet cable simulates the server losing connection without having the node leave the cluster successfully. I have not seen the same behavior when running only one node.

[...]

PR #1742, opened 2024-04-03
Log HTTP writer failures

otoolep:


otoolep:
Address one specific issue seen in https://github.com/rqlite/rqlite/issues/1741#issuecomment-2034975798

PR #1749, opened 2024-04-18
Atomic seqNum

arturmelanchyk:


otoolep:
Thanks, definitely neater.

PR #1751, opened 2024-04-18
Perms hashset

arturmelanchyk:


otoolep:
I understand this a small optimization, is that right?

otoolep:
Or is it just a style thing?

PR #1782, opened 2024-05-17
chore: update installation with docker

jarqvi:


.github/workflows/release-notes.py
```diff
@@ -19,7 +19,7 @@
 Run a single node as follows:
 ```
 docker pull rqlite/rqlite
-docker run -p4001:4001 rqlite/rqlite
+docker run -d -p 4001:4001 -v data:/rqlite rqlite/rqlite
```
otoolep, line 22: Can you explain why we should enable this flags?

jarqvi: Thus, when I ran this container, I saw logs that apparently information such as snapshots are stored in the /rqlite path, so I preferred to store this data in a volume.
And I used the -d flag to run this container in the background.
240517_19h15m25s_screenshot

otoolep:
The goal of the Docker command is to show the simplest possible way to run rqlite.

.github/workflows/release-notes.py
jarqvi: OK, so at least it's good to add this to the documentation that important data is in the /rqlite path.

PR #1785, opened 2024-05-28
impl max() func

haifeiWu:
When I compile software from the source code, I found that the max function is not implemented.


otoolep:
Thanks, but that's because you are not using a newer version of Go.

https://pkg.go.dev/builtin#max

https://rqlite.io/docs/install-rqlite/building-from-source/#building-rqlite

otoolep:
Use Go 1.22.

Issue #1798, opened 2024-06-10
Very long starting time at every launch of rqlited


softwaremarkdahl:
Good morning.
I started a single node rqlited (planning to add nodes and make it a cluster of 3) and imported a big sqlite database (~450Gb) via .boot command in CLI.
Everything went as planned (it took some hours to import and validate the database), but when I stopped and restarted rqlited, it said there was a preexisting snapshot present and did a restore operation that lasted more than 1 hour.
Here's the log (some data is hidden by *****)

[rqlited] 2024/06/10 14:51:19 rqlited starting, version v8.25.0, SQLite 3.45.1, commit 312d6b01f44630fc4f425d353bdc8dde1f34d860, branch master, compiler (toolchain) gc, compiler (command) gcc
[rqlited] 2024/06/10 14:51:19 go1.22.3, target architecture is amd64, operating system target is linux
[rqlited] 2024/06/10 14:51:19 launch command: rqlited -node-id 2 -http-addr 0.0.0.0:4001 -http-adv-addr *****:4001 -raft-addr 0.0.0.0:4002 -raft-adv-addr *****:4002 -auth config.json -http-cert ./server-cert.crt -http-key ./server-key.pem ./rqlite/file/data
[rqlited] 2024/06/10 14:51:19 preexisting node state detected in /media/NVMEdisk2/rqlite/file/data
[mux] 2024/06/10 14:51:19 mux serving on [::]:4002, advertising *****:4002
[cluster] 2024/06/10 14:51:19 service listening on *****:4002
[http] 2024/06/10 14:51:19 secure HTTPS server enabled with cert ./server-cert.crt, key ./server-key.pem, mutual TLS disabled
[http] 2024/06/10 14:51:19 execute queue processing started with capacity 1024, batch size 128, timeout 50ms
[http] 2024/06/10 14:51:19 service listening on [::]:4001
[store] 2024/06/10 14:51:19 opening store with node ID 2, listening on *****:4002
[store] 2024/06/10 14:51:19 ensuring data directory exists at /media/NVMEdisk2/rqlite/file/data
[store] 2024/06/10 14:51:19 old snapshot directory /media/NVMEdisk2/rqlite/file/data/snapshots does not exist, nothing to upgrade
[snapshot-store] 2024/06/10 14:51:19 store initialized using /media/NVMEdisk2/rqlite/file/data/rsnapshots
[...]

PR #1799, opened 2024-06-10
Re-use existing SQLite database on startup

otoolep:
The checksums provide protection against a corrupted SQLite file. A SQLite file that passes the checksum test is the same logical database that would be restored if a copy was read and installed from the Snapshot store.


otoolep:
Possible fix for https://github.com/rqlite/rqlite/issues/1798

PR #1817, opened 2024-06-22
Add prebuild workflow

LuKks:
Leaving this here in case it's useful. I will be using something like this for https://github.com/like/rqlite-runtime

Tested:
- `linux-amd64` Ubuntu Noble (Focal and Bionic via Docker containers, static linking worked)
- `linux-arm64` Oracle server on Ampere
- `linux-arm` Online rented Raspberry Pi 4 (armv7l)

Untested:
- `darwin-amd64`
- `darwin-arm64`
- `win32-amd64`

Missing:
- Windows ARM64
- Raspberry Pi 5 (Maybe `linux-arm64` works there, untested)
- Android? (Same, maybe with `linux-arm64`)

For anyone reading, the platform/arch in the matrix strategy can be misleading because in reality has zero effect on the GitHub runner, it's just for naming and GOARCH. The os/runs-on is the important part, for example, macos-12 actually runs on x64 and macos-14 does truly runs on ARM64

### DIFF
--- a/.github/workflows/prebuild.yml
+++ b/.github/workflows/prebuild.yml
@@ -1,0 +1,59 @@
+name: Prebuild
+
+on:
+  workflow_dispatch:
+
+jobs:
+  prebuild:
+    strategy:
+      matrix:
+        include: [
+          { os: ubuntu-22.04, platform: linux,  arch: amd64,  compiler: gcc, ld: -static },
+          { os: ubuntu-22.04, platform: linux,  arch: arm64,  compiler: aarch64-linux-gnu-gcc, ld: -static },
+          { os: ubuntu-22.04, platform: linux,  arch: arm,    compiler: arm-linux-gnueabi-gcc, ld: -static },
+          { os: macos-12,     platform: darwin, arch: amd64 },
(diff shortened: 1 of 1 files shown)


LuKks:
The main idea is that you could manually or via tags trigger this prebuild action, download the binaries, and manually run a script that does your current `Upload Release Assets` and `Update the release notes`. This way releases would have binaries for all platforms

otoolep:
I just want to check something. I already publish binaries for most (if not all) of the architectures referenced in this PR anytime a release. Are you saying I'm missing something?

E.g. https://github.com/rqlite/rqlite/releases/tag/v8.26.2

LuKks:
Oh you added for Windows recently, and for Mac there is always `brew` so all cool then

LuKks:
I can see Windows was actually always available via some CI link

otoolep:
Thanks -- I may use part of this as I want to improve my GitHub workflow (mostly do some stuff in parallel). I'll leave this open as a reference.

youk:
For reference this is a discussion regarding Android builds – https://github.com/rqlite/rqlite/discussions/1987.

PR #1879, opened 2024-09-04
Support automatic PRAGMA optimize

otoolep:


store/store.go
```diff
@@ -510,6 +510,9 @@ func (s *Store) Open() (retErr error) {
 	if err != nil {
 		return fmt.Errorf("failed to create on-disk database: %s", err)
 	}
+	if err := s.db.Optimize(db.OptimizeAtOpen); err != nil {
```
otoolep, line 513: Actually no point doing this since the database is empty when created at the start!

Issue #1922, opened 2024-09-28
[Feature request] - Linearizable Follower read


lalalalatt:
I saw the store layer already have some implementations like `waitForCommitIndex`, `waitForFSM`, etc
But, currently, if the strong read request would be reject by follower and maybe(I guess) tell the client right leader.

Do you think it's necessary to add the follower read feature to both reduce leader load and guarantee strong read by utilizing `ReadIndex` of the Raft protocol?

**Additional context**
The follower read I refer here doesn't contain the "stale follower read", which might need to add another clock... and I think it would be a lot of work.😱

**Reference**
- [follower read - strongly consistent reads](https://docs.pingcap.com/tidb/stable/follower-read#strongly-consistent-reads)

otoolep:
Thanks for your request. Before I answer, have you studied the rqlite Read Consistency docs?

https://rqlite.io/docs/api/read-consistency/

Is there something missing there? What problem are you trying to solve?

peterxcli:
@otoolep Thanks for providing the doc. 
Yes I original though is "if the strong read request located on follower then it would be rejected and need to retry to send to leader"

Till I read the docs, under the [Strong](https://rqlite.io/docs/api/read-consistency/#strong) Section, it says: "The Follower waits for the response from the Leader, and then returns that response to the client."
Then I take a look to codebase, I found there is redirect logic exist actually😅
https://github.com/rqlite/rqlite/blob/a123aac03077555ba73e0ce2a2a12604a4fd96a2/http/service.go#L1266-L1298

That logic looks awesome, but I think we could make it better by adopting follower read optimization. Fetching the `commitIdx ` from leader and wait until follower's FSM also catch up leader's `commitIdx ` instead of transporting entire request and response payload through network.

I am not trying to solve any problem, I just think we maybe could make this better(if it is ok, I would like to make some contribution on this😄)

If there is anything wrong, please correct me~ thanks~

otoolep:
It sounds like a reasonable request, PRs are always welcome. You will need to handle the case (via a timeout) where the Follower never catches up. This could happen due to a network fault.

The only thing is to be clear what it gets us over using a Read Consistency of `None` along with suitable values for `freshness` and `freshness_strict`.  Maybe the part about the Follower waiting to cactch up instead of just returning an error.

peterxcli:
@otoolep Got it.

peterxcli:
@otoolep 
[...]

PR #1923, opened 2024-09-29
Node can retrieve commit index of other node

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/1922

PR #1929, opened 2024-10-02
New Strong Read Consistency option: Indexed

lalalalatt:
https://github.com/rqlite/rqlite/issues/1922

### DIFF
--- a/store/store.go
+++ b/store/store.go
@@ -881,6 +881,36 @@ func (s *Store) CommitIndex() (uint64, error) {
 	return s.raft.CommitIndex(), nil
 }
 
+// SafeLeaderCommitIndex returns the current Raft commit index if this node is the leader.
+// If verifyLeader is true, it first verifies that this node is still the leader before
+// returning the commit index. This function provides a safe way to get the leader's
+// commit index, ensuring that the node is actually the leader at the time of the call.
+//
+// rqlite's underlying Raft implementation uses leader leases, which, with proper
+// leader lease timeout configuration, ensures that two leaders cannot exist simultaneously.
+// However, clock skew may still occur. The verifyLeader option allows re-verification
+// of leadership status by querying a majority of Raft peers, providing an additional
+// safeguard against potential inconsistencies due to clock skew.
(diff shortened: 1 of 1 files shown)


cluster/client.go
```diff
@@ -145,6 +145,35 @@ func (c *Client) GetCommitIndex(nodeAddr string, retries int, timeout time.Durat
 	return a.CommitIndex, nil
 }
 
+func (c *Client) GetLeaderCommitIndex(nodeAddr string, retries int, trustLeaderLease bool, timeout time.Duration) (uint64, error) {
```
otoolep, line 148: Instead of adding a new function, just modify the existing `GetCommitIndex` so it takes a new boolean called `verifyLeader`. If set, do your leader check.

I am not concerned about this software talking to nodes that don't understand the new field in the proto because:

- I tell folks not to run clusters with a mix of versions, except during the small window they are upgrading
- I tell folks that if they do run clusters with a mix of versions they might get a query error until all nodes are upgraded.

This will be the case here, so I'm not worried about changing existing RPC calls.

otoolep: See https://github.com/rqlite/rqlite/blob/master/cluster/client.go#L129 to see what should change, and not add a new RPC call.

otoolep: As a result, you'll end up with something like this:
```
message NodeMeta {
    string url = 1;
    uint64 commit_index = 2;
    bool verify_leader = 3;
}
```

As I said, I'm not super-concerned about changes like this, as it's purely implementation details, and not issue in practise. Folks just need to upgrade all nodes, and not run mixed versions. I always make this clear in release notes.

peterxcli:
- [ ] add test for indexed follower read
    - [ ] multi-node tests where all followers are up (query should work)
    - [ ] multi-node tests where a quorum of followers are up (query should work). You do this by creating a cluster of 3 nodes, but kill one of the followers.
        - [ ] also need to test read consistency
    - [ ] multi-node tests where a quorum of followers are not up (query should fail). This would work by creating a cluster (of 3 nodes) and then killing 2 followers. Then you run the query against the remaining node.

PR #1930, opened 2024-10-02
Add Leader Index Read Optimization

lalalalatt:
## Background

@otoolep @peterxcli As we discuss in https://github.com/rqlite/rqlite/issues/1922 \[ https://github.com/rqlite/rqlite/issues/1922#issuecomment-2385585241, https://github.com/rqlite/rqlite/issues/1922#issuecomment-2385617205 \]
I add read index optimization for strong read consistency described in raft phd paper.

Also, I refer to [hashicorp/consul](https://github.com/hashicorp/consul)'s implementation of that optimization, and found that they do not use [lease read](https://www.pingcap.com/blog/lease-read/).
I guess they may also not trust the bounded clock drift🤔. Since we use same raft library, I think we should take the defensive way as they did.

## Description

Enhanced the Query method in store/store.go to support the new optimization:
- If leader read optimization is enabled, it calls a new leaderRead method instead of applying the query through Raft.
- The leaderRead method verifies leadership and then executes the query directly on the leader's database, avoiding additional consensus.

## Docs Modification
Under Read Consistency/Strong, add some description about the experimental query flag `leader_read_opt`: 

> [!Note]
> ### Strong
> 
> ...some existing docs...
> 
> #### `leader_read_opt` (Experimental flag)
> 
> If this flag is set to `true`, the system will attempt to optimize the query by using [read index optimization](https://www.pingcap.com/blog/lease-read/). This avoids the need to apply the query through the full Raft consensus process, thus reducing latency for reads that only require the leader's current state.
> 
> To enable leader read optimization, simply add this flag in your query parameters:
> 
> ```bash
> curl -G 'localhost:4001/db/query?level=strong&leader_read_opt' --data-urlencode 'q=SELECT * FROM foo'
> ```
> 

#### Call Hierarchy
- [raft.Backend.Read()](https://github.com/hashicorp/consul/blob/a6898939910b175db7495f02131918c0cc73027c/internal/storage/raft/backend.go#L105-L127)
	- [raft.Backend.leaderRead()](https://github.com/hashicorp/consul/blob/a6898939910b175db7495f02131918c0cc73027c/internal/storage/raft/backend.go#L129-L134)
		- [raft.Backend.ensureStrongConsistency()](https://github.com/hashicorp/consul/blob/a6898939910b175db7495f02131918c0cc73027c/internal/storage/raft/backend.go#L291-L296)
			- [consul.raftHandle.EnsureStrongConsistency()](https://github.com/hashicorp/consul/blob/a6898939910b175db7495f02131918c0cc73027c/agent/consul/raft_handle.go#L27-L29)
				- [consul.Server.consistentReadWithContext()](https://github.com/hashicorp/consul/blob/a6898939910b175db7495f02131918c0cc73027c/agent/consul/rpc.go#L1061-L1100)


otoolep:
Thanks for your contribution. How is this PR related to https://github.com/rqlite/rqlite/pull/1929 ?

http/service.go
```diff
@@ -1256,10 +1256,11 @@ func (s *Service) handleQuery(w http.ResponseWriter, r *http.Request, qp QueryPa
...
 		},
-		Timings:         qp.Timings(),
-		Level:           qp.Level(),
-		Freshness:       qp.Freshness().Nanoseconds(),
-		FreshnessStrict: qp.FreshnessStrict(),
+		Timings:                      qp.Timings(),
+		Level:                        qp.Level(),
+		Freshness:                    qp.Freshness().Nanoseconds(),
+		FreshnessStrict:              qp.FreshnessStrict(),
+		EnableLeaderReadOptimization: qp.EnableLeaderReadOptimization(),
```
otoolep, line 1263: Let's call this field `EnableLeaderRO`. Verbose variable names is not Go style.

store/store.go
```diff
@@ -1177,6 +1180,18 @@ func (s *Store) Query(qr *proto.QueryRequest) ([]*proto.QueryRows, error) {
 	return s.db.Query(qr.Request, qr.Timings)
 }
 
+func (s *Store) leaderRead(qr *proto.QueryRequest) ([]*proto.QueryRows, error) {
+	future := s.raft.VerifyLeader()
```
otoolep, line 1184: How does `VerifyLeader` actually work? I reviewed the Hashicorp docs, they simply state "ensure this peer is still the leader". How does that work? The Hashicorp code

From looking at the Hashicorp code the function claims to "ensure this peer is still the leader". What prevents the node losing leadership immediately after this call returns, but before it does the read?

From looking at the code it appears to tell the Followers to heartbeat immediately. Presumably that means that the Leader won't step down for at least an election timeout. Is that it?

http/query_params.go
```diff
@@ -133,6 +133,11 @@ func (qp QueryParams) Compress() bool {
 	return qp.HasKey("compress")
 }
 
+// EnableLeaderReadOptimization returns true if the query parameters request leader read optimizations.
+func (qp QueryParams) EnableLeaderReadOptimization() bool {
+	return qp.HasKey("enable_leader_read_optimization")
```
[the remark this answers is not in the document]
otoolep: `ro` might mean read-only to some people. But we need a *much* shorter query flag. `lro` might work.

http/service.go
otoolep: `EnableLRO`

otoolep: OK, well let's just do `LeaderReadOpt` -- that's what we should go with. There is no need for "Enable" since it's a boolean.

command/proto/command.proto
```diff
@@ -39,6 +39,7 @@ message QueryRequest {
 	Level level = 3;
 	int64 freshness = 4;
 	bool freshness_strict = 5;
+	bool enable_leader_read_optimization = 6;
```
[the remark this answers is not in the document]
otoolep: Oh, sorry, I see it is used, my mistake.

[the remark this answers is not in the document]
otoolep: The field name is too long however, let's use "leader_read_opt".

otoolep:
This seems reasonable, and is of interest to me. But it needs test coverage. The tests I would like to see are:
- single node tests where you actually test the query (query should work)
- multi-node tests where all followers are up (query should work)
- multi-node tests where a quorum of followers are up (query should work)
- multi-node tests where a quorum of followers are not up (query should fail). This would work by creating a cluster (of 3 nodes) and then killing 2. Then you run the query.

I think you'll be able to do these tests in [store_test.go ](https://github.com/rqlite/rqlite/blob/master/store/store_test.go)and [store_multi_test.go](https://github.com/rqlite/rqlite/blob/master/store/store_multi_test.go). You can follow the pattern of existing tests.

store/store.go
lalalalatt: yes

otoolep:
>Since we use same raft library, I think we should take the defensive way as they did.

Can you explain this statement? What do they do in Consul which is "defensive"?

store/store.go
```diff
@@ -1157,6 +1157,9 @@ func (s *Store) Query(qr *proto.QueryRequest) ([]*proto.QueryRows, error) {
 			return nil, err
 		}
 
+		if qr.LeaderReadOpt {
```
otoolep, line 1160: I would structure this code differently. In the event that the "leader read" is selected, you/ve done a compression in the lines about for no good reason.

I'll add more comments in the context.

otoolep: Change:
`if qr.Level == proto.QueryRequest_QUERY_REQUEST_LEVEL_STRONG`
to
`if qr.Level == proto.QueryRequest_QUERY_REQUEST_LEVEL_STRONG && !qr.LeaderReadOpt()`

then, add the very end of the function add this:

```Go
if qr.LeaderReadOpt() {
    if err := s.verifyLeader(); err != nil {
        return nil, err
    }
}
return s.db.Query(qr.Request, qr.Timings)
```
`s.verifyLeader()` will need to be written, it will just call the Raft code, not do any query.

peterxcli:
> Can you explain this statement? What do they do in Consul which is "defensive"?
They do not use lease read optimization.

store/store.go
```diff
@@ -1174,6 +1174,25 @@ func (s *Store) Query(qr *proto.QueryRequest) ([]*proto.QueryRows, error) {
 	if qr.Level == proto.QueryRequest_QUERY_REQUEST_LEVEL_NONE && s.isStaleRead(qr.Freshness, qr.FreshnessStrict) {
 		return nil, ErrStaleRead
 	}
+
+	if qr.LeaderReadOpt {
+		return s.leaderRead(qr)
```
otoolep, line 1179: This is not quite what we should do. It should read like this:
```Go
if qr.LeaderReadOpt() {
    if err := s.VerifyLeader(); err != nil {
        return nil, err
   }
}
return s.db.Query(qr.Request, qr.Timings)
```

We then delete `s.leaderRead()` and add `s.VerifyLeader()`
```Go
func (s *Store) VerifyLeader() err {
    return future := s.raft.VerifyLeader().Error()
}
```
This also has the advantage of exposing `VerifyLeader()`, making it easy to unit test.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	if err := s0.Join(joinRequest(s1.ID(), s1.Addr(), true)); err != nil {
+		t.Fatalf("failed to join to node at %s: %s", s0.Addr(), err.Error())
+	}
+
+	// Join the third node to the first as a non-voting node.
+	if err := s0.Join(joinRequest(s2.ID(), s2.Addr(), false)); err != nil {
+		t.Fatalf("failed to join to node at %s: %s", s0.Addr(), err.Error())
+	}
+
+	if _, err := s0.WaitForLeader(10 * time.Second); err != nil {
```
otoolep, line 1309: You don't need this call. `s0` will remain Leader throughout the Joins.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
```
otoolep, line 1398: Not needed.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	_, err := s0.Execute(er)
+	if err != nil {
+		t.Fatalf("failed to execute on leader: %s", err.Error())
+	}
+
+	// Kill one follower, still quorum
+	s2.Close(true)
+
+	// Wait for the leader to be elected again
+	if _, err := s0.WaitForLeader(10 * time.Second); err != nil {
```
otoolep, line 1416: Not needed. Since `s2` is not the leader, there will no election. If there was, something else has gone.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+		`CREATE TABLE foo (id INTEGER NOT NULL PRIMARY KEY, name TEXT)`,
+		`INSERT INTO foo(id, name) VALUES(1, "fiona")`,
+	}, false, false)
+	_, err := s0.Execute(er)
+	if err != nil {
+		t.Fatalf("failed to execute on leader: %s", err.Error())
+	}
+
+	// Kill two followers, no quorum
+	s1.Close(true)
```
otoolep, line 1510: Check the err on each Close() call. You will know that the nodes are done when it returns without error.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	s2.Close(true)
+
+	// Execute some data on the leader, wait for leadership to be lost
+	er = executeRequestFromStrings([]string{
```
otoolep, line 1514: Unnecessary. There is no need to wait for leadership to be lost. You killed the two nodes above. If the system is working as intended the query below should fail immediately.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	}, false, false)
+	_, err = s0.Execute(er)
+	if err == nil {
+		t.Fatalf("expected leader to not be able to commit, but it did")
+	} else if !strings.Contains(err.Error(), "leadership lost while committing log") {
+		t.Fatalf("unexpected error on leader: %s", err.Error())
+	}
+
+	// Perform a strong read consistency query with leader read optimization on the remaining nodes (should fail)
+	if s0.IsLeader() {
```
otoolep, line 1526: Remove this. Not because it isn't correct, but because the query below should fail immediately once you kill two node. Go straight to that, don't wait around, and it makes the test less clear.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+		t.Fatalf("expected leader to not be s0, but it was")
+	}
+	qr := queryRequestFromString("SELECT * FROM foo", false, false)
```
otoolep, line 1529: After killing the nodes, run this query immediately.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+		`CREATE TABLE foo (id INTEGER NOT NULL PRIMARY KEY, name TEXT)`,
+		`INSERT INTO foo(id, name) VALUES(1, "fiona")`,
+	}, false, false)
+	_, err := s0.Execute(er)
+	if err != nil {
+		t.Fatalf("failed to execute on leader: %s", err.Error())
+	}
+
+	// Kill two followers, no quorum
+	s1.Close(true)
```
otoolep, line 1595: Check the returned errors.

otoolep (commented):
Tests looking good -- thanks. Some more changes requested.

store/store.go
```diff
@@ -1174,6 +1174,25 @@ func (s *Store) Query(qr *proto.QueryRequest) ([]*proto.QueryRows, error) {
 	if qr.Level == proto.QueryRequest_QUERY_REQUEST_LEVEL_NONE && s.isStaleRead(qr.Freshness, qr.FreshnessStrict) {
 		return nil, ErrStaleRead
 	}
+
+	if qr.LeaderReadOpt {
```
otoolep, line 1178: Now that I think about it, this should be:
`if proto.QueryRequest_QUERY_REQUEST_LEVEL_STRONG && qr.LeaderReadOpt()`

In other words the flag should only have an effect if strong is selected.

otoolep:
This is close to being ready to merge, just a few more changes and we should be good. Of course, for it to be completely useful, you need to add support for automatically forwarding the request from a follower. But that can be a follow-up PR.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	if err := s0.Join(joinRequest(s1.ID(), s1.Addr(), true)); err != nil {
+		t.Fatalf("failed to join to node at %s: %s", s0.Addr(), err.Error())
+	}
+
+	// Join the third node to the first as a non-voting node.
+	if err := s0.Join(joinRequest(s2.ID(), s2.Addr(), false)); err != nil {
+		t.Fatalf("failed to join to node at %s: %s", s0.Addr(), err.Error())
+	}
+
+	if _, err := s0.WaitForLeader(10 * time.Second); err != nil {
```
otoolep, line 1495: Not necessary.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
```
otoolep, line 1580: Not necessary.

store/store_multi_test.go
```diff
@@ -1266,3 +1266,347 @@ func Test_MultiNodeStoreLogTruncation(t *testing.T) {
...
+	_, err := s0.Execute(er)
+	if err != nil {
+		t.Fatalf("failed to execute on leader: %s", err.Error())
+	}
+
+	// Kill two followers, no quorum
+	s1.Close(true)
+	s2.Close(true)
+
+	// Don't wait for the leadership to be lost, test if the leader read optimization can handle this case
```
otoolep, line 1598: Maybe I misunderstand the feature, but you should never have to wait for leadership to be lost. Since you killed the two other nodes, the Leader will be unable to heartbeat with a quorum of them (that is what the Hashicorp code does -- it reaches out to the followers, but in these tests those followers are gone).  Once you kill the two nodes if should not be possible to do a leader_read__opt read, right?

lalalalatt: yes, that wait is redundant

otoolep:
@lalalalatt @peterxcli -- I have one more question about this. I might be missing something.

What about this scenario.
- Let's say I have 3 nodes, each node is running on a VM. Let's call the nodes A, B, and C. A is the leader.
- The election timeout is 1 sec.
- I sent a read request to A, with `leader_read_opt` enabled.
- Node A does `VerifyLeader()` sending out heartbeats to each node.
- Node A receives the heartbeats from B and C in 50ms.
- Node A has just confirmed its the leader.
- Node A is just about to initiate the read of its local SQLite database.
- The hypervisor has a bug, and locks up the network interface and disk for node A for 5 seconds. The client is left waiting.
- An election happens since the other nodes are getting heartbeast, Node B is elected (by voting for itself and getting a vote from Node C).
- Another client comes in and does a write via Node B.
- The hypervisor bug resolves itself, Node A is running again, responds to the first client, unaware it's no longer leader.

The first client gets stale data. Is this a violation of strong consistency? The data the client gets back is stale.

peterxcli:
> @lalalalatt @peterxcli -- I have one more question about this. I might be missing something.
> 
> What about this scenario.
> 
> * Let's say I have 3 nodes, each node is running on a VM. Let's call the nodes A, B, and C. A is the leader.
> * The election timeout is 1 sec.
> * I sent a read request to A, with `leader_read_opt` enabled.
> * Node A does `VerifyLeader()` sending out heartbeats to each node.
> * Node A receives the heartbeats from B and C in 50ms.
> * Node A has just confirmed its the leader.
> * Node A is just about to initiate the read of its local SQLite database.
> * The hypervisor has a bug, and locks up the network interface and disk for node A for 5 seconds. The client is left waiting.
> * An election happens since the other nodes are getting heartbeat, Node B is elected (by voting for itself and getting a vote from Node C).
> * Another client comes in and does a write via Node B.
> * The hypervisor bug resolves itself, Node A is running again, responds to the first client, unaware it's no longer leader.
> 
> The first client gets stale data. Is this a violation of strong consistency? The data the client gets back is stale.

I think that's fine, 
Consider if there is no block, but the network between client and cluster has non-trivial latency, since all requests sent to B are **concurrent** with the blocked read request to A, so the read request may return any state among all **concurrent** DML and DDL
So, for client, it is not stale, and it may just think - my network is so bad...

otoolep:
>I think that's fine,

Thanks for your response. Just to be clear, I'm talking about the hypervisor freezing up after the hearbeats have been received from the followers, but **before the read is started**. 

I do not believe the sequence I outlined above is "linearizable" nor is it "strongly consistent". To quote https://jepsen.io/consistency/models/linearizable

_if operation A completes before operation B begins, then B should logically take effect after A._

Operation A -- the second client coming in and writing to the system -- does complete before operation B (B being the read that is eventually unblocked when the hypervisor bug self-resolves), but operation B does not reflect the write that takes place in operation A. 

Again, I could be wrong, but I do not see how this change provides strong consistency -- not without assuming something about when the Leader can deposed.

otoolep:
>Again, I could be wrong, but I do not see how this change provides strong consistency -- not without assuming something about when the Leader can deposed.

I just want to be clear on this point, so the docs we write are correct. I notice that YugabyteDB writes about this also:

https://www.yugabyte.com/blog/low-latency-reads-in-geo-distributed-sql-with-raft-leader-leases/

otoolep:
Seeing if the raft-dev mailing group can help:

https://groups.google.com/g/raft-dev/c/4QlyV0aptEQ/m/1JxcmSgRAwAJ

peterxcli:
> Seeing if the raft-dev mailing group can help:
> 
> https://groups.google.com/g/raft-dev/c/4QlyV0aptEQ/m/1JxcmSgRAwAJ

Thanks for posting that. I saw Jordan Halterman said:  **that means after reading the leader’s state machine but before sending the response to the client.** 

So, Should we reverse the order of the verifyleader and db.query(), cause we currently read first

peterxcli:
Also, I found, in [readIndex read implementation](https://www.pingcap.com/blog/lease-read/) of TiDB, they also wait for the fsm to apply to the current commit Index([highlight](https://arc.net/l/quote/qgycniet)).
But our approach and consul's are not taking that approach.

peterxcli:
> Operation A -- the second client coming in and writing to the system -- does complete before operation B (B being the read that is eventually unblocked when the hypervisor bug self-resolves), but operation B does not reflect the write that takes place in operation A.

Does operation B begin after operation A complete(client receive the response)?  If not, I think they are still concurrent, and  B might not logically take effect after A.

> Also, I found, in [readIndex read implementation](https://www.pingcap.com/blog/lease-read/) of TiDB, they also wait for the fsm to apply to the current commit Index([highlight](https://arc.net/l/quote/qgycniet)).

But I think that's ok too, same reason as above, if operation B begin after operation A complete, at the moment rqlite receive the operation B, the fsm(sqlite) must already apply to the commitIdx, and log store at commitIdx stored operation A

peterxcli:
> Also, I found, in [readIndex read implementation](https://www.pingcap.com/blog/lease-read/) of TiDB, they also wait for the fsm to apply to the current commit Index([highlight](https://arc.net/l/quote/qgycniet)). But our approach and consul's are not taking that approach.

https://groups.google.com/g/raft-dev/c/epxGXp461qc, seeing~

peterxcli:
@otoolep Is there any way to bootstrap multi node rqlite cluster in single machine(like some docker compose script?)? I am doing some benchmarking on this PR. Though it get a great improvement on read speed, but the test is on standalone node, it is necessary to simulate the nodes communicate overhead to get we closer to the real production environment performance gain.

BTW, below is the single node test result:
1. without leader read opt
```
rqbench -p '/db/query?level=strong' -b 2 -m 2000 -n 10000 "SELECT COUNT(*) FROM foo" -a localhost:4001
Test target: http://localhost:4001/db/query?level=strong
2000 requests completed in 10.513917ms
4000 requests completed in 7.899875ms
6000 requests completed in 7.988208ms
8000 requests completed in 10.916125ms
Total duration: 1m34.230106202s
Requests/sec: 106.12
Statements/sec: 212.25
```
2. with leader read opt
```
 rqbench -p '/db/query?level=strong&leader_read_opt' -b 2 -m 2000 -n 10000 "SELECT COUNT(*) FROM foo" -a localhost:4001
Test target: http://localhost:4001/db/query?level=strong&leader_read_opt
2000 requests completed in 136.666µs
4000 requests completed in 79.875µs
6000 requests completed in 79.708µs
8000 requests completed in 90.084µs
Total duration: 997.57477ms
Requests/sec: 10024.31
Statements/sec: 20048.62
```
Duration decreased by $\frac{94.23 \text{ seconds}}{0.99757 \text{ seconds}} \approx 94.5 \times$ faster.

otoolep:
I'm still interested in pursuing this option, but everyone seems a bit unclear on whether this (fundamentally) provides any stronger guarantees over _weak_. If we can figure it out, it would be helpful.

Also, since this type of read is mentioned in the Raft paper, that's another reason to add it. However the technical documents will need to be clear on what this type of read gets you. We certainly don't want to state this type of read is "linearizable" if it actually isn't.

otoolep:
I went back and re-read the Jepsen blog posts.

- https://aphyr.com/posts/313-strong-consistency-models
- https://aphyr.com/posts/316-jepsen-etcd-and-consul

The key statement is the following (from the second post): "Consistent reads, which require a round-trip delay so the leader can confirm it is still authoritative before responding."

One way to do this is to send the read through the Raft log, which is what _Strong_ means today. But it seems clear to me now that this is equivalent to doing this:
1. node records the current term
2. node does the read
3. node verifies that it is the leader via the heartbeat check
4. check that the current term is the same value after the heartbeat check as it was before the read started. If so respond with the data read in step 2.

If we do that, we can be sure that the read is not returning stale data and that the system remains _linearizable_. Any writes that complete after the read will be seen by every other client as also taking place after the read (assuming those clients also request strong/linearizable reads). So we're good.

@lalalalatt @peterxcli -- if we can ensure that this PR will enforce the conditions above I'm in favour of adding this new option. It does seem like a great way to get those strong reads, but in a faster manner. I think @freeekanayaka has it right -- it's basically a lightweight commit, a commit without having to write anything to the Raft log. I may even make it the way Strong works in 9.0.

Ref: https://groups.google.com/g/raft-dev/c/4QlyV0aptEQ/m/W8s5VzubBwAJ

otoolep:
Just to be clear though, there is one final twist here. Inside rqlite there is a delay between an already committed log actually being applied to the SQLite database. I believe this will need to be accounted for, but that can be a follow-up PR. The functionality outlined earlier still needs to be added.

So I do think the final sequence will be like this:

1. Node checks that it's Leader using a quick local state check. Same as weak.
2. Record commit index
3. Record term
4. Wait until the commit index has been applied. This will require code that should be out of scope of this PR IMHO.
5. Do the read
6. Verify leadership
7. Ensure the term hasn't changed from step 3.
8. Return the data read in step 5.

With all this in place, the data returned will be consistent. No other read will ever return data inconsistent with what the first read returns. It will be interesting to see how much faster this is than today's Strong reads. In the event that the Applied Index is caught up with the Commit Index it should be much faster.

otoolep:
@lalalalatt @peterxcli -- I just want to explain why this PR is not quite the same as the current _Strong_ implementation, though I do believe it's _linearizable_, even without the commit-is-applied check.

rqlite works as follows: a log entry is _committed_ to the Raft log in response to a write request, and then the Raft system later _applies_ the log entry to the state machine (in the case of rqlite the state machine is a SQLite database). While the time between the two steps is very short (milliseconds), it's not zero. In practical terms it usually doesn't usually matter as a write request to rqlite blocks until the log entry is **applied**.

_Strong_ sends the read through the Raft log. This means that any previous write is **both** committed and applied. However this PR just waits until all previous writes are committed the node has remained the Leader throughout the read, but not necessarily applied. So while this PR is quite close to _Strong_, it's not quite the same. This would prevent us from switching _Strong_ to what this PR does. In practise I would be surprised if anyone noticed this difference, but it's a theoretical possibility.

If we add the commit-is-applied check, then I believe it's effectively the same as _Strong_.  This should be faster in the case that the Leader's applied index is already caught up with the commit index, but if not, it may not be as fast as your initial testing showed.

So I believe that all this (very interesting!) work is giving us two things:

- linearizable reads without going through Raft. If we define the write is being complete when applied to the SQLite database this PR will be linearizable because the read in this mode will pick up any write that has been applied to the SQLite before the read starts (mostly thanks to SQLite's own transaction semantics -- the read will either see the entire write in effect, or none of it). 
- a sometimes faster way to get the equivalent of today's _Strong_ reads, if we add wait-until-the-commit-is-applied option.

peterxcli:
> So I do think the final sequence will be like this:
> 
> 1. Node checks that it's Leader using a quick local state check. This would be the same as weak if it did no other checks.
> 2. Record commit index
> 3. Record term
> 4. Wait until the commit index has been applied. This will require code that should be out of scope of this PR IMHO.
> 5. Do the read
> 6. Verify leadership
> 7. Ensure the term hasn't changed from step 3.
> 8. Return the data read in step 5.

Thx! This process looks great.
I will add these additional checks including wait-until-the-commit-is-applied option and two-phase-commit-index-check.
So I think we can merge this PR already, should we add these discussions into rqlite site after merging this PR? or wait until the follow-up PR is finished?

otoolep:
Merging now, I'm going to follow up with some changes because I do think there is a new consistency level here.

Thanks very much for your contribution.

PR #1937, opened 2024-10-09
Read Term from Raft level

otoolep:
The Term could be updated by log entries that are never passed to the FSM layer.


otoolep:
This value should really be exposed by the Hashicorp library, I'll look into submitting a PR.

otoolep:
Will use this if and when it merges: https://github.com/hashicorp/raft/pull/620

Issue #1971, opened 2024-11-03
`.status` reports `invalid character 'r' looking for beginning of value`


williamdes:
**What version are you running?**

v8.35.5

**What Operating System are you running?**

Linux

**If you are running Linux, which distribution are you using?**

Debian 12

**Are you using Docker or Kubernetes to run rqlite?**

No

**Are you running a single node or a cluster?**

3 node cluster, does the same error when single node

**What did you do?**

On the CLI I ran: `.status`.
`.nodes` and `.ready` work fine.

**What did you expect to happen?**

To not give a cryptic error

**What happened instead?**

`invalid character 'r' looking for beginning of value`

otoolep:
Can you show me your full session? I cannot reproduce any issues. When I download the 8.32.5, run it, and run the shell from the same release `.status` runs fine.

Also show me the output of the `/status` endpoint on the node you're hitting. You can do this using `curl`. 

https://rqlite.io/docs/guides/monitoring-rqlite/#status-api

williamdes:
Okay, curl helps here
The first line says:
```json
registered stats: route ip+net: netlinkrib: address family not supported by protocol
{"build":{"branch":"unknown","build_time":"unknown","commit":"unknown","compiler_command":"unknown","compiler_toolchain":"gc","version":"8"},"cluster":{"addr":".........
```

I guess this is due to the systemd restrictions I put on the service.
`RestrictAddressFamilies=AF_INET AF_INET6`

otoolep:
I'll make a new release shortly, perhaps you can run the test again and send me the log.

otoolep:
I'm going to consider this closed. The next step would be to understand better how to handle this in the network diagnostics code.

PR #1973, opened 2024-11-04
Include any failure message in `/status` output

otoolep:


otoolep:
Better handling of https://github.com/rqlite/rqlite/issues/1971

PR #1979, opened 2024-11-14
Preserve backward compatibility for backup endpoint

otoolep:


otoolep:
Do what was done in https://github.com/rqlite/helm-charts/pull/37 in rqlite code.

@jtackaberry -- WDYT?

aws/s3.go
```diff
@@ -72,6 +72,9 @@ func NewS3Client(endpoint, region, accessKey, secretKey, bucket, key string, opt
 	s3 := s3.NewFromConfig(cfg, func(o *s3.Options) {
 		if opts != nil {
 			if endpoint != "" {
+				if !isAWSEndpoint(endpoint) && !strings.Contains(endpoint, "://") {
```
jtackaberry, line 75: Why is it necessary here to avoid prepending https:// if the endpoint ends with `amazonaws.com`?

otoolep: Actually, you're right -- this is not needed. I forgot that the current end-to-end testing of actual S3 leaves endpoint blank. I thought it was setting the endpoint to an actual AWS endpoint (but without a protocol) so didn't want to disturb that. But I was wrong.

otoolep: Fixed.

PR #2000, opened 2024-12-25
Reduce memory allocations in queryStmtWithConn

arturmelanchyk:
Avoid unneeded allocations in for cycle

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -950,10 +950,11 @@ func (db *DB) queryStmtWithConn(ctx context.Context, stmt *command.Statement, xT
 	}
 	needsQueryTypes := containsEmptyType(xTypes)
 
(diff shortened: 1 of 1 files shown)


db/db.go
```diff
@@ -950,9 +950,10 @@ func (db *DB) queryStmtWithConn(ctx context.Context, stmt *command.Statement, xT
 	}
 	needsQueryTypes := containsEmptyType(xTypes)
 
+	mem := make([]interface{}, len(columns)*2) // single allocation for both dest and ptrs
```
otoolep, line 953: This makes sense to me, but this might be *too* clever. Instead just perform two allocations, and the code is clearer. The difference is time is small, especially if amortized over many loops.

arturmelanchyk: ok

otoolep:
Thanks for your contribution.

Issue #2011, opened 2025-01-08
Extensions lost on restart after node restore


a-h:
**What version are you running?**

v8.36.3

**What Operating System are you running?**

MacOS

**Are you using Docker or Kubernetes to run rqlite?**

No

**Are you running a single node or a cluster?**

Single node.

**What did you do?**

Created a database that uses the vector extension, which worked fine, then shut it down. When I started it back up, it says that the extension is loaded, but attempts to use the extension result in an error.

**What did you expect to happen?**

I expected the vector extension to be available for use in queries.

**What happened instead?**

I got the "no such module: vec0" error.

**Startup logs**

Running rqlite with the extension, I'd expect it to load.

```
rqlited -auth=auth.json -extensions-path=/nix/store/r0by5rl8xqkyzwz9qqsizjrw6hjq0bml-sqlite-vec-0.1.6/lib /Users/adrian/ragserver
```

And indeed, I can see `[db] 2025/01/08 14:32:59 loaded extensions: vec0.dylib` in the logs:

```
            _ _ _
           | (_) |
  _ __ __ _| |_| |_ ___
 | '__/ _  | | | __/ _ \   The lightweight, distributed
 | | | (_| | | | ||  __/   relational database.
 |_|  \__, |_|_|\__\___|
         | |               www.rqlite.io
         |_|

[rqlited] 2025/01/08 14:32:59 rqlited starting, version 8, SQLite 3.46.1, commit unknown, branch unknown, compiler (toolchain) gc, compiler (command) unknown
[rqlited] 2025/01/08 14:32:59 go1.23.3, target architecture is arm64, operating system target is darwin
[rqlited] 2025/01/08 14:32:59 launch command: rqlited -auth=auth.json -extensions-path=/nix/store/r0by5rl8xqkyzwz9qqsizjrw6hjq0bml-sqlite-vec-0.1.6/lib /Users/adrian/ragserver
[mux] 2025/01/08 14:32:59 mux serving on 127.0.0.1:4002, advertising localhost:4002
[rqlited] 2025/01/08 14:32:59 preexisting node state detected in /Users/adrian/ragserver
[cluster] 2025/01/08 14:32:59 service listening on localhost:4002
[http] 2025/01/08 14:32:59 execute queue processing started with capacity 1024, batch size 128, timeout 50ms
[http] 2025/01/08 14:32:59 service listening on 127.0.0.1:4001
[...]

PR #2014, opened 2025-01-09
Use existing driver when swapping databases

otoolep:


otoolep:
https://github.com/rqlite/rqlite/issues/2011

PR #2019, opened 2025-01-16
expose default prometheus metrics endpoint via golang library

NerdyShawn:
I cannot test but I believe this might try to add a metrics endpoint and expose default values. 
- https://prometheus.io/docs/guides/go-application/#how-go-exposition-works

We should be able to expand out further with custom metrics beyond defaults.
- https://prometheus.io/docs/guides/go-application/#adding-your-own-metrics


otoolep:
If I add more telemetry, it will be in OpenTelemetry form.

PR #2027, opened 2025-01-19
add bash completions

andreasstieger:
Add a simple bash completion script for the client

Are you inclined to include these?
There are generators our there (including https://github.com/posener/complete) but this seems straightforward

### DIFF
--- a/tools/rqbench.sh
+++ b/tools/rqbench.sh
@@ -1,0 +1,38 @@
+# rqbench completion                                       -*- shell-script -*-
+
+_rqbench_completions_filter() {
+  local words="$1"
+  local cur=${COMP_WORDS[COMP_CWORD]}
+  local result=()
+
+  if [[ "${cur:0:1}" == "-" ]]; then
+    echo "$words"
+
+  else
+    for word in $words; do
+      [[ "${word:0:1}" != "-" ]] && result+=("$word")
+    done
+
+    echo "${result[*]}"
+
+  fi
+}
+
+_rqbench_completions() {
+  local cur=${COMP_WORDS[COMP_CWORD]}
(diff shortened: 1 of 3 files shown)


otoolep:
Thanks for this. Is the idea folks would install it in their $PATH?

andreasstieger:
Not `$PATH` but the system-wide (`/etc/bash_completion.d`) or user-specific bash completions directories (`$BASH_COMPLETION_USER_DIR`, defaults to `$XDG_DATA_HOME/bash-completion` or `~/.local/share/bash-completion` if `$XDG_DATA_HOME` is not set)

otoolep:
Is there a specific way you suggest delivering this, or just add it to the repo so folks can install if the wish?

andreasstieger:
These are usually delivered along with the distribution package and dropped in the right place.

otoolep:
OK, that's what I thought.

otoolep:
I would be interested in distributing this, perhaps as part of Linux packages. Given best practices to follow I would create them as part of my existing GitHub actions (which are triggered on generation of a tag).
- https://github.com/rqlite/rqlite/actions
- https://github.com/rqlite/rqlite/tree/master/.github/workflows

andreasstieger:
I would recommend that unless you have a very specific reason for providing distribution packages, you leave that to the distribution packagers. Keep making good releases and they will be picked up - along with an upstream flow of improvements and fixes if needed.

[Packaging status](https://repology.org/project/rqlite/versions)

In any case, the three completions work for me at this time, ready for review.

otoolep:
Thanks for your contribution. Is there any other path in the repo you think others would naturally look for these? If not I might move them to a "completions" subdirectory.

andreasstieger:
`tools` seems reasonable, maybe `contrib`, `misc` or similar.

PR #2042, opened 2025-02-05
Log the addresses provided during Bootstrapping

otoolep:


otoolep:
Help with https://github.com/rqlite/rqlite/discussions/2040

PR #2106, opened 2025-06-12
Add initial CDC Service skeleton

otoolep:


cdc/service.go
```diff
@@ -0,0 +1,123 @@
...
+
+func (s *Service) readEvents() {
+	defer s.wg.Done()
+	<-s.done
+	return
+}
+
+func (s *Service) createStateTable() error {
+	er := executeRequestFromString(`
+CREATE TABLE IF NOT EXISTS_rqlite_cdc_state (
```
otoolep: This won't work because if the database is restored to a new cluster then the Raft indexes will be completely different, possibly lower. Latest Raft index transmitted to the webhook truly is cluster-specific.

So it needs to be stored separately from the SQLite database, which implies back to store in the BoltDB store. But what about when a node joins? It needs to get the high watermark. It would probably be OK, would pick it up eventually, but not sure.

Ideas:
- expand what's in a snapshot. But I hate changing snapshot format.
- have the new node reach out to other nodes to get it. Don't like that.

otoolep: Other ideas:
- reset the index in the CDC state table back to zero at restore time?
- convince myself via reasoning that it isn't actually an issue that any node will node be not leader before getting and applying a CDC index state message.

otoolep: There is a more fundamental issue here. A snap is sent over, along with logs. Those logs will have increasing entries of the high water mark. But those will be found out too early. 

Sending it along with the heartbeat is becoming more appealing.

Issue #829, opened 2021-07-21
Make it possible to interpolate the latest log index as part of a statement


marctrem:
## Use cases
It would be nice to be able to maintain sequential consistency when doing a sequence of queries outside of a read transaction.

## Tentative solution
Being able to add context from the log _inside_ statements solves this.

For example, if one could insert statements such as:
```sql
INSERT INTO facts (subject, relation, object, log_idx) VALUES (?, ?, ?, __RQLITE_NEXT_RAFT_INDEX__)
```

They could later ask rqlite (a leader or a replica) about the latest index they are aware of and issue requests such as:
```sql
SELECT * FROM facts WHERE log_idx <= ? AND <additional conditions>
parameters: [last_observed_raft_index]
```

This is great because it allows hedging of requests across replicas while still ensuring sequential consistency.

What do you think of that?

otoolep:
This is interesting, and reminds me of something else I had considered -- which might be easier to implement in the short term. What if the response to an INSERT came back with the Raft index in the response? That could also be useful.

I do like your idea, but it might need me to add a SQL parser to the rqlite code, perhaps, not sure. But it's interesting, no doubt about it.

That said, you could maybe achieve the same thing today -- at least very close -- by performing your insert, and after the node responded, querying the `status/` endpoint for the Raft state. That tells you the Raft index. You could then query the Raft state on a different node to be confirm whether that node had the same state.

Is this idea just for run, or do you have a particular application in mind?

SimonRichardson:
Alternatively, you could just fire a [barrier](https://github.com/hashicorp/raft/blob/main/api.go#L770) after the insert.

otoolep:
>Alternatively, you could just fire a barrier after the insert.

There is no need. The `INSERT` HTTP request doesn't respond to the client until the change has been applied to the actual database.

PR #2114, opened 2025-06-18
Support RaftIndex display and inter-node

otoolep:


otoolep:
Part of the fix for https://github.com/rqlite/rqlite/issues/829

PR #2119, opened 2025-06-18
Add RQLITE_HOST environment variable support

### LINKED ISSUE #2088: rqlite shell should support environment variables for rqlite cluster connection
Started with https://github.com/rqlite/rqlite/issues/2010

The rqlite shell program should check the following environment variables on startup:
- RQLITE_HOST

If set this envariable is non-empty it is split into a scheme, host, and port and then rqlite responds as though the command line flags -s, -H, and -p had been individually set. If parsing fails then it is as though the environment variable had been empty.

If both command line flags are set and env variables are set, command-line flags take precedence.

m04f:
Closes #2088.

Notes: 
  - RQLITE_HOST should be in the format [scheme:]//host[:port]
    This is the format expected by the [net/url](https://pkg.go.dev/net/url#URL) package


cmd/rqlite/main.go
```diff
@@ -273,6 +278,44 @@ func toggleFlag(op string, flag *bool) error {
...
+		return false
+	}
+
+	p, err := strconv.Atoi(uri.Port())
+	if err != nil {
+		// uri.Port() is empty
+		port = 4001
+	} else {
+		port = uint16(p)
+	}
```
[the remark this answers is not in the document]
m04f: The [Port](https://pkg.go.dev/net/url#URL.Port) method validates the value, if the value is invalid it returns an empty string

cmd/rqlite/main_test.go
```diff
@@ -0,0 +1,100 @@
+package main
+
+import (
+	"os"
+	"testing"
+)
+
+func TestParseHostEnv(t *testing.T) {
```
otoolep: This testing approach is very similar to something Copilot did for me a while back. See https://github.com/rqlite/rqlite/pull/2089

(I didn't merge it).

Did you use any LLM technology to generate this PR? Perfectly fine, I'm using LLMs too, so am curious.

m04f: I used LLMs to generate these tests, I didn't try to generate the entire PR with them

otoolep: These tests are good, but I think we should move these functions -- and tests -- into the http module inside the rqlite shell module. I don't have any unit tests alongside the commands.

I suggest we create new files called `env.go` and `env_test.go` inside https://github.com/rqlite/rqlite/tree/master/cmd/rqlite/http

That's where we can code up the environment variable parsing functions, and the unit tests. Make sense?

m04f: ok

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,42 @@
...
+import (
+	"net/url"
+	"os"
+	"strconv"
+)
+
+func ParseHostEnv() (protocol, host string, port uint16, succ bool) {
+	succ = false
+
+	hostEnv := os.Getenv("RQLITE_HOST")
```
otoolep: This should be a parameter to the function. What this module should be are general-use environment variable parsing code.

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,42 @@
...
+
+	if uri.Scheme == "" {
+		protocol = "http"
+	} else if uri.Scheme != "http" && uri.Scheme != "https" {
+		return
+	} else {
+		protocol = uri.Scheme
+	}
+
+	if uri.Hostname() != "" {
```
otoolep: Code would be simpler if it was just `if url.Hostname() == "" return` then you don't need the `else`, right?

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,42 @@
+package http
+
+import (
+	"net/url"
+	"os"
+	"strconv"
+)
+
+func ParseHostEnv() (protocol, host string, port uint16, succ bool) {
+	succ = false
```
otoolep: What is the point of this variable?

cmd/rqlite/http/env.go
otoolep: I think using LookupEnv is better, more explicit.

cmd/rqlite/http/env.go
m04f: It indicates whether the parsing was successful. This helps prevent the use of incomplete results for example, a protocol without a hostname. I realize this might seem redundant, but should I instead return default values (like empty strings) for all fields if parsing fails?

otoolep: Oh, sorry, I missed that you were returning it as a named returned. That's because it's not the style I use. Instead I tend have it in the return itself. E.g. `return "http", "host", 4001, false`. 

Let's take a step back. Why isn't returning an error going to work?

m04f: I didn't put much thought into it, the boolean just ended up being how I implemented it initially. You're right though, returning an error would be more idiomatic and clear. should I refactor it? is there anything else that can be improved?

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,38 @@
+package http
+
+import (
+	"fmt"
+	"net/url"
+	"os"
+	"strconv"
+)
+
+func ParseHostEnv() (protocol, host string, port uint16, err error) {
```
otoolep: This function should take the name of the Env variable. In other words it will be called as follows:
```Go
ParseHostEnv("RQLITE_HOST")
```

The environment names will be defined as constants in main.go

otoolep: Add a Go style comment explaining the function's purpose.

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,38 @@
...
+	if uri.Hostname() == "" {
+		return "", "", 0, fmt.Errorf("RQLITE_HOST does not contain a hostname")
+	}
+	host = uri.Hostname()
+
+	p, err := strconv.Atoi(uri.Port())
+	if err != nil {
+		// uri.Port() is empty
+		return protocol, host, 0, nil
+	}
```
otoolep, line 48: Check if port is greater than the max port number (65535?), return an error if so.

cmd/rqlite/main.go
```diff
@@ -93,6 +93,21 @@ func main() {
 			return nil
 		}
 
+		if !(ctx.IsSet("--host", "-H") || ctx.IsSet("--port", "-p") || ctx.IsSet("--scheme", "-s")) {
```
otoolep, line 98: Just so we're clear, describe in English your policy for deciding if the Env variables is to be parsed. We will then need to add that to the docs.

m04f: The environment variables for host, port, and scheme are only used if none of the corresponding command-line flags are provided.

This ensures that configuration comes from either all environment variables or all command-line flags, avoiding any mix between the two sources.

otoolep: So command line flags take precedence. Is it a mix? If I set the port via the command line flag, but also set the Env variable to both host and port? What happens then? Do I end up with a mix? I can read the code but I want to know what you want the policy to be. IMHO a mix is confusing. The policy I would recommend is the following:
- if any connection-related flag is set then use only connected-related flags
- if no connection related flags are set, then check for the environment variable. If that is not available then use the defaults.
- if the env variable is set, but there is an error parsing it, the program should exit.
-
WDYT?

cmd/rqlite/http/env.go
```diff
@@ -0,0 +1,49 @@
...
+
+	p, err := strconv.Atoi(uri.Port())
+	if err != nil {
+		if uri.Port() == "" {
+			return protocol, host, 0, nil
+		}
+		return "", "", 0, err
+	}
+
+	if p > 65535 {
```
otoolep: No magic numbers. Define this as a constant e.g. `maxPortNumber`.

otoolep:
Let me know when this is ready for final review. Also, be sure to actually test that it works -- set the env in your environment and confirm the shell actually uses it.

cmd/rqlite/main_test.go
```diff
@@ -0,0 +1,100 @@
...
+			expectResult: false,
+			expectArgs:   argT{},
+		},
+		{
+			name:         "invalid URL",
+			envValue:     "http://[invalid:url",
+			expectResult: false,
+			expectArgs:   argT{},
+		},
+	}
```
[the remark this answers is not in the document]
otoolep: These are good test cases to add. If you look at the port logic, you're still missing a check. The port number that is parsed must be >0 and < 65536. You're only check for one of those conditions last time I check.

m04f:
I have tested it and it works, I think it is ready for the final review

cmd/rqlite/main.go
m04f: This is the policy I used:
- If any of the connection flags is set the env variable is ignored
- If non of the connection flags is set check the env variable
- if an error occured during parsing or if the variable is not set, use the default connection parameters

m04f: I changed the policy so that if an error occured while parsing the env variable the program exits immediatly

otoolep: Works for me.

otoolep:
Thanks @m04f

PR #2122, opened 2025-06-22
Safe database swap

### LINKED ISSUE #2103: Swapping a database should not delete the files unless swapping worked
File that needs updating: https://github.com/rqlite/rqlite/blob/master/db/swappable_db.go

Function that needs updating: func (s *SwappableDB) Swap(path string, fkConstraints, walEnabled bool) error {

Today this function starts by deleting the old database files. What it should do is rename those files to a temp name. If the function succeeds, then actually delete them. If it is doesn't succeed, swap the renamed files back in.

In other words the function should completely succeed, or make no changes at all.

m04f:
This PR implements safe database swap functionality to address the data loss issue described in #2103.

### Testing Note
The recovery test uses `syscall.Mkfifo()` to create a file that passes initial SQLite header validation but fails during database opening. This  simulates a swap failure scenario, though I think won't run on Windows platforms. (This is the only way I had in mind to pass the IsValidSQLiteFile guard)

Closes #2103

### DIFF
--- a/db/swappable_db_test.go
+++ b/db/swappable_db_test.go
@@ -2,8 +2,11 @@ package db
 
 import (
 	"os"
+	"syscall"
 	"testing"
 )
+
+var sqliteHeader = []byte("SQLite format 3\x00")
 
 // Test_OpenSwappable_Success tests that OpenSwappable correctly opens a database and returns
 // a valid SwappableDB instance.
@@ -90,6 +93,11 @@ func Test_SwapSuccess(t *testing.T) {
 	if exp, got := `[{"columns":["id","name"],"types":["integer","text"],"values":[[1,"test"]]}]`, asJSON(rows); exp != got {
 		t.Fatalf("unexpected results after swap, expected %s, got %s", exp, got)
 	}
+
+	// Ensure the backup was deleted
+	if _, err := os.Stat(backupDbPath(swappablePath)); err == nil {
+		t.Fatalf("backup file should have been deleted")
+	}
 }
 
 func Test_SwapSuccess_Driver(t *testing.T) {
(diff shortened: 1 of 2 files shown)


db/swappable_db_test.go
```diff
@@ -164,3 +172,55 @@ func Test_SwapInvalidSQLiteFile(t *testing.T) {
...
+	defer os.Remove(swapPath)
+
+	err = syscall.Mkfifo(swapPath, 0666)
+	if err != nil {
+		t.Fatalf("failed to create swap file: %s", err)
+	}
+	defer os.Remove(swapPath)
+
+	// Avoid blocking the thread
+	go func() {
```
otoolep, line 203: Why this? Why a goroutine?

db/swappable_db_test.go
```diff
@@ -164,3 +172,55 @@ func Test_SwapInvalidSQLiteFile(t *testing.T) {
...
+	}
+	mustExecute(swappableDB.db, "CREATE TABLE original (id INTEGER NOT NULL PRIMARY KEY, name TEXT)")
+	mustExecute(swappableDB.db, `INSERT INTO original(name) VALUES("original_data")`)
+	defer swappableDB.Close()
+
+	// Create a file that would pass the IsValidSQLiteFile check but fail when opened
+	swapPath := mustTempPath()
+	defer os.Remove(swapPath)
+
+	err = syscall.Mkfifo(swapPath, 0666)
```
otoolep, line 196: If we modify Swap so that it always leaves the original database place if the swap fails then we can remove the SQLite validity check from the implementation. That was only there to try to protect us because Swap() was a one-way operation. But if we fix that we can remove the check.

db/swappable_db_test.go
m04f: Writing to a fifo blocks until the other end is opened for reading. Without the goroutine, the test would hang indefinitely at the write, since Swap() (which triggers the read attempt) hasn't been called yet.

otoolep: I see. Well, as per my other comment if we can improve this function such that if the passed in database doesn't open we just put the original DB back in, it might be OK to remove the SQLite check.

m04f:
Currently, the Swap method works by renaming the existing database files to a backup location, then renaming the new database into place, and finally attempting to open it. If opening the new database fails, it attempts to restore the original files. However, this backup-and-restore mechanism is itself fallible -- for example, the rename of WAL might fail after the database file has been renamed, potentially leading to data loss.

A potentially safer approach would be to open and validate the new database file before performing any file renames. This would guarantee that we only touch the original database files if we’re confident the new database is valid. However, this would mean that the active database path would change after the swap.

Would it be acceptable for the database path to change in this case? Based on a quick review of Swap's call sites, it doesn't seem like anything depends on the path staying the same, but I wanted to confirm.

otoolep:
No, you can't change the database path. It's a core assumption of rqlite that the path, once set, is fixed.

PR #2131, opened 2025-06-30
Initial FIFO implementation for CDC

otoolep:


cdc/fifo.go
```diff
@@ -0,0 +1,260 @@
...
+		// We call cond.Wait(), which does three things atomically:
+		// 1. Unlocks the mutex (q.mu).
+		// 2. Puts the current goroutine to sleep.
+		// 3. When woken up by q.cond.Signal(), it re-locks the mutex before returning.
+		// This atomic unlock/wait/relock prevents the "lost wakeup" problem.
+		q.cond.Wait()
+	}
+
+	// At this point, we have found an item (key and val are populated) and we still
+	// hold the mutex lock. Now we can safely delete it.
```
otoolep: Actually I don't want to delete. This should be done in response to highwatermarks.

cdc/fifo.go
```diff
@@ -0,0 +1,260 @@
...
+}
+
+// Close closes the underlying BoltDB database.
+func (q *Queue) Close() error {
+	return q.db.Close()
+}
+
+// Enqueue adds an item to the queue with a given index.
+// The FIFO order is determined by the index; lower indices are considered "earlier" in the queue.
+func (q *Queue) Enqueue(idx uint64, item []byte) error {
```
otoolep, line 116: Things to think about:

- only accept items which are not already present in the queue.
- only accept items which have a higher key than the last object already present in the queue (which will enforce the first point).

In other words while an attempt may be made to insert an object using an earlier or equal index to the last object in the queue, if an object is added with an index great than the last current object, that index must be exactly more more than the currently last index.

otoolep: Actually, not quite true. Some stuff goes through the Raft log, bumping the Raft Index, but would not result in a CDC event -- a strong query, for example.

Still, do not insert any older events. One an index is inserted, all future insertions must be a greater index.

otoolep: And when deleting due to high water marked, always leave the greatest index object in place, as a placeholder to prevent earlier insertion. Just skip it when reading for transmission.

PR #2167, opened 2025-07-16
Add logger interface to allow usage of custom loggers.

### LINKED ISSUE #2163: Logging in public packages
Hi, would you be open to exposing the logger setup in public packages?

For example in the current tcp.Mux implementation:
```go
func NewMux(ln net.Listener, adv net.Addr) (*Mux, error) {
	addr := adv
	if addr == nil {
		addr = ln.Addr()
	}

	return &Mux{
		ln:      ln,
		addr:    addr,
		m:       make(map[byte]*listener),
		Timeout: DefaultTimeout,
		Logger:  log.New(os.Stderr, "[mux] ", log.LstdFlags),
	}, nil
}
```
We could introduce a new function with options to set up logger as well
```go
type MuxOption interface{
	apply(*Mux)
}

func NewMuxWithOptions(ln net.Listener, adv net.Addr, opts ...MuxOption) (*Mux, error){
	mux, err := NewMux(ln, adv)
	for _, opt := range opts {
		opt.apply(mux)
	}
	return mux, err
}
```

What would be also helpful is to start using some logger interface for example `hclog.Logger`

In our project we are using quite a big chunk of rqlite public packages and currently we have no means to set up custom/structured logging due to usage of `log` package and hard coded loggers `log.New(os.Stderr, "[mux] ", log.LstdFlags)`.

richardbizik:
@otoolep this was the least invasive option I could think of that would allow us to use custom logging. 
This could be also used to provide logs in json format by adding cli flag and wrapping `slog`.

closes #2163

### DIFF
--- a/rqlog/std_log.go
+++ b/rqlog/std_log.go
@@ -1,0 +1,64 @@
+package rqlog
+
+import (
+	"io"
+	"log"
+)
+
+type RqLogger struct {
+	stdLog *log.Logger
+}
+
+func NewRqLogger(out io.Writer, prefix string, flag int) *RqLogger {
+	return &RqLogger{
+		stdLog: log.New(out, prefix, flag),
+	}
+}
+
+func (l *RqLogger) Print(v ...any) {
+	l.stdLog.Print(v...)
+}
+
+func (l *RqLogger) Printf(format string, v ...any) {
+	l.stdLog.Printf(format, v...)
+}
+
+func (l *RqLogger) Println(v ...any) {
+	l.stdLog.Println(v...)
+}
+
+func (l *RqLogger) Fatal(v ...any) {
+	l.stdLog.Fatal(v...)
+}
+
+func (l *RqLogger) Fatalf(format string, v ...any) {
+	l.stdLog.Fatalf(format, v...)
+}
+
+func (l *RqLogger) Fatalln(v ...any) {
+	l.stdLog.Fatalln(v...)
+}
+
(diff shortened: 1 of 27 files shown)


rqlog/std_log.go
```diff
@@ -0,0 +1,64 @@
+package rqlog
+
+import (
+	"io"
+	"log"
+)
+
+type RqLogger struct {
```
otoolep, line 8: This should be called simply Logger, because the package name itself is `rq`.

otoolep: Oh, hang on, I see you have named the interface that. Hmmm.

otoolep: OK, well following Go style, should the type be simply "Log", and constructure "New()". Because the package name is "rq" you don't need to include it in the types and functions inside the package.

richardbizik: If we don't want to add another implementation right now like `NewJsonLogger` I can rename it to just `New()` and `Log`. If we want another implementation I would suggest something like `NewStandardLogger` and `StandardLogger` for the struct name.

otoolep:
Hi @richardbizik -- thanks for this.

It's not that invasive, I agree, but I've decided not to merge it. My primary goal is rqlite development itself (or the functionality the database offers) and this change doesn't do anything to help those goals. This adds a layer of indirection which is unneeded as far as rqlite is concerned. Hope that makes sense.

PR #2396, opened 2025-12-03
Kkostial/minor refactorings

The127:
This PR fixes a few static code analysis issues mostly regarding comment formatting, import aliases, unreachable code, etc.
Each commit tackles one such issue and touches files with similar issues.
I'm fine with breaking this up PR differently if required.


otoolep:
Thanks for your contribution, couple of bad copy-n-paste issues in there.

PR #2400, opened 2025-12-04
Run rqlited as a non-root user within Docker

otoolep:


otoolep:
This also fixes the exposed ports.

Dockerfile
```diff
@@ -66,9 +71,17 @@ COPY --from=builder /app/rqlite /bin
...
+RUN mkdir -p /rqlite/file && \
+    chown -R rqlite:rqlite /rqlite/file
+
 VOLUME /rqlite/file
-EXPOSE 4001 4001
+
+EXPOSE 4001 4002
+
+# Switch to the non-root user.
+USER rqlite
```
eschbach-dd, line 84: Why not switching user before creating the dir?

otoolep: User `rqlite` does not have permission to make directories on the root file system.
```
0.938 mkdir: can't create directory '/rqlite/': Permission denied
```

PR #2427, opened 2026-01-04
Remove temporary WAL file if snapshot persist fails

otoolep:


otoolep:
Clean up scratch WAL files as shown in https://github.com/rqlite/rqlite/discussions/2414

PR #2450, opened 2026-01-29
Upgrade SQL parser

otoolep:


otoolep:
> ## Code Review
> 
> This pull request correctly upgrades the `github.com/rqlite/sql` dependency. The changes across `go.mod`, `go.sum`, and `CHANGELOG.md` are all in order.
> 
> While reviewing the provided context, I found a potential bug in `command/sql/processor.go` that is worth addressing in a follow-up. The `ContainsTime` function, which is part of a fast-path optimization, fails to detect the `strftime` function. As a result, SQL statements using `strftime('...', 'now')` will not be rewritten with a deterministic timestamp, potentially causing data inconsistency across the cluster.
> 
> A simple fix would be to add `"strftime("` to the `targets` in `ContainsTime`. This would also resolve what appears to be a failing test case in `Test_Time_Rewrites`.
> 
> Since this file isn't in the diff, I'm just noting it here. The current PR is good to go.

Wrong. The test for ""time(" will catch it.

PR #2511, opened 2026-03-05
command,db: add error_code field to API responses

### LINKED ISSUE #2271: Concrete errors codes from API?
Add a new field, error_code, which might contain a code from the SQLite source code. e.g. SQLITE_MISUSE.

TanayK07:
Add an error_code integer field alongside the existing error string in query and execute results. When a statement fails with a SQLite error, the extended error code (e.g. SQLITE_CONSTRAINT_UNIQUE = 2067) is now included in the response, allowing clients to programmatically handle specific error conditions.

Changes:
- Add error_code to ExecuteResult and QueryRows protobuf messages
- Extract SQLite extended error code via errors.As in db layer
- Propagate error_code through JSON encoding structs

The error_code field uses omitempty, so it only appears when non-zero. Fully backward compatible — the existing error string is unchanged.

Fixes #2271


otoolep:
Thanks for your contribution, however it's clear this isn't actually having any effect. For example this unit test should fail but it doesn't:

https://github.com/rqlite/rqlite/blob/master/db/db_common_test.go#L1104

I am interested in merging this (or something like it), but it needs more work.

otoolep:
I add some more details about this issue here: https://github.com/rqlite/rqlite/issues/2271#issuecomment-4005527064

You're on the right track, but the change will be more complicated that what you're done on this first round.

otoolep:
I suggest you proceed in steps. Just modify the ExecuteResponse first, get everything working (tests updated, etc) and then prepare that PR. Then we can move to Query and Request. But if you try do everything at once, it may be difficult.

command/proto/command.pb.go
```diff
@@ -1,7 +1,7 @@
 // Code generated by protoc-gen-go. DO NOT EDIT.
 // versions:
 // 	protoc-gen-go v1.36.11
-// 	protoc        v6.33.0
+// 	protoc        v4.25.1
```
otoolep: Your protobuf compiler looks too old. You will need to upgrade this tool chain.

Alternatively you can proceed as is, I will be rebuild the proto files once this merges to master.

TanayK07:
Thanks for the detailed feedback — really appreciate it!

I've reworked this PR based on your suggestions:

**Structured `error_v2` instead of flat `error_code`** — The response now lifts the full `sqlite3.Error` struct from `go-sqlite3`, so you get `code`, `extended_code`, `system_errno`, and `message`:

```json
{
  "results": [
    {
      "error": "UNIQUE constraint failed: foo.id",
      "error_v2": {
        "code": 19,
        "extended_code": 1555,
        "system_errno": 0,
        "message": "UNIQUE constraint failed: foo.id"
      }
    }
  ]
}
```

**Scoped to Execute path only** — As you suggested, I'm taking this incrementally. This PR only touches `ExecuteResult`. Query and Request paths will come in follow-up PRs.

**Bug fixes** — Fixed `handleError` and the defer in `executeStmtWithConn` so they no longer overwrite the structured error with a plain string. This was why the tests weren't catching the missing error codes before.

**Proto** — I've added the `SQLiteError` message and `error_v2` field to `command.proto` and manually wired up the Go struct in `command.pb.go` (with a `SQLiteError` Go-only struct in `sqlite_error.go`). Happy for you to regenerate with protoc once this looks good.

Let me know if anything needs adjusting!

otoolep:
Tests need to be all green before I will review in any detail.

otoolep:
That test failure looks unrelated, I'm looking into a fix.

otoolep:
You should merge master to your branch.

TanayK07:
Hi @otoolep , took me some time to fix the CIs and pass through the tests, can you PTAL when you are available. Thanks

TanayK07:
> Hi @otoolep , took me some time to fix the CIs and pass through the tests, can you PTAL when you are available. Thanks


hi @otoolep just wanted to ping as a reminder. Thanks

otoolep:
Thanks @TanayK07 - what is your response to the feedback from Gemini at https://github.com/rqlite/rqlite/pull/2511/changes#r2890345835 ?

TanayK07:
Hey @otoolep, good question — I actually went back and looked at the Gemini feedback again carefully.

**On the "error_code is lost" bug:** This was a valid concern on the orginal version of the PR, but it's been fixed since. The `handleError` in `executeWithConn` now checks `if result.Result == nil` before overwriting, so it won't clobber the structured `ExecuteResult` that `executeStmtWithConn` sets. Similarly, the `defer` in `executeStmtWithConn` now only overwrites the result if the error was actually rewritten (e.g. to a timeout) or if nothing was set yet. The unit tests (like `Test_DB_SimpleFailingStatements_Execute`) verify that `error_v2` makes it all the way through — including for UNIQUE constraint violations, syntax errors, etc.

**On the "breaking change" concern:** The Gemini comment was pointing out that the `oneof result` in proto switches from the `error` string variant to the `ExecuteResult` variant (`e`) for statement errors. That's technically true at the protobuf level, but rqlite doesn't expose a public gRPC API — the protobuf is only used internally for cluster communication between nodes. The actual public API is HTTP/JSON, and there the change is purely additive: the `error` string field is still there unchanged, `error_v2` is a new field with `omitempty`. Any existing client parsing `error` keeps working exactly as before.

So basically Gemini was reviewing the first iteration of the code and its feedback was accurate for that version, but both issues have been addressed in the reworked version.


Heh, I was just online to see this(unrelated)

command/proto/command.proto
```diff
@@ -69,6 +76,7 @@ message ExecuteResult {
 	int64 rows_affected = 2;
 	string error = 3;
 	double time = 4;
+	SQLiteError error_v2 = 5;
 }
```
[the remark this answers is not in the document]
TanayK07: Valid but cosmetic

db/db.go
```diff
@@ -1067,8 +1075,11 @@ func (db *DB) executeStmtWithConn(ctx context.Context, stmt *command.Statement,
...
 		if err != nil {
-			response.Result = &command.ExecuteQueryResponse_Error{
-				Error: err.Error(),
+			response.Result = &command.ExecuteQueryResponse_E{
+				E: &command.ExecuteResult{
+					Error:   err.Error(),
+					ErrorV2: sqliteErrorV2(err),
+				},
 			}
 			return response, err
```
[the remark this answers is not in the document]
TanayK07: intentionally out of scope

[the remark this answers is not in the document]
TanayK07: for now

otoolep:
Thanks @TanayK07 -- I will review in the coming couple of weeks.

db/db.go
```diff
@@ -982,8 +982,10 @@ func (db *DB) executeWithConn(ctx context.Context, req *command.Request, xTime b
 	// whether the caller should continue processing or break.
 	handleError := func(result *command.ExecuteQueryResponse, err error) bool {
 		stats.Add(numExecutionErrors, 1)
-		result.Result = &command.ExecuteQueryResponse_Error{
-			Error: err.Error(),
+		if result.Result == nil {
```
otoolep, line 1059: Why was this `if` clause added?

db/db.go
```diff
@@ -1027,10 +1029,16 @@ func (db *DB) executeWithConn(ctx context.Context, req *command.Request, xTime b
 func (db *DB) executeStmtWithConn(ctx context.Context, stmt *command.Statement, xTime bool, eq execerQueryer, timeout time.Duration) (res *command.ExecuteQueryResponse, retErr error) {
 	defer func() {
 		if retErr != nil {
+			origErr := retErr
 			retErr = rewriteContextTimeout(retErr, ErrExecuteTimeout)
 			if res != nil {
-				res.Result = &command.ExecuteQueryResponse_Error{
-					Error: retErr.Error(),
+				// Only overwrite the result if the error was rewritten (e.g., timeout),
```
otoolep, line 1109: Is this change required for this PR?

db/db.go
```diff
@@ -1399,10 +1410,16 @@ func (db *DB) RequestWithContext(ctx context.Context, req *command.Request, xTim
...
-				Result: &command.ExecuteQueryResponse_Error{
-					Error: err.Error(),
+				Result: &command.ExecuteQueryResponse_E{
+					E: &command.ExecuteResult{
+						Error:   err.Error(),
+						ErrorV2: sqliteErrorV2(err),
+					},
 				},
 			})
+			if abortOnError(err) {
```
otoolep: Why is this change here?

db/db_commit_hook_test.go
```diff
@@ -58,7 +58,7 @@ func Test_CommitHook_Rollback(t *testing.T) {
 	if err != nil {
 		t.Fatalf("error executing statement: %s", err.Error())
 	}
-	if r[0].GetError() == "" {
+	if r[0].GetError() == "" && r[0].GetE().GetError() == "" {
```
otoolep, line 61: Was this a necessary change?

db/db_preupdate_test.go
```diff
@@ -461,7 +461,7 @@ func Test_Preupdate_Constraint(t *testing.T) {
 
 	// Insert a row with the same ID, should not trigger the hook.
 	r, err := db.ExecuteStringStmt("INSERT INTO foo(id, name) VALUES(5, 'fiona2')")
-	if err == nil && r[0].GetError() == "" {
+	if err == nil && r[0].GetError() == "" && r[0].GetE().GetError() == "" {
```
otoolep, line 464: Why this change? What are you concerned about?

store/store_test.go
```diff
@@ -898,7 +898,11 @@ func Test_SingleNodeExecuteQueryFail(t *testing.T) {
 	if err != nil {
 		t.Fatalf("failed to execute on single node: %s", err.Error())
 	}
-	if exp, got := "no such table: foo", r[0].GetError(); exp != got {
+	errMsg := r[0].GetError()
```
otoolep, line 903: I agree that the code should check the presence of the new error_v2 object, but it's not clear to me why you check it here. Is there a specific reason?

otoolep (commented):
Thanks for this, I think this change will make it easier to robustly code against rqlite. But since this is an addition to the API, we must be very focused on the details.

My plan will be to make this new error information be part of the v10 release.

db/db.go
TanayK07: the PR changes executeStmtWithConn to return ExecuteQueryResponse_E (with the        
  structured ExecuteResult including ErrorV2) instead of ExecuteQueryResponse_Error. Without this guard,
  handleError would overwrite the structured error with a plain string error, losing the ErrorV2 data as per my understanding, I might be wrong here

db/db.go
TanayK07: the defer block would unconditionally overwrite res.Result with a 
  plain _Error string, which kills the structured ExecuteResult (with ErrorV2) that the ExecContext path just 
  set. So basically the origErr comparison is there to say: if the error got rewritten to a timeout, we do  
  want to overwrite

db/db.go
TanayK07: he `ErrorV2` part is for consistency with the other execute error paths, so `sqliteErrorV2()` can extract the structured error if `StmtReadOnlyWithConn` happens to return a SQLite error.

The `abortOnError` yeah looking at it again, that's more of a consistency fix (the query and execute paths already call it) and not strictly related to `error_v2`. I can pull that out into a separate commit or drop it entirely from this PR if you'd prefer to keep scope tight. Let me know.

db/db_commit_hook_test.go
TanayK07: he execute error path now returns `ExecuteQueryResponse_E` (with the `ExecuteResult` containing both `Error` and `ErrorV2`) instead of `ExecuteQueryResponse_Error`. So `GetError()` alone returns empty since the result is in a different oneof variant. The test needs to check both to not get a false pass.

To be clear though, this only affects the **internal protobuf** rhe JSON API still has `error` as a plain string field alongside the new `error_v2`, so existing HTTP clients don't break.

db/db_preupdate_test.go
TanayK07: constraint violations (like duplicate ID / unique constraint) now go through the `ExecContext` path which returns `_E` with the structured error instead of `_Error`. Without checking `GetE().GetError()`, the test would silently pass even if the constraint violation didn't actually produce an error, which is the opposite of what we want here.

store/store_test.go
TanayK07: this test executes an INSERT on a non-existent table (`foo`), which goes through the `ExecContext` path. Since that path now returns `ExecuteQueryResponse_E` instead of `ExecuteQueryResponse_Error`, `GetError()` at the top level comes back empty. The fallback to `GetE().GetError()` is needed to actually extract the error string for the assertion.

Basically any test that exercises the execute error path needs this two-step check now, since the oneof variant changed. This was one of the tests that broke without it.

otoolep:
Can you resolve conflicts? Let me know if you have any questions.

TanayK07:
> Can you resolve conflicts? Let me know if you have any questions.
hI, I have resolved merge conflicts. 

2 questions --

1. currently drop error_v2  a client hitting a follower gets [{"error":"no such table: qux"}] while a client hitting the leader gets the full structured form (commit c6e18429 ) reverts a test to document this). That feels like a blocker for a clustered deployment. Should I propagate error_v2 through the cluster RPC in this PR, or is the leader-only behavior acceptable for v1?

2. Should I change the PR title and PR description?

otoolep:
For the feature to be complete it has to be propagated, but that can be a follow up PR. Let's focus on the non-cluster version first, ensure we are happy with the API changes. Once we're happy, and it merges, we can then do clustering.

TanayK07:
Makes sense. 

One API thing worth calling out before you dig in  SQLiteError.message is just sqErr.Error(), and the top-level error is err.Error(). For SQLite errors they end up as the exact same string in every response.                                                                                                                                                                                              

Options in my mind- 
 
  a. Keep both being redundant but ergonomic.                                                                                                                                                                
 b. Drop message from SQLiteError since callers already have error at the top level.
 c. Eventually deprecate top-level error and let clients migrate to error_v2.message.

otoolep:
I am pretty certain option c) will never happen. The API hasn't been changed in a breaking manner in over 10 years, and I don't plan to change that. It's difficult enough to get client code support, making all the existing client code require changes? Will never happen.

otoolep:
It's why I'm considering ultimately putting this change behind a query param. I don't want to jeopardize breaking the existing user base (though clearer error codes should have been in from the start - it was an oversight)

TanayK07:
Thanks. To keep to your suggestion of stepping through Execute to  Query/Request to  cluster in separate PRs,

 I'll scope this one back to Execute only (reverting the Request-path commit and the unrelated abortOnError change).

If I am correct in understanding

TanayK07:
questions before I push  also for future scope -- 
 1. Drop message from SQLiteError? Top-level error already carries the string. Keeping only code, extended_code, system_errno in error_v2 avoids redundancy. Does something like this work?           
                              
  2. Query-param gate — name and scope? Proposal: ?error_v2 on /db/execute in this PR; same flag extended to /db/query and /db/request in   default off, so existing responses are byte-identical.

otoolep:
> questions before I push also for future scope --
> 
> 1. Drop message from SQLiteError? Top-level error already carries the string. Keeping only code, extended_code, system_errno in error_v2 avoids redundancy. Does something like this work?

Yeah, I saw drop it.

> 2. Query-param gate — name and scope? Proposal: ?error_v2 on /db/execute in this PR; same flag extended to /db/query and /db/request in   default off, so existing responses are byte-identical.

Right, that's what I was thinking -- thanks.

otoolep:
If we agree it will default to off, then we can consider doing this change in steps. First we concentrate on the model model changes. Agree what the changes are there, and the low level gather of the information by the DB layer. We merge that first, then we proceed to integration with the Store and HTTP layers.

This is how I tend to work when I do similar work. Put the base-level infra in place (in this case data modeling and DB layer), then move to the rest of the feature. It makes it easier to review what is really the code piece -- the data model.

TanayK07:
Agreed on both points. The layered split makes sense. I will scope this PR down to just the proto message changes (code, extended_code, system_errno, and no message, right?) along with the DB-layer extraction in the Execute path, and include DB-layer tests only.
The Store, HTTP, and the ?error_v2 gate can go into the next PR. I will push the updated branch shortly.

TanayK07:
hi @otoolep ,

pushed a trimmed-down version of the branch. removed the message field from SQLiteError (it’s just code, extended_code, system_errno now), and dropped the Request-path and abortOnError changes. kept it limited to the proto model and the DB-layer ExecContext extraction. 

one thing to flag: ForceQuery (RETURNING) errors still go through the plain _Error path, so they won’t have error_v2 yet. felt okay leaving that for later at this layer.

if this looks good and we merge it, i was thinking the next PR would wire up the Store and HTTP layers behind an ?error_v2 query param (off by default), and then follow up with Query and Request paths.

does that sequencing sound reasonable?

PR #2516, opened 2026-03-06
Add table-qualified column names for JOIN queries

### LINKED ISSUE #530: Include table name at least for joins in the resulting columns array
If you join two tables in a query the resulting columns array has ambiguous columns. 

Query:
`select * from "contacts" join "titles" on "contacts"."title_id" = "title"."id" `

Result:
` ['id', 'name', 'first_name', 'created_at', 'updated_at', 'title_id', 'id', 'name', 'created_at', 'updated_at']`

Should be something like:
`  ['contact.id', 'contact.name', 'contact.first_name', 'contact.created_at', 'contact.updated_at', 'contact.title_id', 'title.id', 'title.name', 'title.created_at', 'title.updated_at']`

TanayK07:
## Summary

Resolves #530.

When doing `SELECT * FROM contacts JOIN titles ON ...`, the `columns` array currently returns ambiguous names like `["id", "name", "id", ...]`. This PR adds an opt-in `?qualify` query parameter that prefixes each column with its originating table name:

```json
// Before: /db/query?q=SELECT * FROM contacts JOIN titles ON ...
{"columns": ["id", "name", "id", "contact_id", "title"]}

// After: /db/query?q=SELECT * FROM contacts JOIN titles ON ...&qualify
{"columns": ["contacts.id", "contacts.name", "titles.id", "titles.contact_id", "titles.title"]}
```

### Approach

Uses SQLite's C API [`sqlite3_column_table_name()`](https://www.sqlite.org/c3ref/column_database_name.html) which is already exposed in `rqlite/go-sqlite3` as `SQLiteStmt.ColumnTableName()`. This was the approach [suggested in #530](https://github.com/rqlite/rqlite/issues/530#issuecomment-2804063558) by @zmedico.

The feature is gated behind the `sqlite_column_metadata` build tag (which enables `SQLITE_ENABLE_COLUMN_METADATA` in the SQLite compilation). Without the tag, `?qualify` is silently ignored — no breaking change.

### Implementation

- **Proto**: Added `qualify_columns` field to `QueryRequest` and `ExecuteQueryRequest`
- **HTTP**: New `?qualify` query parameter parsed via `QualifyColumns()`
- **Store → DB**: Flag flows via `context.Value` — no function signature changes needed
- **DB**: `qualifyRowColumns()` uses `conn.Raw()` to access the driver-level `SQLiteStmt.ColumnTableName()` and prefix columns
- **Build tag**: `db/qualify_columns.go` (real impl) and `db/qualify_columns_omit.go` (no-op fallback)
- **Dockerfile**: Updated to build with `-tags sqlite_column_metadata`

Works for all consistency levels including strong reads through Raft, and for both `/db/query` and `/db/request` endpoints.

### Comparison with PR #2116

PR #2116 uses SQL string parsing + `PRAGMA table_info()` to guess column→table mapping. That approach only works for `SELECT *`, uses fragile parsing, and doesn't leverage SQLite's built-in metadata. This PR uses the proper SQLite C API instead.

## Test plan

- [x] 5 new unit tests: JOIN, single table, explicit columns, LEFT JOIN, Request path
- [x] All existing `db`, `command`, `http`, and `store` tests pass
- [x] `go vet` passes
- [x] Builds with and without `sqlite_column_metadata` tag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,17 +1,25 @@
 ## v10.0.0 (unreleased)
-_Not ready for production use._
 
 This release introduces a major improvement to the Raft [_Snapshot and Log Truncation_](https://youtu.be/8XbxQ1Epi5w?t=492) process. Previously the Snapshotting process would be blocked on the Leader if the Leader was streaming a previously-taken Snapshot to a Follower. In most cases this wasn't an issue, as the Snapshotting process would be retried later. However in the event of a continuously slow Follower which continually needed Snapshots, this eventually starved the Leader of the ability to Snapshot. With v10 this is no longer the case and Snapshotting on the Leader is now decoupled from streaming pre-existing Snapshots to other nodes.
 
 This release also introduces a new built-in console app, making it more convenient to work with an rqlite system. 
 
-There are no breaking API changes in this release, nor any changes to how clustering operates. However the command line option `-on-disk-path` has been removed as it provided little benefit, but made it too easy for operators to inadvertently corrupt a rqlite node.
+There are no breaking API changes in this release, nor any changes to how clustering operates. However there have been some command line option changes:
+ - `-on-disk-path` has been removed as it provided little benefit, but made it too easy for operators to inadvertently corrupt a rqlite node.
+ - `-raft-timeout` has been renamed to `-raft-heartbeat-timeout` to better reflect its purpose.
 
-Upgrading to this release from a v7 release (or later) is seamless and been extensively tested. Rolling upgrades are also supported. However it is strongly recommended you [backup your rqlite system](https://rqlite.io/docs/guides/backup/) before upgrading it. **Downgrading from v10 is not supported**, and that is why the major release number has been incremented. If you do need to downgrade then backup your v10 system and use that backup to deploy a new pre-v10 system.
+ **Upgrading to v10**
+
+Upgrading to this release from a v7 release (or later) is seamless and been extensively tested. Rolling upgrades are also supported, but **joining a new v10 node to a v9 (or earlier) cluster is not supported**. Upgrade your cluster to v10 first before adding any new nodes. As always, it is strongly recommended you [backup your rqlite system](https://rqlite.io/docs/guides/backup/) before upgrading it.
+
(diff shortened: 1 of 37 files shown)


otoolep:
Thanks for your contribution. However if changes are needed in the SQLite code and build time, they should be made in the fork rqlite manages. See https://github.com/rqlite/go-sqlite3

Make the required changes there first, and then we can make a new release if those changes merge to its master.

otoolep:
In other words we should change the build in the fork so that it builds with that flag enabled permanently going forward. Start there.

otoolep:
Summary: this is somewhat on the right track, but will not be merged.  Propose a change, via a PR, to rqlite/go-sqlite3 which enables any build flags permanently going forward. If accepted, we can upgrade this repo to a new release of rqlite/go-sqlite3.

Then we start a new PR to address the issue.

TanayK07:
Hey, I've opened a PR on go-sqlite3 to enable `SQLITE_ENABLE_COLUMN_METADATA` permanently — https://github.com/rqlite/go-sqlite3/pull/45

It just removes the `sqlite_column_metadata` build tag so the flag is always compiled in, similar to how `SQLITE_ENABLE_PREUPDATE_HOOK` is already handled. Once that's merged and tagged, I'll open a fresh PR here with the updated dependency and the rest of the changes


Is this something that works? Kindly LMK , thanks. @otoolep

otoolep:
Released and upgrading this repo to the latest code: https://github.com/rqlite/rqlite/pull/2534

Tanay10x:
Thanks! I've merged master (with the go-sqlite3 v1.45.0 upgrade) into this branch. All tests pass locally — both the standard suite and the qualify-specific tests with the `sqlite_column_metadata` build tag.

TanayK07:
Superseded by #2574 — cleaned up to remove the build-tag split now that `rqlite/go-sqlite3` v1.45.0 enables `SQLITE_ENABLE_COLUMN_METADATA` permanently. The new PR has a single implementation file (no `qualify_columns_omit.go` stub), no Dockerfile changes, and a clean diff against current master.

Issue #45, opened 2015-05-13
Best way to implement a whitelist of nodes for a cluster(IE only let certain IPs join the rqlite cluster)?


tchoulihan:
Hey @otoolep , I'm not sure if this is an issue or not, but I'd like your expertise on the best way to implement a whitelist of nodes.

So lets say I only want **certain** IP addresses or connection strings to be able to join the network, for security reasons. 

One solution I thought about is to just use an abstraction layer in my own project, to pass a parameter like:
`-whitelist IP_ADDR_1 IP_ADDR_2 ` , etc. 

I can do this in my own project as an abstraction layer, but let me know if this is something you'd be interested in building, and I won't do this.

otoolep:
Hello - no, I would consider network security outside the scope of rqlite. It is very unlikey I would implement something like this. 

There are lots of other options. Firewalls, iptabled, EC2 security groups, etc.

tchoulihan:
@otoolep No problem, very understandable.

PR #2574, opened 2026-03-20
db,http,store: add table-qualified column names for JOIN queries

### LINKED ISSUE #530: Include table name at least for joins in the resulting columns array

TanayK07:
## Summary

Resolves #530. Supersedes #2516 with a cleaner implementation.

Adds an opt-in `?qualify` query parameter that prefixes each column name with its originating table name:

```
// Before: /db/query?q=SELECT * FROM contacts JOIN titles ON ...
{"columns": ["id", "name", "id", "contact_id", "title"]}

// After: /db/query?q=SELECT * FROM contacts JOIN titles ON ...&qualify
{"columns": ["contacts.id", "contacts.name", "titles.id", "titles.contact_id", "titles.title"]}
```

### Approach

Uses SQLite's `sqlite3_column_table_name()` C API, exposed as `SQLiteStmt.ColumnTableName()` in `rqlite/go-sqlite3`. This is the approach originally suggested in #530 by @zmedico.

Since `SQLITE_ENABLE_COLUMN_METADATA` is now permanently enabled in `rqlite/go-sqlite3` v1.45.0 (merged via #45, released 2026-03-08, upgraded in this repo via #2534), there is no need for build tags — the API is always available.

### What changed from #2516

- **Removed the build-tag split entirely** — no `//go:build sqlite_column_metadata` guards, no `qualify_columns_omit.go` no-op stub, no `-tags` in Dockerfile
- **Single file** `db/qualify_columns.go` with both the context helpers and the implementation
- **Clean diff** — only the qualify-columns changes, no merge noise from upstream

### Implementation

- **Proto:** Added `qualify_columns` field (tag 7) to `QueryRequest` and `ExecuteQueryRequest`
- **HTTP:** New `?qualify` query parameter via `QualifyColumns()`
- **Store → DB:** Flag flows via `context.Value` — no function signature changes
- **DB:** `qualifyRowColumns()` uses `conn.Raw()` to access `SQLiteStmt.ColumnTableName()` and prefix columns
- Works for all consistency levels (including strong reads through Raft) and both `/db/query` and `/db/request`

## Test plan

- [x] 5 new unit tests: JOIN, single table, explicit columns, LEFT JOIN, Request path
- [x] All existing db, http, store, and command tests pass
- [x] `go vet ./...` passes
- [x] `go fmt ./...` clean


db/qualify_columns.go
```diff
@@ -0,0 +1,55 @@
...
+		return nil
+	})
+	if err != nil {
+		return err
+	}
+	for i, col := range rows.Columns {
+		if tableNames[i] != "" {
+			rows.Columns[i] = tableNames[i] + "." + col
+		}
+	}
```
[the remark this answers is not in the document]
TanayK07: Thanks for flagging this! I investigated and this is actually a false positive — the double-qualification doesn't occur.

SQLite's `Stmt.Columns()` always returns the **raw/alias** column name, never the table-qualified form from the SELECT statement. So even with `SELECT contacts.id FROM contacts`, `rows.Columns[i]` is `"id"` (not `"contacts.id"`), and after qualification it correctly becomes `"contacts.id"`.

I verified with tests:

| Query | `rows.Columns` (from SQLite) | After qualify |
|---|---|---|
| `SELECT contacts.id FROM contacts` | `["id"]` | `["contacts.id"]` ✓ |
| `SELECT contacts.id AS cid FROM contacts` | `["cid"]` | `["contacts.cid"]` ✓ |
| `SELECT 1+1 AS calc, name FROM contacts` | `["calc", "name"]` | `["calc", "contacts.name"]` ✓ |

The dot-in-column-name case would only matter for computed/aliased columns that contain dots, which isn't a realistic scenario in SQLite.

db/qualify_columns.go
```diff
@@ -0,0 +1,55 @@
...
+	return context.WithValue(ctx, qualifyColumnsKey, true)
+}
+
+func shouldQualifyColumns(ctx context.Context) bool {
+	v, _ := ctx.Value(qualifyColumnsKey).(bool)
+	return v
+}
+
+// qualifyRowColumns prefixes each column name in rows with its originating table name.
+func (db *DB) qualifyRowColumns(conn *sql.Conn, query string, rows *command.QueryRows) error {
```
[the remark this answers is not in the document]
TanayK07: Good question. Yes, the tradeoff was considered acceptable:

1. **Opt-in only** — the re-prepare only runs when the user explicitly passes `?qualify`, so there's zero overhead for normal queries.

2. **No alternative** — SQLite's `sqlite3_column_table_name()` C API requires a prepared statement handle. There's no way to get column-to-table mappings from result rows alone. This is the approach suggested in the original issue #530 by @zmedico, referencing the [SQLite docs](https://www.sqlite.org/c3ref/column_database_name.html).

3. **Prepare-only, no execute** — we only call `Prepare()` (parsing + planning), not `Step()` (execution). The query was already executed on the same connection, so SQLite's query plan cache should make this very fast.

TanayK07:
Hi @otoolep PTAL when you get the chance? Thanks

otoolep:
Thanks @TanayK07 -- I will review in the coming couple of weeks.

command/proto/command.proto
```diff
@@ -45,6 +45,7 @@ message QueryRequest {
 	int64 freshness = 4;
 	bool freshness_strict = 5;
 	int64 linearizable_timeout = 6;
+	bool qualify_columns = 7;
```
otoolep: This field should be on the `Request` object. It's similar to `transaction' as it affects the SQL statements, not how the request is processed by the cluster.

command/proto/command.proto
```diff
@@ -78,6 +79,7 @@ message ExecuteQueryRequest {
...
```
otoolep: Same, shouldn't be here. Should be on the `Request` object.

db/qualify_columns.go
otoolep: From a code construction point of view I don't see why this method is broken out into a separate source file. It should be part of db.go to be consistent.

If you want it to be a separate function then make it take a DB object. Don't place it on the DB object -- you will see that pattern in a few places. NormalizeValues for example (IIRC).

db/qualify_columns.go
```diff
@@ -0,0 +1,55 @@
...
+	sqlite3 "github.com/mattn/go-sqlite3"
+	command "github.com/rqlite/rqlite/v10/command/proto"
+)
+
+type qualifyColumnsKeyType struct{}
+
+var qualifyColumnsKey = qualifyColumnsKeyType{}
+
+// NewContextWithQualifyColumns returns a new context with the qualify columns flag set.
+func NewContextWithQualifyColumns(ctx context.Context) context.Context {
```
otoolep: This approach is not consistent with the rest of the code. Once you update the proto you will have access to the filed via the request object itself. You don't need to do anything with Context objects.

http/query_params_test.go
```diff
@@ -40,6 +40,7 @@ func Test_NewQueryParams(t *testing.T) {
 		{"Sync with timeout", "sync&timeout=2s", QueryParams{"sync": "", "timeout": "2s"}, false},
 		{"Byte array with associative", "byte_array&associative", QueryParams{"byte_array": "", "associative": ""}, false},
 		{"Requesting Raft Index", "raft_index", QueryParams{"raft_index": ""}, false},
+		{"Qualify columns", "qualify", QueryParams{"qualify": ""}, false},
```
otoolep: Let's be more explicit -- the query param should be `qualify_columns`.

otoolep (changes requested):
This appears fundamentally correct, but is not following the patterns established in the code. Move the flag to the `Request` object and it will be easier to follow the existing patterns.

command/proto/command.proto
TanayK07: moved it to `Request` alongside `transaction`. Makes much more sense there.

command/proto/command.proto
TanayK07: Done, both removed.

db/qualify_columns.go
TanayK07: moved it into `db.go` as a standalone function. It didn't even use the receiver, so it's just `qualifyRowColumns(conn, query, rows)` now.

db/qualify_columns.go
TanayK07: removed the context plumbing entirely. It reads `req.QualifyColumns` directly now, same pattern as `req.Transaction`.

http/query_params_test.go
TanayK07: renamed to `qualify_columns`.

TanayK07:
caSignedCertExampleDotCom expired March 26, 2026 . You might want to get thsi fixed, this is making one of the e2e tests fail

otoolep:
Merge master into your branch, the cert is fixed there.

command/proto/command.proto
```diff
@@ -34,6 +34,7 @@ message Request {
 	repeated Statement statements = 2;
 	int64 dbTimeout = 3;
 	bool rollbackOnError = 4;
+	bool qualify_columns = 5;
```
otoolep: Inconsistent style - should be qualifyColumns

store/command_processor.go
```diff
@@ -54,7 +55,7 @@ func (c *CommandProcessor) Process(data []byte, db *sql.SwappableDB) (*proto.Com
 		if err := command.UnmarshalSubCommand(cmd, &qr); err != nil {
 			panic(fmt.Sprintf("failed to unmarshal query subcommand: %s", err.Error()))
 		}
-		r, err := db.Query(qr.Request, qr.Timings)
+		r, err := db.QueryWithContext(context.Background(), qr.Request, qr.Timings)
```
otoolep: This change is no longer necessary.

store/command_processor.go
```diff
@@ -68,7 +69,7 @@ func (c *CommandProcessor) Process(data []byte, db *sql.SwappableDB) (*proto.Com
 		if err := command.UnmarshalSubCommand(cmd, &eqr); err != nil {
 			panic(fmt.Sprintf("failed to unmarshal execute-query subcommand: %s", err.Error()))
 		}
-		r, err := db.Request(eqr.Request, eqr.Timings)
+		r, err := db.RequestWithContext(context.Background(), eqr.Request, eqr.Timings)
```
otoolep: Same.

otoolep:
This looks good, I'm ready to merge. Let me know when you're finished.

TanayK07:
> This looks good, I'm ready to merge. Let me know when you're finished.

Hi, yes good  and done from my end. Thanks

otoolep:
Thanks for your contribution.

TanayK07:
> Thanks for your contribution.

Looking forward to doing more of it over time

otoolep:
This functionality will be part of the v10 release. I'll update the docs at rqlite.io when that release ships.

PR #2622, opened 2026-04-21
Escape single quotes in VacuumInto path

### LINKED ISSUE #2621: VacuumInto is SQL-injectable
db.go:847 — path interpolated directly into SQL via fmt.Sprintf. A path with single quote breaks it. SQLite doesn't support parameterized VACUUM INTO.

This is not an issue in practise as the path is not user-controllable. But is an important DB layer API fix.

mahi-ma:
## Summary
- Fix SQL injection vulnerability in `VacuumInto` by escaping single quotes in the file path (doubling `'` → `''` per SQL standard)
- SQLite does not support parameterized `VACUUM INTO`, so string escaping is the correct approach
- Added a test case that vacuums into a path containing a single quote and verifies the resulting database

Fixes #2621

## Test plan
- [x] Existing `Test_DBVacuumInto` tests pass
- [x] New test case with single-quote path passes
- [x] Full `db` package tests pass (119/119)
- [x] `go vet ./...` clean

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -847,7 +847,7 @@ func (db *DB) Vacuum() error {
 
 // VacuumInto VACUUMs the database into the file at path
 func (db *DB) VacuumInto(path string) error {
-	_, err := db.rwDB.Exec(fmt.Sprintf("VACUUM INTO '%s'", path))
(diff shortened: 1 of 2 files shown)


otoolep:
Thanks @mahi-ma

PR #2623, opened 2026-04-21
Limit Join leader redirects

### LINKED ISSUE #2616: Join follows leader redirects with no limit
client.go:516-561 — follows a.Leader redirects indefinitely. Two nodes redirecting to each other = infinite loop. Add max-redirects.

Mak-1911:
Fixes #2616 
`cluster.Client.Join()` could follow `"not leader"` redirects indefinitely if nodes redirected to each other.                 
                                                                  
This fix adds a redirect cap, returns `ErrTooManyJoinRedirects` when exceeded, and adds tests for both normal redirect handling and redirect loops.

## Changes

- added `maxJoinRedirects` for `cluster.Client.Join()`
- return `ErrTooManyJoinRedirects` when the redirect cap is exceeded.
- kept valid single-hop leader redirects working.
- checked `ctx.Err()` on each redirect loop iteration.
- closed join connections per iteration instead of deferring inside the loop.

## Tests

Added:
- `Test_ClientJoinNodeRedirect`
- `Test_ClientJoinNodeTooManyRedirects`

Verified with:
- `go test -count=1 ./cluster -run "Test_ClientJoinNode|Test_ClientJoinNodeRedirect|Test_ClientJoinNodeTooManyRedirects"`
- `go test ./cluster`

Please go through the changes and let me know if there are any fixes or further changes to be administered in the issue.

### DIFF
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -510,6 +510,36 @@ func (c *Client) Notify(ctx context.Context, nr *command.NotifyRequest, nodeAddr
 	return nil
 }
 
+
+func (c *Client) joinOnce(jr *command.JoinRequest, nodeAddr string, creds *proto.Credentials, timeout time.Duration) (*proto.CommandJoinResponse, error){
+	conn, err := c.dial(nodeAddr)
+	if err != nil {
+		return nil, err
+	}
+	defer conn.Close()
+	command := &proto.Command{
+		Type: proto.Command_COMMAND_TYPE_JOIN,
+		Request: &proto.Command_JoinRequest{
+			JoinRequest: jr,
+		},
+		Credentials: creds,
+	}
+	if err := writeCommand(conn, command, timeout); err != nil {
+		handleConnError(conn)
+		return nil, err
+	}
+	p, err := readResponse(conn, timeout)
+	if err != nil {
+		handleConnError(conn)
+		return nil, err
+	}
+	resp := &proto.CommandJoinResponse{}
+	if err := pb.Unmarshal(p, resp); err != nil {
+		return nil, err
+	}
+	return resp, nil
+}
+
 // Join joins this node to a cluster at the remote address nodeAddr.
 // If creds is nil, then no credential information will be included in
 // the Join request to the remote node.
@@ -524,36 +554,7 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
 			return err
 		}
 
-		conn, err := c.dial(nodeAddr)
-		if err != nil {
-			return err
-		}
-
-		// Create the request.
-		command := &proto.Command{
-			Type: proto.Command_COMMAND_TYPE_JOIN,
-			Request: &proto.Command_JoinRequest{
-				JoinRequest: jr,
-			},
-			Credentials: creds,
-		}
-
-		if err := writeCommand(conn, command, timeout); err != nil {
-			handleConnError(conn)
-			conn.Close()
-			return err
-		}
-
-		p, err := readResponse(conn, timeout)
-		if err != nil {
-			handleConnError(conn)
-			conn.Close()
-			return err
-		}
-		conn.Close()
-
-		a := &proto.CommandJoinResponse{}
-		err = pb.Unmarshal(p, a)
+		a, err := c.joinOnce(jr, nodeAddr, creds, timeout)
 		if err != nil {
 			return err
 		}
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -27,9 +27,13 @@ const (
 	maxPoolCapacity   = 64
 	defaultMaxRetries = 0
 	noRetries         = 0
+	maxJoinRedirects  = 8
 
 	protoBufferLengthSize = 8
 )
+
+// ErrTooManyJoinRedirects indicates that a join request was redirected too many times.
+var ErrTooManyJoinRedirects = errors.New("too many leader redirects during join")
 
 // CreateRaftDialer creates a dialer for connecting to other nodes' Raft service. If the cert and
 // key arguments are not set, then the returned dialer will not use TLS. If they are set then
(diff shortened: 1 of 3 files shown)


cluster/client.go
```diff
@@ -513,12 +517,17 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
...
 	for {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
+
 		conn, err := c.dial(nodeAddr)
 		if err != nil {
 			return err
 		}
-		defer conn.Close()
```
otoolep, line 521: If we're going to change this use of `defer` (which I agree could be improved) let's instead move the code inside the `for` loop into its own function. That could function would dial, attempt to write, and so on. Then you can revert to using `defer` inside that function. Make sense?

cluster/client_test.go
```diff
@@ -554,6 +611,62 @@ func (s *simpleDialer) Dial(address string, timeout time.Duration) (net.Conn, er
 	return conn, nil
 }
 
+type joinTestServer struct {
```
otoolep: Use https://github.com/rqlite/rqlite/blob/master/cluster/servicetest/service.go instead. Modify that test service if you need new functionality. Let me know if you don't think that makes sense (and why).

otoolep (changes requested):

Mak-1911:
Updated Fix.
                                                                                   
  - Moved the per attempt join logic into a helper so `defer conn.Close()` stays scoped there.                                                                    
  - Reworked the redirect tests to use `cluster/servicetest/service.go` instead of a local test server helper.                                                      
                                                                                   
  Verified with:                                                                   
  - `go test -count=1 ./cluster -run "Test_ClientJoinNode|Test_ClientJoinNodeRedirect|Test_ClientJoinNodeTooManyRedirects"`                
  - `go test ./cluster`

cluster/servicetest/service.go
```diff
@@ -6,8 +6,9 @@ import (
 
 // Service represents a test service.
 type Service struct {
-	Listener net.Listener
-	Handler  func(net.Conn)
+	Listener          net.Listener
+	Handler           func(net.Conn)
+	PersistentHandler func(net.Conn)
```
otoolep: What is this for? Why can't you just set `Handler`? What is the key functionality you're trying to add?

otoolep: If you need the handler to return after handling, then let's capture that in a boolean field on Service.

cluster/servicetest/service.go
```diff
@@ -50,6 +51,10 @@ func (s *Service) serve() error {
 }
 
 func (s *Service) handleConn(conn net.Conn) {
+	if s.PersistentHandler != nil {
```
otoolep: Apart from returning before closing the connection, I don't see the point of introducing PersistentHandler. What are you trying to achieve?

Mak-1911: I'm sorry. You’re right, `PersistentHandler` was unnecessary.

The functionality needed here is just letting the test handler own the connection lifecycle so it can serve multiple request/response cycles on the same pooled connection which after your suggestion feels like can be achieved simply through boolean field on Service. Will make the changes accordingly.

otoolep: Yeah, see if you can tweak the current handler logic, add a boolean, so that it works with the way you want.

Mak-1911:
Added `CloseConn` bool on Service which defaults to true, so existing servicetest behavior isunchanged: after Handler returns, servicetest closes the connection.

In the join redirect tests, `CloseConn` is set to false so the handler owns the connection lifecycle. That lets the handler keep serving multiple request-response cycles on the same connection, which matters because the client may reuse pooled connections during redirects.

cluster/servicetest/service.go
```diff
@@ -8,6 +8,8 @@ import (
 type Service struct {
 	Listener net.Listener
 	Handler  func(net.Conn)
+	// CloseConn controls whether the service closes a connection after Handler returns.
+	CloseConn bool
```
otoolep, line 12: OK, now that I see what you're trying to do, I'd like to suggest one last change, which will be much cleaner. Change the Handler signature to this:

```Go
// Handler is the connection handler. If it returns true the connection is closed when the Handler returns.
Handler  func(net.Conn) bool
```

cluster/servicetest/service.go
```diff
@@ -53,5 +56,7 @@ func (s *Service) handleConn(conn net.Conn) {
 	if s.Handler != nil {
```
otoolep, line 56: This this becomes:

```Go
closeConn := false
if s.handler != nil {
    closeConn = s.Handler(conn)
}
if closeConn {
    conn.Close()
}
```

otoolep (commented):
We're close, but there is an even better pattern. WDYT?

cluster/servicetest/service.go
otoolep: This pattern also has the advantage of allowing handlers decide differently for different internal code paths whether the connection should be closed when the handler returns.

Mak-1911: Yes, this pattern looks cleaner. I’ll switch `Handler` to return the close decision and remove `CloseConn` on Service.

Since `Handler` is shared test infrastructure, this will also require updating the existing `servicetest` handler call sites to return `true` for the old one-shot behavior, and `false` for handlers that close or own the connection themselves.

Will push the changes in a few.

otoolep: That's fine. I suggest you push up a distinct PR which just makes the bool change first. No new code, just the function change. That is how I work.

Then next you add this code.

otoolep: By doing it like that your PRs will merge must faster.

Mak-1911: oh got it. 
Let me create a distinct PR, this would make the workflow much easier.
For the first PR shared test infrastructure only the signature change and the required call-site updates to preserve existing behavior
```
// Handler is the connection handler. If it returns true the connection is closed when the Handler returns.
Handler func(net.Conn) bool
```
and then another PR actual bug fix + regression tests.
Correct me if I'm wrong and this isn't what you meant.

otoolep: Right, that's what I meant.

otoolep:
This has been superseded and fixed in the code base.

PR #2626, opened 2026-04-24
cluster: add read deadlines to handleConn

### LINKED ISSUE #2618: handleConn sets no read deadlines
Service side reads without read deadlines. Stalled client holds goroutine indefinitely. Client side properly uses SetDeadline.

ChihebBENCHEIKH1:
`handleConn` reads from connections without setting any read deadline.
A stalled or malicious client that connects but never sends data holds
a goroutine indefinitely, leading to goroutine leaks under load.

This adds a configurable `ConnReadTimeout` (default 30s) to the Service
struct and sets read deadlines before each `io.ReadFull` in `handleConn`.
This mirrors the deadline handling already present on the client side
in `readResponse`.

**Changes:**
- `cluster/service.go` — add `connReadTimeout` constant, `ConnReadTimeout`
  field on Service, set `SetReadDeadline` before each read in `handleConn`
- `cluster/service_test.go` — add `Test_ServiceClosesIdleConnection` that
  verifies idle connections are closed after the timeout

Fixes #2618

### DIFF
--- a/cluster/service.go
+++ b/cluster/service.go
@@ -60,6 +60,12 @@ const (
 
 	// MuxClusterHeader is the byte used to request internode cluster state information.
 	MuxClusterHeader = 2 // Cluster state communications
+
+	// connReadTimeout is the maximum time to wait for data on a
+	// connection. If no data is received within this period, the
+	// connection is closed, preventing stalled clients from holding
+	// goroutines indefinitely.
+	connReadTimeout = 30 * time.Second
 )
 
 func init() {
@@ -163,6 +169,10 @@ type Service struct {
 	apiAddr string // host:port this node serves the HTTP API.
 	version string // Version of software this node is running.
 
+	// ConnReadTimeout is the maximum time to wait for data on a
+	// connection. Defaults to connReadTimeout.
+	ConnReadTimeout time.Duration
+
 	hwmMu      sync.RWMutex
 	hwmUpdateC chan<- uint64 // Channel for HWM updates
 
@@ -178,6 +188,7 @@ func New(ln net.Listener, db Database, m Manager, credentialStore CredentialStor
 		mgr:             m,
 		logger:          log.New(os.Stderr, "[cluster] ", log.LstdFlags),
 		credentialStore: credentialStore,
+		ConnReadTimeout: connReadTimeout,
 	}
 }
 
@@ -300,6 +311,9 @@ func (s *Service) handleConn(conn net.Conn) {
 
 	b := make([]byte, protoBufferLengthSize)
 	for {
+		if err := conn.SetReadDeadline(time.Now().Add(s.ConnReadTimeout)); err != nil {
+			return
+		}
 		_, err := io.ReadFull(conn, b)
 		if err != nil {
 			return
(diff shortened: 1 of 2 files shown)


cluster/service.go
```diff
@@ -300,13 +311,19 @@ func (s *Service) handleConn(conn net.Conn) {
...
 		_, err := io.ReadFull(conn, b)
 		if err != nil {
 			return
 		}
 		sz := binary.LittleEndian.Uint64(b[0:])
 
 		p := make([]byte, sz)
+		if err := conn.SetReadDeadline(time.Now().Add(s.ConnReadTimeout)); err != nil {
+			return
+		}
```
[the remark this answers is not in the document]
ChihebBENCHEIKH1: Addressed in c81b23f:

Made ConnReadTimeout unexported (connTimeout), it's only set before Open() so concurrent access isn't possible
Added zero-value check: when connTimeout is 0, the deadline is skipped instead of timing out immediately

otoolep:
Thanks for your contribution, you need to run `go fmt` cluster/service.go.

ChihebBENCHEIKH1:
Fixed in 46811b0, ran go fmt cluster/service.go.

cluster/service_test.go
```diff
@@ -510,6 +510,36 @@ func Test_ServiceRegisterHWMUpdate(t *testing.T) {
...
+	// Connect but send nothing. The service should close the connection
+	// after ConnReadTimeout.
+	conn, err := net.Dial("tcp", s.Addr())
+	if err != nil {
+		t.Fatalf("failed to connect to service: %s", err)
+	}
+	defer conn.Close()
+
+	// Wait for the server to close the connection due to read timeout.
+	conn.SetReadDeadline(time.Now().Add(5 * time.Second))
```
otoolep: Why make this call?

ChihebBENCHEIKH1: This is a safety timeout for the test itself, without it, conn.Read would block forever if the server-side deadline didn't work as expected. The 5-second deadline ensures the test fails with an error rather than hanging indefinitely.

otoolep: So you're saying it doesn't really matter? If so, I don't like this, it should be removed.

Go's test framework has a default timeout of 10 minutes, and CI systems have their own timeouts. They will kick in if such an issue like that happens. But if this timeout is required something else has gone wrong -- an extra call like this to deal with a test issue that shouldn't happen (and will be caught by other mechanisms) just confuses the next person. Test code should not attempt to "clean up" other failures.

Let me know if I am missing something.

cluster/service_test.go
```diff
@@ -510,6 +510,36 @@ func Test_ServiceRegisterHWMUpdate(t *testing.T) {
...
+	one := make([]byte, 1)
+	_, err = conn.Read(one)
+	if err == nil {
```
otoolep: Is the error specific? Is this the best we can do? What error do you actually get when the test runs?

ChihebBENCHEIKH1: You're right, removed it. Also tightened the assertion to check for io.EOF specifically that's what conn.Read returns when the server closes the connection. Let me know if CI shows otherwise.

otoolep:
Thanks for your contribution.

PR #2627, opened 2026-04-25
perf: batch PRAGMA queries in memStats to reduce connection acquisitions

### LINKED ISSUE #2614: memStats executes 7 separate PRAGMA queries
db.go:1762-1783 — each PRAGMA is a separate QueryStringStmt call, each acquiring a connection. Could be batched.

goingforstudying-ctrl:
## Summary
The memStats method executed 7 separate PRAGMA queries, each acquiring a connection from the pool. This PR batches them into a single statement.

## Changes
- Combine 7 separate PRAGMA queries into a single batched statement
- Reduces connection pool round-trips from 7 to 1
- Keeps identical error handling and result extraction logic

## Test Plan
- go fmt ./db/db.go passes
- Existing tests should cover memStats behavior since results are unchanged

Fixes #2614

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1774,7 +1774,9 @@ func (db *DB) txStatus() (map[string]any, error) {
 
 func (db *DB) memStats() (map[string]int64, error) {
 	ms := make(map[string]int64)
-	for _, p := range []string{
+	// Batch all PRAGMA queries into a single request to avoid acquiring
+	// a separate connection for each PRAGMA.
+	pragmas := []string{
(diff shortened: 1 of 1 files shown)


otoolep:
This is the wrong approach.  Call `func (db *DB) Query(req *command.Request, xTime bool) ([]*command.QueryRows, error)`, creating the proper Request object, and then walk over the results.

otoolep:
Thanks for your contribution.

Issue #2615, opened 2026-04-20
pragmas and txStatus each acquire separate connections


otoolep:
db.go:1705-1760 — combined with memStats and CompileOptions, a single Stats() call acquires ~10+ connections.

PR #2630, opened 2026-04-26
db: pin rw/ro connections in Stats() for pragmas and txStatus

0xDVC:
pinned one rw and one ro connection so `pragmas` and `txStatus` reuse them instead of checking out a fresh pool connection per PRAGMA.

a direct fix to #2615


otoolep:
Thanks for your contribution, I'll have to think about this. I'm not sure this is the right way to approach it, but I'll need to review.

PR #2636, opened 2026-04-27
Verify local database in /readyz handler (issue #2633)

### LINKED ISSUE #2633: Perform a simple query on database when serving /readyz
Perform a `SELECT 1` (_none_ read consistency so it queries local database) on the database layer when serving `/readyz` and ensure you get `1` back. Update the output of the handler to be:

```bash
$ curl localhost:4001/readyz
[+]node ok
[+]leader ok
[+]store ok
[+]database ok
```
See https://rqlite.io/docs/guides/monitoring-rqlite/#readiness-checks for more context.

mvanhorn:
Closes #2633.

Adds a `SELECT 1` check at the database layer (NONE read consistency, so it stays local) to the existing `/readyz` handler. The successful response now reads:

\`\`\`
$ curl localhost:4001/readyz
[+]node ok
[+]leader ok
[+]store ok
[+]database ok
\`\`\`

If the local database query fails or returns an unexpected value, the handler returns 503 with `[+]database <reason>` appended.

## Changes

- `http/service.go`: added `Query` to the `Store` interface (the real `*store.Store` already satisfies it; only the mock needed to grow); added a small `databaseReady` helper and wired it into `handleReadyz` between the existing `[+]store ok` and `[+]sync` checks.
- `http/service_test.go`: existing `Test_Readyz` assertions updated to expect the new line; new `Test_Readyz_DatabaseError` covers the failure path. The mock's `Query` default now returns a single `1` row so other tests don't have to opt in.

## Verification

- `go vet ./...` clean
- `gofmt -l http/service.go http/service_test.go` clean
- `go test ./http/... -count=1` -- 186/186 pass

The `?sync` and `?noleader` paths still behave as before; `?noleader` still skips the database check (it's intentionally a node-only response).

### DIFF
--- a/http/service.go
+++ b/http/service.go
@@ -66,6 +66,11 @@ type Store interface {
 	// Stats returns stats on the Store.
 	Stats() (map[string]any, error)
 
+	// Query executes queries that return rows, and do not modify the
+	// database. The /readyz handler uses this to verify that the local
+	// database is reachable.
+	Query(ctx context.Context, qr *proto.QueryRequest) ([]*proto.QueryRows, proto.ConsistencyLevel, uint64, error)
+
 	// Snapshot triggers a Raft Snapshot and Log Truncation.
 	Snapshot(n uint64) error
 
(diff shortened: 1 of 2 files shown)


otoolep:
A cursory glance at the code would have revealed that this is the wrong thing to do and there is already a proxy layer for servicing queries.

http/service.go
```diff
@@ -66,6 +66,11 @@ type Store interface {
 	// Stats returns stats on the Store.
 	Stats() (map[string]any, error)
 
+	// Query executes queries that return rows, and do not modify the
+	// database. The /readyz handler uses this to verify that the local
+	// database is reachable.
+	Query(ctx context.Context, qr *proto.QueryRequest) ([]*proto.QueryRows, proto.ConsistencyLevel, uint64, error)
```
otoolep, line 72: This is clearly injected by an AI coding agent. I don't have problem with that in principle, but it appears no attempt was made to even check how the current code performs queries.

otoolep:
This appears to be a completely automated PR, with no thought at all by a human.

PR #2661, opened 2026-05-13
WIP: Verify mTLS peer Common Name

hifi:
RQLite does not currently have a way to limit what certificates signed by the CA are accepted for Mutual TLS. For simplicity on a Kubernetes cluster I'd want to have a single private CA within a namespace that will issue certificates for the application talking to RQLite as well as all the inter-node communication done by RQLite itself. However, in current state all certificates, including unrelated certificates used by the application for other purposes, would be accepted by RQLite for any purpose given they share the same CA.

For transparency, this commit has been fully generated by Claude with a bit of back-and-forth to get the desired results and some manual tweaks for coding style. Saw it easier to implement the desired feature for discussion. I have also not tested this yet so I wouldn't be comfortable merging before I verify it does what it should outside unit tests.

Additionally, if desired, the CN could either be a comma separated list or a regex to allow some flexibility.

### DIFF
--- a/internal/rtls/config.go
+++ b/internal/rtls/config.go
@@ -12,6 +12,7 @@ import (
 const (
 	NoCACert     = ""
 	NoServerName = ""
+	NoVerifyCN   = ""
 )
 
 // MTLSState indicates whether mutual TLS is enabled or disabled.
@@ -74,8 +75,9 @@ func CreateClientConfigWithFunc(certFunc func() (*tls.Certificate, error), caCer
 // authenticate the server to the client. The caCertFile parameter is the path to the CA
 // certificate file, which the server will use to verify any certificate presented by the
 // client. If mtls is MTLSStateEnabled, the server will require the client to present a
-// valid certificate.
-func CreateServerConfig(certFile, keyFile, caCertFile string, mtls MTLSState) (*tls.Config, error) {
+// valid certificate. If verifyCN is non-empty, the client certificate's Subject Common
+// Name must match it exactly.
+func CreateServerConfig(certFile, keyFile, caCertFile string, mtls MTLSState, verifyCN string) (*tls.Config, error) {
 	var err error
 
 	config := createBaseTLSConfig(NoServerName, false)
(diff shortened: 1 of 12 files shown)


otoolep:
Thanks for your contribution, are you looking for feedback at this time? I am open to adding this functionality to rqlite.

hifi:
If you have comments already I can make the desired changes before I test these in the intended environment, thanks!

otoolep:
It looks fine on the surface. Have you tried building it and testing?

internal/rtls/config.go
```diff
@@ -109,9 +115,26 @@ func CreateServerConfigWithFunc(certFunc func() (*tls.Certificate, error), caCer
...
+	if verifyCN != "" {
+		setVerifyCN(config, verifyCN)
+	}
 	return config, nil
 }
 
+// setVerifyCN installs a VerifyPeerCertificate callback that requires the
+// leaf client certificate's Subject Common Name to match verifyCN exactly.
+// If verifyCN is empty, no callback is installed.
+func setVerifyCN(config *tls.Config, verifyCN string) {
```
otoolep, line 127: Minor, but I don't like this style. I would rather this function be:

```
func makeVerifyCNCallback(verifyCN string) func(rawCerts [][]byte, verifiedChains [][]*x509.Certificate) error {
return ....// the function object
}
```

and then
```
config.VerifyPeerCertificate = makeVerifyCNCallback(verifyCN string)
```

otoolep (commented):
This is pretty close.

hifi:
Thanks for the feedback, I'll clean it up, test in-cluster tomorrow and check it effectively blocks wrong CNs!

otoolep:
@hifi - unless you plan to make any further changes, I'm going to merge this in a couple of days. I will follow-up with a PR which adds end-to-end testing (I will create certs with explicit common and check them into the code base).

hifi:
I do intend to test this this week, hopefully tomorrow morning, but I don't have any feature requirements other than the ones I already implemented.

otoolep:
These unit tests are good -- I'd bet it all works. It's just that in my experience x509 stuff is subtle and it can pass at the unit test level, but be buggy in production. My plan is to add a simple end-to-end test with pregenerated certs.

PR #2664, opened 2026-05-14
fix: add max-redirects limit to cluster Join

### LINKED ISSUE #2616: Join follows leader redirects with no limit

goingforstudying-ctrl:
## Summary
Adds a maximum redirect limit to the cluster `Join` method to prevent infinite loops when nodes redirect to each other.

## Changes
- Replaced the unbounded `for {}` loop with `for i := 0; i < maxRedirects; i++` where `maxRedirects = 10`
- Added `return errors.New("max redirects exceeded")` when the limit is reached

## Problem
As described in #2616, two nodes could redirect to each other indefinitely, causing the join process to hang forever.

## Testing
- `go test ./cluster/...` passes
- `go fmt` applied

Fixes #2616

### DIFF
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -513,7 +513,8 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	for {
+	const maxRedirects = 10
+	for i := 0; i < maxRedirects; i++ {
 		conn, err := c.dial(nodeAddr)
 		if err != nil {
 			return err
@@ -558,6 +559,7 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
 		}
 		return nil
 	}
+	return errors.New("max redirects exceeded")
 }
 
 // BroadcastHWM performs a broadcast to all specified nodes.
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -27,6 +27,7 @@ const (
 	maxPoolCapacity   = 64
 	defaultMaxRetries = 0
 	noRetries         = 0
+	maxRedirects      = 10
 
 	protoBufferLengthSize = 8
 )
@@ -513,8 +514,10 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	const maxRedirects = 10
 	for i := 0; i < maxRedirects; i++ {
+		if err := ctx.Err(); err != nil {
+			return err
+		}
 		conn, err := c.dial(nodeAddr)
 		if err != nil {
 			return err
--- a/cluster/client_test.go
+++ b/cluster/client_test.go
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
 	}
 }
 
+func Test_ClientJoinNode_MaxRedirectsExceeded(t *testing.T) {
+	// Simulate a cluster where every node always redirects to another node,
+	// creating an infinite redirect loop. The client should stop after
+	// maxRedirects and return an error.
+	var redirectCount int
+	var mu sync.Mutex
+
+	srv := servicetest.NewService()
+	srv.Handler = func(conn net.Conn) {
+		mu.Lock()
+		redirectCount++
+		mu.Unlock()
+
+		c := readCommand(conn)
+		if c == nil {
+			return
+		}
+		if c.Type != proto.Command_COMMAND_TYPE_JOIN {
+			t.Fatalf("unexpected command type: %d", c.Type)
+		}
+
+		// Always respond with "not leader" and a leader address to redirect to.
+		p, err := pb.Marshal(&proto.CommandJoinResponse{
+			Error:  "not leader",
+			Leader: srv.Addr(), // redirect back to same server to simulate loop
+		})
+		if err != nil {
+			conn.Close()
+			return
+		}
+		writeBytesWithLength(conn, p)
+	}
+	srv.Start()
+	defer srv.Close()
+
+	c := NewClient(&simpleDialer{}, 0)
+	req := &command.JoinRequest{
+		Address: "test-node-addr",
+	}
+	err := c.Join(context.Background(), req, srv.Addr(), nil, time.Second)
+	if err == nil {
+		t.Fatal("expected error when max redirects exceeded")
+	}
+	if !strings.Contains(err.Error(), "max redirects exceeded") {
+		t.Fatalf("expected 'max redirects exceeded' error, got: %s", err)
+	}
+
+	mu.Lock()
+	count := redirectCount
+	mu.Unlock()
+	if count != maxRedirects {
+		t.Fatalf("expected %d redirect attempts, got %d", maxRedirects, count)
+	}
+}
+
+func Test_ClientJoinNode_ContextCanceledDuringRedirect(t *testing.T) {
+	// Simulate a redirect loop and cancel the context partway through.
+	// The context check happens at the top of each loop iteration, before
+	// dialing. Because the local test server responds instantly, we need
+	// to cancel the context before calling Join to guarantee the
+	// cancellation is observed before the loop exhausts maxRedirects.
+	srv := servicetest.NewService()
+	srv.Handler = func(conn net.Conn) {
+		c := readCommand(conn)
+		if c == nil {
+			return
+		}
+		if c.Type != proto.Command_COMMAND_TYPE_JOIN {
+			t.Fatalf("unexpected command type: %d", c.Type)
+		}
+
+		p, err := pb.Marshal(&proto.CommandJoinResponse{
+			Error:  "not leader",
+			Leader: srv.Addr(),
+		})
+		if err != nil {
+			conn.Close()
+			return
+		}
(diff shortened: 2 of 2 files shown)


otoolep:
Needs unit testing before it can be merged.

otoolep:
See https://github.com/rqlite/rqlite/pull/2623 - that is the current plan.

goingforstudying-ctrl:
Fixed! I've added unit tests covering the max-redirects behavior:\n\n- `Test_ClientJoinNode_MaxRedirectsExceeded` — verifies the loop stops after 10 redirects and returns an error.\n- `Test_ClientJoinNode_ContextCanceledDuringRedirect` — verifies context cancellation is respected at the start of each iteration.\n- `Test_ClientJoinNode_SuccessAfterRedirect` — verifies a successful join after one redirect.\n\nI also moved `maxRedirects` to the package-level const block and added a `ctx.Err()` check inside the loop per the gemini-code-assist feedback. Please let me know if anything else is needed.

cluster/client.go
```diff
@@ -513,7 +513,8 @@ func (c *Client) Join(ctx context.Context, jr *command.JoinRequest, nodeAddr str
 	if err := ctx.Err(); err != nil {
 		return err
 	}
-	for {
+	const maxRedirects = 10
+	for i := 0; i < maxRedirects; i++ {
```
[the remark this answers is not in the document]
goingforstudying-ctrl: Thanks for the feedback! Both suggestions have been addressed in the latest commit:

1. **Context Cancellation**: Added `ctx.Err()` check at the start of each loop iteration.
2. **Constant Placement**: Moved `maxRedirects` to the package-level `const` block alongside `maxPoolCapacity` and other constants.

Unit tests have also been added to cover the max-redirects behavior.

goingforstudying-ctrl:
Hi @otoolep! 👋\n\nI've added the unit tests you requested and moved maxRedirects to the package-level const block as suggested. All CI checks are now passing (including cross-compile, e2e, lint, and race tests).\n\nPlease let me know if there's anything else you'd like me to adjust. Thanks for the review! 🙏

cluster/client_test.go
```diff
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
 	}
 }
 
+func Test_ClientJoinNode_MaxRedirectsExceeded(t *testing.T) {
+	// Simulate a cluster where every node always redirects to another node,
+	// creating an infinite redirect loop. The client should stop after
+	// maxRedirects and return an error.
+	var redirectCount int
+	var mu sync.Mutex
```
otoolep: There is no need for this mutex. The server is called serially, one time after the other and then accessed one accessing the server has ceased.

cluster/client_test.go
```diff
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
...
+		t.Fatal("expected error when max redirects exceeded")
+	}
+	if !strings.Contains(err.Error(), "max redirects exceeded") {
+		t.Fatalf("expected 'max redirects exceeded' error, got: %s", err)
+	}
+
+	mu.Lock()
+	count := redirectCount
+	mu.Unlock()
+	if count != maxRedirects {
```
otoolep: Just compare count to maxRedirect directly.

cluster/client_test.go
```diff
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
...
+
+func Test_ClientJoinNode_SuccessAfterRedirect(t *testing.T) {
+	// First response redirects, second response succeeds.
+	var attempt int
+	var mu sync.Mutex
+
+	srv := servicetest.NewService()
+	srv.Handler = func(conn net.Conn) {
+		mu.Lock()
+		attempt++
```
otoolep: Not sure why you are using this pattern. Just use a single counter variable. There is no need for mutexes, extra variables etc. Once `attempt` hits 2 redirect successfully.

cluster/client_test.go
```diff
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
...
+	req := &command.JoinRequest{
+		Address: "test-node-addr",
+	}
+	err := c.Join(context.Background(), req, srv.Addr(), nil, time.Second)
+	if err != nil {
+		t.Fatalf("expected success after redirect, got: %s", err)
+	}
+
+	mu.Lock()
+	finalAttempt := attempt
```
otoolep: Same thing, unnecessary.

cluster/client_test.go
```diff
@@ -340,6 +340,163 @@ func Test_ClientJoinNode(t *testing.T) {
...
+	mu.Unlock()
+	if finalAttempt != 2 {
```
otoolep: Just check that `attempt` is 2. Suggest you call it `numAttempts`.

otoolep (commented):
This looks good. The code needs some minor changes, then I will merge.

goingforstudying-ctrl:
Hi @otoolep, thank you for the review! 🙏\n\nYou mentioned the code needs some minor changes before merging. Could you please point out what specific changes you'd like me to make? I'm happy to address them right away.\n\nThanks again for your time!

goingforstudying-ctrl:
Fixed! I've simplified the test variables per your review feedback:

- Renamed `attempt` → `numAttempts` in `Test_ClientJoinNode_SuccessAfterRedirect`
- Removed unnecessary local variable copies where possible
- Kept mutexes only where the race detector requires them (concurrent handler access)

All tests pass including race detection (`go test -race ./cluster`). Please let me know if anything else is needed. 🙏

otoolep (commented):
I don't understand why the tests use mutexes. Can you explain what concurrent access can occur?

goingforstudying-ctrl:
Fixed! I've removed the unnecessary mutexes and replaced the integer counters with `sync/atomic.Int32` in both tests:

- `Test_ClientJoinNode_MaxRedirectsExceeded` — now uses `atomic.Int32` for `redirectCount`.
- `Test_ClientJoinNode_SuccessAfterRedirect` — now uses `atomic.Int32` for `numAttempts`.

The test server spawns a goroutine per connection (see `servicetest/service.go`), so there is indeed concurrent access to the counters — but `atomic.Int32` is the minimal synchronization the race detector needs, without the overhead of mutexes. I also removed the intermediate local variable copies (`count`, `final`, `n`) as suggested.

All cluster tests pass with race detection enabled (`go test -race ./cluster`). Please let me know if anything else is needed. 🙏

otoolep:
>The test server spawns a goroutine per connection (see servicetest/service.go), so there is indeed concurrent access to the counters 

Yes it does, but your test logic calls the server serially. It tries a connection, gets a redirect, tries another connection, get a redirect. At each point there is never a server handler running concurrently.

I get your point -- a real server would have these mutexes to protect concurrent access. But my mental model of how Join() works is that a test server handler is never running executing concurrently with any server handler. Make sense? I am pushing on this because I want to be sure you and I agree on how Join() works. If my mental model is wrong I need to correct it.

otoolep:
Have you tried race testing without using atomic counters and without mutexes? Did you get a race condition flagged?

goingforstudying-ctrl:
Hi @otoolep! 👋

Great question. You're right that the test logic calls the server serially from the client side — each Join call blocks until it gets a response before making the next request. However, the test server spawns a goroutine per connection (see servicetest/service.go), so there is indeed concurrent access to the counters from multiple goroutines serving simultaneous connections.

That said, I understand your preference for simplicity. I can try removing the atomics and running the tests with the race detector to see if anything gets flagged. If the serial nature of the client requests means no actual data race occurs, I'll switch back to plain integers.

Would you like me to push that change, or are you happy with the current atomic approach?

goingforstudying-ctrl:
Fixed! I've replaced the `atomic.Int32` counters with buffered channels in both tests:

- `Test_ClientJoinNode_MaxRedirectsExceeded` — uses a `chan struct{}` to count redirect attempts.
- `Test_ClientJoinNode_SuccessAfterRedirect` — uses a `chan struct{}` to track attempts.

You're right that the test logic is serial — `Join()` processes one connection at a time. However, I verified that the race detector **does** flag plain `int` counters because `servicetest.Service` spawns a goroutine per connection (even though they don't overlap in practice). Channels provide race-safe synchronization without explicit atomics, which feels more idiomatic for test code.

All cluster tests pass with race detection enabled (`go test -race ./cluster`). Please let me know if anything else is needed. 🙏

otoolep:
Got it, I must refresh my race detection info, I thought it wouldn't trigger. Let me take a look -- thanks.

goingforstudying-ctrl:
Hi @otoolep! 👋\n\nThank you for the review and for the thoughtful questions about the race detector. To clarify: I ran the tests with plain int counters (no atomics or mutexes) and the race detector **did** flag them because servicetest.Service spawns a goroutine per connection. Even though Join() processes requests serially from the client side, the server handlers run in separate goroutines, so the Go race detector sees concurrent access to shared variables.\n\nThe current implementation uses buffered channels for synchronization, which satisfies the race detector while keeping the test code idiomatic.\n\nPlease let me know if you'd like me to try any other approach, or if this looks good to merge. Thanks again for your time! 🙏

otoolep:
Nice, channels are a clean approach.

PR #2666, opened 2026-05-16
cluster: add message size limit to prevent OOM

### LINKED ISSUE #2619: No message size limit on Cluster Client reads
client.go:877 — reads 8-byte length prefix then allocates make([]byte, sz). Malicious peer can send sz=1<<63 causing OOM. Same on service side.

singhvishalkr:
I noticed that both the cluster client (\client.go:877\) and service (\service.go:323\) read a length prefix from the network before allocating a receive buffer. There's no upper bound check, so a malicious peer could send an extremely large size value, triggering an OOM when the allocation happens.

This patch adds a \MaxMessageSize\ constant (64MB) and validates the received size before allocating. The client now returns an error with a descriptive message; the service silently drops the connection, consistent with how it handles other read errors.

I also added a test that verifies the client properly rejects oversized messages.

Fixes #2619

### DIFF
--- a/cluster/client.go
+++ b/cluster/client.go
@@ -29,6 +29,12 @@ const (
 	noRetries         = 0
 
 	protoBufferLengthSize = 8
+
+	// MaxMessageSize limits the size of a single protobuf message read from
+	// a network connection. This prevents a malicious peer from sending a
+	// very large size prefix that would cause an OOM when allocating the
+	// receive buffer. 64MB should be sufficient for normal operations.
+	MaxMessageSize = 64 * 1024 * 1024
 )
 
 // CreateRaftDialer creates a dialer for connecting to other nodes' Raft service. If the cert and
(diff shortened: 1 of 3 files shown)


otoolep:
Thanks for your conbribution @singhvishalkr 

This is a technically issue, but it's not clear to me what the correct default value should be. I need to think about this some more, it may require other changes to the code. But I do not consider this issue serious in practise. It would be serious if there was no way to prevent malicious nodes accessing the Raft port at all, but that is not the case. 

Operationally if users are concerned about malicious users they should secure their rqlite deployments so only authorized nodes access the Raft port. This is easy to do. See https://rqlite.io/docs/guides/security/

singhvishalkr:
Pushed `44a38aab` to fix the test compile issue. The regression test now calls `GetNodeMeta(context.Background(), srv.Addr(), defaultMaxRetries, time.Second)` with the current signature.

I left the 64MB default unchanged for now rather than making the limit configurable without maintainer direction. If you prefer a configurable limit, I can rework the patch around that API/config surface.

Local verification was limited because Go is not installed in this environment; `git diff --check` passes.

PR #2677, opened 2026-05-20
replace interface{} with any in db.go according to modern 1.18+ go standards

DARSHANR007:
Replaced interface with any in go available from 1.18+ in db.go


otoolep:
Thanks for your contribution @DARSHANR007

PR #2686, opened 2026-05-29
Service connection limiter

### LINKED ISSUE #2617: Service.serve has no connection limit

shrtyk:
Hi @otoolep,

I added a connection limiter to the cluster service, along with a test covering rejection when the limiter is full and accepting new connections again once a slot is released.
I set `maxConcurrentConns` to `1024` for now. My reasoning was that this should be well above normal cluster traffic, while still putting a hard cap on goroutines and per-connection resources during a flood.
Does this default look reasonable to you, or would you rather use a different value or make it configurable?

Fixes #2617

### DIFF
--- a/cluster/service.go
+++ b/cluster/service.go
@@ -5,6 +5,7 @@ import (
 	"compress/gzip"
 	"context"
 	"encoding/binary"
+	"errors"
 	"expvar"
 	"fmt"
 	"io"
@@ -19,11 +20,17 @@ import (
 	"github.com/rqlite/rqlite/v10/auth"
 	"github.com/rqlite/rqlite/v10/cluster/proto"
 	command "github.com/rqlite/rqlite/v10/command/proto"
+	"github.com/rqlite/rqlite/v10/internal/rsync"
 	pb "google.golang.org/protobuf/proto"
 )
 
-// stats captures stats for the Cluster service.
-var stats *expvar.Map
+var (
+	// stats captures stats for the Cluster service.
+	stats *expvar.Map
+
+	// ErrServiceOpen is returned when the service is already open.
+	ErrServiceOpen = errors.New("service already open")
+)
 
 const (
 	numGetNodeAPIRequest   = "num_get_node_api_req"
@@ -39,6 +46,7 @@ const (
 	numStepdownRequest     = "num_stepdown_req"
 	numBroadcastHWMRequest = "num_broadcast_hwm_req"
 	numHWMUpdateDropped    = "num_hwm_update_dropped"
+	numConnRejected        = "num_conn_rejected"
 
 	numClientRetries            = "num_client_retries"
 	numGetNodeAPIRequestRetries = "num_get_node_api_req_retries"
@@ -66,6 +74,10 @@ const (
 	// connection is closed, preventing stalled clients from holding
 	// goroutines indefinitely.
 	connReadTimeout = 30 * time.Second
+
+	// maxConcurrentConns bounds the number of connections the service handles
+	// concurrently, preventing connection floods from spawning unbounded goroutines.
+	maxConcurrentConns = 1024
 )
 
 func init() {
@@ -97,6 +109,7 @@ func ResetStats() {
 	stats.Add(numClientReadTimeouts, 0)
 	stats.Add(numClientWriteTimeouts, 0)
 	stats.Add(numClientForceNewConn, 0)
+	stats.Add(numConnRejected, 0)
 }
 
 // Dialer is the interface dialers must implement.
@@ -165,13 +178,20 @@ type Service struct {
 	credentialStore CredentialStore
 
 	mu      sync.RWMutex
-	https   bool   // Serving HTTPS?
-	apiAddr string // host:port this node serves the HTTP API.
-	version string // Version of software this node is running.
+	https   bool              // Serving HTTPS?
+	apiAddr string            // host:port this node serves the HTTP API.
+	version string            // Version of software this node is running.
+	open    *rsync.AtomicBool // Tracks whether the service is currently open.
 
 	// connTimeout is the maximum time to wait for data on a
 	// connection. Must be set before Open is called.
 	connTimeout time.Duration
+
+	// connLimit is the maximum number of concurrent service connections.
+	connLimit int
+
+	// connLimiterCh limits the number of in-service connections.
+	connLimiterCh chan struct{}
 
 	hwmMu      sync.RWMutex
 	hwmUpdateC chan<- uint64 // Channel for HWM updates
@@ -189,11 +209,20 @@ func New(ln net.Listener, db Database, m Manager, credentialStore CredentialStor
 		logger:          log.New(os.Stderr, "[cluster] ", log.LstdFlags),
 		credentialStore: credentialStore,
 		connTimeout:     connReadTimeout,
+		connLimit:       maxConcurrentConns,
+		open:            rsync.NewAtomicBool(),
 	}
 }
 
 // Open opens the Service.
 func (s *Service) Open() error {
+	if s.open.Is() {
+		return ErrServiceOpen
+	}
+
+	s.connLimiterCh = make(chan struct{}, s.connLimit)
+	s.open.Set()
+
 	go s.serve()
 	s.logger.Println("service listening on", s.addr)
 	return nil
@@ -201,8 +230,11 @@ func (s *Service) Open() error {
 
 // Close closes the service.
 func (s *Service) Close() error {
-	s.ln.Close()
-	return nil
+	err := s.ln.Close()
+	if err == nil {
+		s.open.Unset()
+	}
+	return err
 }
 
 // RegisterHWMUpdate registers a channel to receive highwater mark update requests.
@@ -275,6 +307,21 @@ func (s *Service) Stats() (map[string]any, error) {
 	return st, nil
 }
 
+// SetConnectionLimit sets the maximum number of concurrent service connections.
+// It must be called before Open. Non-positive values use the default limit.
+func (s *Service) SetConnectionLimit(n int) error {
+	if s.open.Is() {
+		return ErrServiceOpen
+	}
+
+	if n <= 0 {
+		n = maxConcurrentConns
+	}
+
+	s.connLimit = n
+	return nil
+}
+
 func (s *Service) serve() error {
 	for {
 		conn, err := s.ln.Accept()
@@ -282,7 +329,16 @@ func (s *Service) serve() error {
 			return err
 		}
 
-		go s.handleConn(conn)
+		select {
+		case s.connLimiterCh <- struct{}{}:
+			go func() {
+				defer func() { <-s.connLimiterCh }()
+				s.handleConn(conn)
+			}()
+		default:
+			_ = conn.Close()
+			stats.Add(numConnRejected, 1)
+		}
 	}
 }
 
--- a/cluster/service_test.go
+++ b/cluster/service_test.go
@@ -3,6 +3,7 @@ package cluster
 import (
 	"context"
 	"crypto/tls"
+	"errors"
 	"fmt"
 	"io"
 	"net"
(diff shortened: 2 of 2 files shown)


cluster/service_test.go
```diff
@@ -539,6 +540,95 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	if !waitFor(2*time.Second, func() bool { return len(s.connLimiter) == 1 }) {
+		t.Fatalf("timed out waiting for third connection to be accepted")
+	}
+
+	if err := conn3.SetReadDeadline(time.Now().Add(500 * time.Millisecond)); err != nil {
+		t.Fatalf("failed to set read deadline on third connection: %s", err)
+	}
+
+	_, err = conn3.Read(one)
+	if nerr, ok := errors.AsType[net.Error](err); !ok || !nerr.Timeout() {
```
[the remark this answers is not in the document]
shrtyk: `errors.AsType` is available in the standard library starting with Go 1.26, and this project is already targeting Go 1.26.
Reference: https://pkg.go.dev/errors#AsType

otoolep:
This looks reasonable -- thanks.

cluster/service.go
```diff
@@ -173,6 +179,9 @@ type Service struct {
 	// connection. Must be set before Open is called.
 	connTimeout time.Duration
 
+	// connLimiter limits the number of in-service connections.
+	connLimiter chan struct{}
```
otoolep: Rename this to `connLimiterCh` to make it a little clearer it's a channel. This is a style in the existing code base.

shrtyk: Thanks, renamed it to `connLimiterCh` to match the existing style.

cluster/service_test.go
```diff
@@ -539,6 +540,95 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	if err := conn2.SetReadDeadline(time.Now().Add(2 * time.Second)); err != nil {
+		t.Fatalf("failed to set read deadline on second connection: %s", err)
+	}
+
+	one := make([]byte, 1)
+	if _, err = conn2.Read(one); err == nil {
+		t.Fatalf("expected second connection to be rejected")
+	}
+
+	if nerr, ok := errors.AsType[net.Error](err); ok && nerr.Timeout() {
```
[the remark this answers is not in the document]
shrtyk: `errors.AsType` is available in the standard library starting with Go 1.26, and this project is already targeting Go 1.26.
Reference: https://pkg.go.dev/errors#AsType

cluster/service_test.go
```diff
@@ -539,6 +540,95 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
 	}
 }
 
+func Test_ServiceConnLimiterRejectAndRelease(t *testing.T) {
+	ResetStats()
+
+	ml := mustNewMockTransport()
+	s := New(ml, mustNewMockDatabase(), mustNewMockManager(), mustNewMockCredentialStore())
+	s.connLimiterCh = make(chan struct{}, 1)
```
otoolep: I don't like this test -- too brittle. Instead add a method to to Service which allows you to set the limit. This will mean the actual capacity of the channel will need to set at Open() time. The capacity can default to 1024, to setting the limit explicitly is not required.

For testing purposes you will set it before you open and ensure the behaviour is what you expect.

Does that make sense?

shrtyk: Yeah, makes sense.
I changed it so the service stores the limit as a value and creates the limiter channel in `Open()`, using whatever was set before startup.
For the test, I added a private `setConnectionLimit(1)` helper instead of replacing the channel directly. I kept it private since it's only needed by the test and does not imply runtime resizing.

shrtyk: One small follow-up: the test still uses `len(s.connLimiterCh)` to wait until the limiter slot is taken or released.
I think that is OK for an internal package test, but I can hide it behind a small helper too if you prefer.

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	}
+
+	conn1, err := net.Dial("tcp", s.Addr())
+	if err != nil {
+		t.Fatalf("failed to dial first connection: %s", err)
+	}
+	t.Cleanup(func() { conn1.Close() })
+
+	// Hold the only limiter slot with an idle connection.
+	if !waitFor(2*time.Second, func() bool { return len(s.connLimiterCh) == 1 }) {
```
otoolep: This is brittle. Good unit tests (generally) test behavior, not internals. The test is (in this case) that the first connection is successful, the second will not be. The unit test should test that, not internal implementation details. 

This is a general rule, which applies here. Some unit testing in rqlite does test internals, but that is only in the case of very critical sections of the code (where database files are written for example). Connection management is not so critical that it requires testing internals. Testing behaviour is sufficient.

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	}
+
+	// The next connection should be rejected immediately, not queued.
+	conn2, err := net.Dial("tcp", s.Addr())
+	if err != nil {
+		t.Fatalf("failed to dial second connection: %s", err)
+	}
+	defer conn2.Close()
+
+	if err := conn2.SetReadDeadline(time.Now().Add(2 * time.Second)); err != nil {
```
otoolep: Why is this needed? If the connection should be rejected - i.e. close, then just attempt a write and see that that the write fails.

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	one := make([]byte, 1)
+	if _, err = conn2.Read(one); err == nil {
+		t.Fatalf("expected second connection to be rejected")
+	}
+
+	if nerr, ok := errors.AsType[net.Error](err); ok && nerr.Timeout() {
+		t.Fatalf("expected second connection to be rejected, got timeout")
+	}
+
+	if len(s.connLimiterCh) != 1 {
```
otoolep: Again, don't test internals.

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+		t.Fatalf("expected rejected connection not to consume limiter slot")
+	}
+
+	// Releasing the idle connection should free its limiter slot.
+	if err := conn1.Close(); err != nil {
```
otoolep: Yes, it should. But test this by checking that a new connection can be opened (as you do below).

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+		t.Fatalf("failed to close first connection: %s", err)
+	}
+
+	if !waitFor(2*time.Second, func() bool { return len(s.connLimiterCh) == 0 }) {
```
otoolep: Brittle, remove.

cluster/service_test.go
```diff
@@ -539,6 +540,93 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	}
+
+	// A new connection should be accepted once capacity is available again.
+	conn3, err := net.Dial("tcp", s.Addr())
+	if err != nil {
+		t.Fatalf("failed to dial third connection: %s", err)
+	}
+	defer conn3.Close()
+
+	if !waitFor(2*time.Second, func() bool { return len(s.connLimiterCh) == 1 }) {
```
otoolep: Remove.

cluster/service.go
```diff
@@ -275,14 +289,32 @@ func (s *Service) Stats() (map[string]any, error) {
 	return st, nil
 }
 
+// setConnectionLimit sets the maximum number of concurrent service connections.
+// It must be called before Open. Non-positive values use the default limit.
+func (s *Service) setConnectionLimit(n int) {
```
otoolep: Make this public. Add https://github.com/rqlite/rqlite/blob/master/internal/rsync/atomic.go#L91 to the Service object to check if it's open or closed. Return an error if this is called when the Service is open.

otoolep (commented):
Thanks for the latest updates. However the unit test is trying to do too much. It should be possible to test the Service's behavior purely using its public behavior.

cluster/service.go
```diff
@@ -275,14 +307,40 @@ func (s *Service) Stats() (map[string]any, error) {
...
+	if n <= 0 {
+		n = maxConcurrentConns
+	}
+
+	s.connLimit = n
+	return nil
+}
+
 func (s *Service) serve() error {
+	defer s.open.Unset()
```
otoolep: Only unset in `Close()`. Unsetting here is not what someone would expect.

cluster/service_test.go
```diff
@@ -539,6 +540,124 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+		t.Fatalf("failed to get node metadata on first connection: %s", err)
+	}
+	if meta.Url != "http://foo" {
+		t.Fatalf("failed to get correct node API address, exp %s, got %s", "http://foo", meta.Url)
+	}
+
+	// The first client's pooled connection remains open and holds the limiter slot.
+	c2 := NewClient(md, 30*time.Second)
+	var getErr error
+	if !waitFor(2*time.Second, func() bool {
```
otoolep: Why does this need to be in a polling function? It either succeeds or it doesn't.

cluster/service_test.go
```diff
@@ -539,6 +540,124 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+	}
+
+	// Releasing the idle connection should free its limiter slot.
+	if err := md.CloseConn(0); err != nil {
+		t.Fatalf("failed to close first connection: %s", err)
+	}
+
+	// A new client request should be accepted once capacity is available again.
+	c3 := NewClient(md, 30*time.Second)
+	if !waitFor(2*time.Second, func() bool {
```
otoolep: Ditto. There is no reason for this to poll.

cluster/service_test.go
```diff
@@ -539,6 +540,124 @@ func Test_ServiceClosesIdleConnection(t *testing.T) {
...
+		_, err := c3.GetNodeMeta(context.Background(), s.Addr(), noRetries, 500*time.Millisecond)
+		return err == nil
+	}) {
+		t.Fatalf("timed out waiting for client request to succeed after release")
+	}
+}
+
+type mockDialer struct {
```
otoolep, line 601: This seems to be getting very complicated. Why not just use a standard dialer?

otoolep:
Hi -- thanks for the latest changes, it's getting better. But the change still seems to be too complicated, polling functions, mock dialers. I don't follow why all of that is needed.

cluster/service_test.go
shrtyk: I used the mock dialer to keep this test on the same `NewClient` / `GetNodeMeta` path as the surrounding service tests.
The tracking in `mockDialer` is only there to make the limiter-slot release deterministic.
I can switch this to a plain `net.Dialer`, but then the test would need to bypass the public client API and use helpers like [`writeCommandReadResponse`](https://github.com/rqlite/rqlite/blob/6809aa3b6f373b86aa4e1265b9c47bf5d09e1819/cluster/client.go#L821).

otoolep: OK, let me take a look at the code again.

cluster/service_test.go
shrtyk: Yeah, you're right. The polling came from the earlier version of the test.
It was only needed because previous version also checked slot release with raw TCP connections. `Close()` doesn't free the server-side slot synchronously. The server goroutine has to notice the close and exit first. A follow-up connection could race that cleanup and be rejected even with correct code.

cluster/service_test.go
shrtyk: And yes, I jumped the gun when I said `writeCommandReadResponse` should be used. This can definitely be done more simply, without overcomplicating it.

shrtyk:
I left a note on the polling question above. Short version: it was carried over from the raw TCP version, where a connection opened right after `Close()` could race with the server goroutine releasing the slot.

Stepping back, I see two options:

1. Keep `mockDialer`, but remove the polling.
2. Go back to raw TCP connections. In that case, we still need a small poll after closing the first connection, because that can race with server-side cleanup.

One caveat: the race does not really disappear with `mockDialer`; the client path mostly hides it. `GetNodeMeta()` does more work before the next request reaches the server, and the retry path can recover if the first attempt is rejected too early. With raw TCP, the test hits the connection boundary directly, so the race is easier to expose.

otoolep:
Also, if you're concerned that calling Close() doesn't shut down the goroutines, yeah, that's an issue. The right way to handle that is `done` channels and so on. But we can fix that up later. But the unset call in the goroutine is not correct.

otoolep:
Keep the mock dialer for now, it's fine. The polling is definitely wrong, as you agree.

shrtyk:
I kept the mock dialer and removed the polling from the test. I also moved `s.open.Unset()` out of the `serve()` goroutine, and trimmed the test so it focuses on the connection limiter rather than the returned node metadata.

otoolep:
Thanks for your contribution @shrtyk

PR #2703, opened 2026-06-28
snapshot/plan: make CalcCRC32 and VerifyDB idempotent on a missing file

thecitymouse:
CalcCRC32 and VerifyDB now return nil when their target is already gone, matching Rename/Remove/WriteMeta/CopyFile.

The reap plan renames the full-snapshot dir as its last op; these two reads run earlier against a file inside it. Since Execute replays from the top on restart, a crash after the rename but before the plan file is removed re-runs the reads against the renamed-away path -> error out of NewStore -> node won't restart.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -58,3 +58,4 @@ _testmain.go
 
 # IDEs
 *.vscode
+.idea/
--- a/snapshot/plan/plan.go
+++ b/snapshot/plan/plan.go
@@ -135,6 +135,27 @@ func (p *Plan) Len() int {
 	return len(p.Ops)
 }
 
+// Done returns whether the plan has been completely executed. It determines
+// this by checking if the last operation in the plan is actually done.
+//
+// If a plan contains no steps then this function returns true.
+func (p *Plan) Done() bool {
+	if len(p.Ops) == 0 {
+		return true
+	}
+	last := p.Ops[len(p.Ops)-1]
+	if last.Type != OpRename {
+		return false
+	}
+	return dirExists(last.Dst) && !dirExists(last.Src)
+}
+
+// dirExists reports whether path exists and is a directory.
+func dirExists(path string) bool {
+	fi, err := os.Stat(path)
+	return err == nil && fi.IsDir()
+}
+
 // AddRename adds a rename operation to the plan.
 func (p *Plan) AddRename(src, dst string) {
 	p.Ops = append(p.Ops, Operation{
--- a/snapshot/store.go
+++ b/snapshot/store.go
@@ -649,11 +649,16 @@ func (s *Store) executeReapPlan(p *plan.Plan, planPath string) (int, int, error)
 	startT := time.Now()
 	defer recordDuration(reapExecuteDuration, startT)
 
+	if p.Done() {
+		s.logger.Printf("reap plan at %s already completed; dropping stale plan", planPath)
+		os.Remove(planPath)
+		return p.NReaped, p.NCheckpointed, nil
+	}
+
 	executor := plan.NewExecutor()
 	if err := p.Execute(executor); err != nil {
 		return 0, 0, fmt.Errorf("executing reap plan: %w", err)
 	}
-
 	if err := fsutil.SyncDirMaybe(s.dir); err != nil {
 		return 0, 0, fmt.Errorf("syncing store dir: %w", err)
 	}
--- a/snapshot/plan/plan_test.go
+++ b/snapshot/plan/plan_test.go
@@ -57,6 +57,38 @@ func TestAddOperations(t *testing.T) {
 	}
 }
 
+func TestPlanDone(t *testing.T) {
+	// Empty plan is finished.
+	if !New().Done() {
+		t.Fatal("empty plan should be Done")
+	}
+	dir := t.TempDir()
+	src := filepath.Join(dir, "src")
+	dst := filepath.Join(dir, "dst")
+	if err := os.MkdirAll(src, 0755); err != nil {
+		t.Fatal(err)
+	}
+	// Terminal rename not yet run -> not done.
+	p := New()
+	p.AddRename(src, dst)
+	if p.Done() {
+		t.Fatal("plan with pending rename should not be Done")
+	}
+	// Run the rename -> done.
+	if err := os.Rename(src, dst); err != nil {
+		t.Fatal(err)
+	}
+	if !p.Done() {
+		t.Fatal("plan with completed rename should be Done")
+	}
+	// Last op is not a rename -> not done.
+	p2 := New()
+	p2.AddRemoveAll(filepath.Join(dir, "whatever"))
+	if p2.Done() {
+		t.Fatal("plan whose last op is not a rename should not be Done")
+	}
+}
+
 // MockVisitor records calls for verification
 type MockVisitor struct {
 	Calls []string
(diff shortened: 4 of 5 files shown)


snapshot/plan/executor.go
```diff
@@ -170,6 +170,11 @@ func (e *Executor) CalcCRC32(dataPath, crcPath string) error {
 
 	sum, err := rsum.CRC32(dataPath)
 	if err != nil {
+		// During replay, this path may already have been moved by a later
```
otoolep: I'm not sure this is correct. I will have to think more about the issue you reported, but this seems too low-level a fix. In some cases being instructed to run a CRC32 on a missing file may be harmless, but in other cases it may indicate a real issue -- and in the latter case simply suppressing the error is wrong.. But that knowledge would exist at a higher-level in the stack, not here.

What I would expect to see is handling the issue you identified in a context-rich point in the execution, not at the leaf like this.

otoolep:
Thanks for your contribution, I will need to take a look when I get back to my dev environment. But first reaction is that this is almost certainly the wrong place for the fix.

snapshot/plan/executor.go
thecitymouse: > I'm not sure this is correct. I will have to think more about the issue you reported, but this seems too low-level a fix. In some cases being instructed to run a CRC32 on a missing file may be harmless, but in other cases it may indicate a real issue -- and in the latter case simply suppressing the error is wrong.. But that knowledge would exist at a higher-level in the stack, not here.
> 
> What I would expect to see is handling the issue you identified in a context-rich point in the execution, not at the leaf like this.

Yeah you're right, the leaf can't tell an already-renamed target from a real missing file, so suppressing it there was wrong. Honestly not sure what the right shape is intuitively it was to short-circuit in executeReapPlan if the terminal rename already landed check dst exists -> skip to cleanup), or maybe pull the CRC/verify out of the plan so replay stays idempotent but I don't know the constraints well enough to say. How would you approach it? Trying to learn the recovery model here more than anything.

otoolep:
>Since Execute replays from the top on restart, a crash after the rename but before the plan file is removed re-runs the reads against the renamed-away path -> error out of NewStore -> node won't restart.

As a first attempt, just explicitly check for this case during the check() phase. If you find a plan but also a non-temp directory, then this crash happened. Delete the plan and keep going?

otoolep:
Looking at the code I think my suggestion is a good approach. The reap plan itself is deleted outside of the plan-execute flow. So by symmetry it's consistent that this one edge case is also handled outside of that plan-execute flow. 

I would add one protection, which is probably necessary to do this right. If a reap plan exists, before deleting it, check its very last step (you will need to load the plan). If that last step is a rename to a snapshot directory that actually exists, then the crash happened at the point you identified and the reap plan is actually finished.

This is a real issue, though obviously very unlikely -- good catch. It does need to be addressed.

otoolep:
Plan file cleaned up outside of plan-execute flow: https://github.com/rqlite/rqlite/blob/v10.2.5/snapshot/store.go#L661

Therefore checking for a failure here should also be handled outside of the plan-execute flow. I would do it in `check()`, after this line: https://github.com/rqlite/rqlite/blob/v10.2.5/snapshot/store.go#L818

- If a plan exists, load it.
- check for its last step
- if it's a rename of a directory to a directory that actually exists, then the execute step completed, and we crashed before removing the plan.

otoolep:
Another way to think about it; here is the principle behind the fix: if the last step of an execution plan has actually been performed, then the system crashed just before removing the plan file itself.

thecitymouse:
> > Since Execute replays from the top on restart, a crash after the rename but before the plan file is removed re-runs the reads against the renamed-away path -> error out of NewStore -> node won't restart.
> 
> As a first attempt, just explicitly check for this case during the check() phase. If you find a plan but also a non-temp directory, then this crash happened. Delete the plan and keep going?

Reworked per your suggestion, executeReapPlan now detects when the terminal rename already landed (dst present, src gone) and drops the stale plan instead of replaying it. Leaf ops back to erroring strictly.

Test reproduces the post-crash state and asserts recovery; confirmed it fails without the fix, passes with it.

snapshot/store.go
```diff
@@ -644,16 +644,35 @@ func (s *Store) reapInternal() (int, int, error) {
 	return s.executeReapPlan(p, s.reapPlanPath)
 }
 
+// reapAlreadyCompleted reports whether the reap plan's terminal rename ran.
+// The rename moves full.path -> finalDir atomically and nothing else removes
+// the source, so dst present with src gone means it completed. A plan with no
+// rename should always be replay safe and reports false.
+func reapAlreadyCompleted(p *plan.Plan) bool {
```
otoolep: This should be a method on the Plan object, which will allow it to be unit tested. 

```Go
// Done returns whether the plan has been completely executed. It determines
// this by checking if the last operation in the plan is actually done.
// 
// If a plan contains no steps then this function returns true.
func (p *Plan) Done() bool {
...

File: .gitignore [hunk 3]
```diff
@@ -58,3 +58,4 @@ _testmain.go
 
 # IDEs
 *.vscode
+.idea/
```
otoolep, line 61: What is this? I don't object to it, but I want to understand what kind of file it is.

otoolep (changes requested):

.gitignore
thecitymouse: IDE ignores: .idea/ is JetBrains and .vscode is for VS Code.

snapshot/store.go
thecitymouse: Done, moved to Plan.Done(), unit tested for empty, pending rename, completed rename, and non-rename last op.

snapshot/plan/plan.go
```diff
@@ -135,6 +135,27 @@ func (p *Plan) Len() int {
 	return len(p.Ops)
 }
 
+// Done returns whether the plan has been completely executed. It determines
+// this by checking if the last operation in the plan is actually done.
+//
+// If a plan contains no steps then this function returns true.
+func (p *Plan) Done() bool {
```
otoolep, line 142: Hi -- I think we need to take a step back here. This doesn't work either, it assumes that the last step of a plan is a Rename. That is not guaranteed. It happens to be when reap writes a plan, but that is specific to reaping, not planning.

A plan must remain generic. The last step could be of any type.

Obviously this change is probably bigger than we initially realized, but the bug must be fixed comprehensively. The earlier PRs you proposed would fix the bug, but I don't believe a fix is urgent. We can take the time to fix this properly. Make sense?

otoolep:
Thanks again @thecitymouse -- I know there has been a lot of back and forth here. I'm happy to let this PR sit for a little while if you like and we can think about it. There is a real issue here, but the code construction may be a bit more complicated that we initially realised.

thecitymouse:
> Thanks again @thecitymouse -- I know there has been a lot of back and forth here. I'm happy to let this PR sit for a little while if you like and we can think about it. There is a real issue here, but the code construction may be a bit more complicated that we initially realised.

Yeah that's fine, this is a much bigger design problem than the actual edge case warrants right now. Closing for now rather than leaving it open. However thanks for the back and fourth actually learned a ton here. :)

otoolep:
Yeah, definitely an interesting - and real - edge case. Thanks again for reporting. I will CC you on future PRs which address this issue.

PR #2704, opened 2026-07-04
Correctly handle reap plan fully executed

otoolep:


otoolep:
Triggered by https://github.com/rqlite/rqlite/pull/2703

cc @thecitymouse

This is a substantial change and shows the cost of dealing with an implementation flaw while still maintaining backwards compatibility and maintainable code.

PR #2724, opened 2026-07-18
db: consolidate pragma and txStatus reads onto a single connection per pool

### LINKED ISSUE #2615: pragmas and txStatus each acquire separate connections

KochC:
## Summary

Fixes #2615.

### Problem

`pragmas()` called `v.QueryRow()` in a loop, potentially acquiring a **separate connection per pragma query** from the pool. On the `rw` pool (`MaxOpenConns=1`) this means each iteration waits for the previous connection to be returned before it can proceed.

`txStatus()` used `defer conn.Close()` **inside** a `for` loop. Go's `defer` runs when the surrounding function returns, not at the end of the loop iteration, so all connections remained open simultaneously until `txStatus()` returned.

### Fix

- **`pragmas()`** — acquire exactly one `*sql.Conn` per pool and run all `PRAGMA` reads on that connection, then close it before moving to the next pool. This mirrors the pattern already used by `memStats()`.
- **`txStatus()`** — extract the check into a `checkTx` helper that acquires one connection, inspects it, and `defer`s `Close()` correctly (scoped to the helper, not to the outer loop), so the connection is released before the next pool is checked.

### Tests

- `Test_DBStats_NoPragmaConnectionLeak` — verifies no extra rw connections remain open after `Stats()` returns.
- `Test_DBStats_PragmaFields` — verifies the pragmas map contains the expected keys for both `rw` and `ro` pools.

Full `go test ./db/` is green.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -1820,28 +1820,41 @@ func (db *DB) StmtReadOnlyWithConn(sql string, conn *sql.Conn) (bool, error) {
 }
 
 func (db *DB) pragmas() (map[string]any, error) {
-	conns := map[string]*sql.DB{
-		"rw": db.rwDB,
-		"ro": db.roDB,
+	pragmaNames := []string{
+		"synchronous",
+		"journal_mode",
+		"foreign_keys",
+		"wal_autocheckpoint",
+		"busy_timeout",
+	}
+
+	// readPragmas acquires a single connection from the given pool and reads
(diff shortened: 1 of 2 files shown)


otoolep:

PR #2736, opened 2026-07-29
failed DB swap could delete the original database

goingforstudying-ctrl:
Was looking at how a node recovers from a failed /boot or load and noticed the swap path in SwappableDB is pretty unforgiving right now. Swap() deletes the existing database files up front, and only then renames the new file into place and opens it. If that rename fails, or the open right after fails, the original database is just gone. The node is left with nothing at that path, which turns a recoverable error into actual data loss.

This is the issue in #2103.

What I changed: instead of deleting the old files, Swap now moves them aside to a .swap-backup name first. If anything fails after that point, a defer puts them back and reopens the original database before returning the error, so Swap either fully succeeds or changes nothing at all. Database paths stay fixed throughout, since that's a core assumption. The cleanup uses a named return plus defer, which keeps the happy path readable.

Also added a small RenameFiles helper in state.go next to RemoveFiles, since the backup and restore steps need to move any -wal and -shm files along with the main db file.

Testing:
- new test forces the rename to fail (source file on a different filesystem, so rename returns EXDEV) and checks the original database is back in place and queryable, with no backup files left behind. I confirmed this test fails against the old code — the next query just gets "sql: database is closed" because the original files were already deleted
- test that a successful swap leaves no backup files behind
- unit test for RenameFiles
- full db package suite passes with -race, plus the store-level load/boot tests that go through Swap (Test_SingleNodeLoad*, Test_SingleNodeBoot*)

One thing I'm not 100% sure about: if the process dies mid-swap, after the old files are moved aside but before the new one is renamed in, the backup is left on disk and nothing restores it on startup. That felt out of scope here, but happy to add recovery on open if you think it's worth handling.

### DIFF
--- a/db/db_test.go
+++ b/db/db_test.go
@@ -243,6 +243,45 @@ func Test_RemoveFiles(t *testing.T) {
 	}
 }
 
+func Test_RenameFiles(t *testing.T) {
+	srcDir := t.TempDir()
+	dstDir := t.TempDir()
+	mustCreateClosedFile(fmt.Sprintf("%s/foo", srcDir))
+	mustCreateClosedFile(fmt.Sprintf("%s/foo-wal", srcDir))
+	mustCreateClosedFile(fmt.Sprintf("%s/foo-shm", srcDir))
+
+	dst := fmt.Sprintf("%s/bar", dstDir)
+	if err := RenameFiles(fmt.Sprintf("%s/foo", srcDir), dst); err != nil {
+		t.Fatalf("failed to rename files: %s", err.Error())
+	}
+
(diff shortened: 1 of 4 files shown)


otoolep:
Please tell me how much an agent was involved with this -- thanks.

goingforstudying-ctrl:
An agent was involved in the code-related part of this contribution. Everything outside the code-related work was done by humans.

otoolep:
I'm going to hold off on this, I'm not sure it's worth the complexity. A failure to swap a database may need to be handled differently.

PR #2738, opened 2026-08-15
db: add slow query logging

Zhanfg:
Closes #447

## Summary
- add a configurable per-DB slow-query threshold with a 10-second default
- log the full SQL statement and elapsed duration when an execute or query exceeds the threshold
- keep `ForceQuery` on the query path so it is logged only once
- allow a non-positive threshold to disable slow-query logging
- add coverage for execute, query, `ForceQuery`, the default threshold, and disabling the feature

## Testing
- `go test ./db`
- Windows slow-query regression stress test (`TestSlowQueryLogging`, 50 runs)
- `gofmt` on all changed Go files
- `git diff --check`

## Coding-agent disclosure
This change was developed with coding-agent assistance and reviewed by me before submission.

PR #2739, opened 2026-08-15
Added configurable slow query logging

### LINKED ISSUE #447: Support slow query log

karangupta982:
## Summary

Closes #447

Adds slow-query logging for SQL statements that exceed a configurable execution-time threshold.

### Changes
- Added `--slow-query-threshold` configuration with a default threshold of 10s.
- Added slow-query logging for both `Execute` and `Query` paths.
- Added `SetSlowQueryThreshold()` to configure the threshold.
- Preserved the configured threshold when `SwappableDB` swaps the underlying database.
- Added tests covering slow `Execute`, slow `Query`, disabled logging, and threshold preservation across database swaps.

### Validation
- go test ./db
- go test ./...

All tests pass.

### DIFF
--- a/cmd/rqlited/config_flags.go
+++ b/cmd/rqlited/config_flags.go
@@ -110,6 +110,8 @@ type Config struct {
 	FKConstraints bool
 	// Maximum number of read-only connections to database
 	DBMaxReadOnlyConns int
+	// Minimum duration for logging slow SQL statements
+	SlowQueryThreshold time.Duration
 	// Period between automatic VACUUMs. If not set, not enabled
 	AutoVacInterval time.Duration
 	// Period between automatic 'PRAGMA optimize'. Set to 0h to disable
@@ -207,6 +209,7 @@ func Forge(arguments []string) (*flag.FlagSet, *Config, error) {
 	fs.DurationVar(&config.RaftReapReadOnlyNodeTimeout, "raft-reap-read-only-node-timeout", mustParseDuration("0h"), "Time after which a non-reachable read replica (non-voting) node will be reaped. If not set, no reaping takes place")
 	fs.BoolVar(&config.FKConstraints, "fk", false, "Enable SQLite foreign key constraints")
 	fs.IntVar(&config.DBMaxReadOnlyConns, "db-max-ro-conns", 256, "Maximum number of read-only connections to database")
+	fs.DurationVar(&config.SlowQueryThreshold, "slow-query-threshold", mustParseDuration("10s"), "Minimum duration for logging slow SQL statements")
 	fs.DurationVar(&config.AutoVacInterval, "auto-vacuum-int", mustParseDuration("0s"), "Period between automatic VACUUMs. If not set, not enabled")
 	fs.DurationVar(&config.AutoOptimizeInterval, "auto-optimize-int", mustParseDuration("24h"), "Period between automatic 'PRAGMA optimize'. Set to 0h to disable")
 	var tmpExtensionPaths string
--- a/cmd/rqlited/flags.toml
+++ b/cmd/rqlited/flags.toml
@@ -533,6 +533,17 @@ This places a limit on the number of read queries the node can service concurren
 default = 256
 
 [[flags]]
+name = "SlowQueryThreshold"
+cli = "slow-query-threshold"
+section = "SQLite database"
+type = "time.Duration"
+short_help = "Minimum duration for logging slow SQL statements"
+long_help = """
+SQL statements taking longer than this duration are logged as slow queries. Set to 0s to disable slow query logging.
+"""
+default = "10s"
+
+[[flags]]
 name = "AutoVacInterval"
 cli = "auto-vacuum-int"
 section = "SQLite database"
--- a/cmd/rqlited/main.go
+++ b/cmd/rqlited/main.go
@@ -432,6 +432,7 @@ func createStore(cfg *Config, ln *tcp.Layer, extensions []string) (*store.Store,
 	str.AutoOptimizeInterval = cfg.AutoOptimizeInterval
 	str.CompressSnapTransport = cfg.CompressSnapTransport
 	str.MaxReadOnlyConns = cfg.DBMaxReadOnlyConns
+	str.SlowQueryThreshold = cfg.SlowQueryThreshold
 	str.NoVerifyDB = true
 
 	if store.IsNewNode(cfg.DataPath) {
--- a/db/db.go
+++ b/db/db.go
@@ -197,6 +197,8 @@ type DB struct {
 	allOptimizedMu sync.Mutex
 
 	logger *log.Logger
+
+	slowQueryThreshold time.Duration // Minimum duration for slow SQL query logging
 }
 
 // PoolStats represents connection pool statistics
@@ -1229,6 +1231,9 @@ func (db *DB) executeStmtWithConn(ctx context.Context, stmt *command.Statement, 
 	}()
 	response := &command.ExecuteQueryResponse{}
 	start := time.Now()
+	if !stmt.ForceQuery {
+		defer db.logSlowQuery(start, stmt.Sql)
+	}
 
 	parameters, err := parametersToValues(stmt.Parameters)
 	if err != nil {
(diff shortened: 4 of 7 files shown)


otoolep:
It is highly unusual to receive two similar PRs for the same issue, in quick succession. Is this connected to https://github.com/rqlite/rqlite/pull/2738 ?

Please review https://github.com/rqlite/rqlite/blob/master/CONTRIBUTING.md#use-of-coding-agents

karangupta982:
Hi @otoolep  ,

No, #2739 and #2738 are not connected. I don't know the author of #2738 and I was not aware of that PR when I started working on #447.

I have been working on understanding #447 and the relevant rqlite architecture for more than two weeks. I used GPT to help me understand the repository architecture and trace the relevant code paths and some part of code where required. The implementation and code changes in this PR were written by me.

I had read the Coding Agents section of CONTRIBUTING.md before starting the work and understood that coding agents are permitted.

karangupta982:
Hi @otoolep , just following up on this PR. Please let me know if there is anything I should modify or address. Thanks!

otoolep:
Thanks for your contribution. I will need to think about whether to enable this flag. The idea for a slow query log was an idea, I'm not sure whether rqlite needs it.

Thinking about this some more, I think it may be better as part of a larger query logging solution generally. Most databases offer a query log. That might be what's actually missing.

I need to think about it.

karangupta982:
Thanks @otoolep , i understand your point that the missing functionality may be broader query logging rather than only a slow-query log.

I did some research into how other databases approach this:

PostgreSQL supports general statement logging through log_statement, as well as duration based logging through log_min_duration_statement. Setting the duration threshold to 0 effectively logs the duration of every statement.
https://www.postgresql.org/docs/16/runtime-config-logging.html

MySQL provides both a General Query Log, which records SQL statements received by the server, and a Slow Query Log, which records statements exceeding a configurable execution-time threshold.
https://dev.mysql.com/doc/refman/26.7/en/query-log.html
https://dev.mysql.com/doc/refman/26.7/en/slow-query-log.html

SQLite already provides APIs that can tell us which SQL query was executed and how long it took. We can check if we can use these APIs for rqlite's query logging.
https://sqlite.org/c3ref/profile.html 

MongoDB has configurable database profiling, with modes for disabled, slow operations, or all operations, along with filtering and sampling options.
https://www.mongodb.com/docs/manual/reference/database-profiler/
https://www.mongodb.com/docs/manual/tutorial/manage-the-database-profiler/

Based on this, I think a possible direction for rqlite would be to introduce a general query logging/profiling facility, rather than making slow-query logging a standalone feature.

At a high level, I would see it working something like:

- Query logging is disabled by default.
- Provide a configuration option to control what gets logged.
- Support logging all SQL statements.
- Support logging only statements exceeding a configurable duration threshold.
- Keep the log entry format useful and consistent, at minimum including the SQL statement and execution duration.
- We should put the logging logic at the database execution layer where queries are actually executed, so that both read and write queries are covered without adding the same logging code in multiple places.
- Add tests for the different logging modes and make sure logging can be disabled without changing normal query execution behaviour.

I haven't started designing or implementing this broader approach yet. Does this direction match what you had in mind, or would you prefer a different approach for rqlite?

If this is roughly what you're thinking, I can look into the existing rqlite architecture in more detail and come back with a complete design and implementation plan before making further changes to the PR.

otoolep:
This sounds reasonable. BTW those SQLite functions are deprecated. https://sqlite.org/c3ref/trace_v2.html is the only one supported.

I agree a more general query logging solution is the right answer here. They would be logged in the DB code. I think rqlited would be modified to accept a single flag, which would take a config file controlling the logging.

But let's start small. I would like see the db/ package modified first, to support query logging. Forget everything else for now. We can proceed in steps and if we like it it will eventually be enabled. But we do the db/ package first, and nothing else. 

Show me the PR for adding query logging to the DB code. Define the config object that is passed to the DB to control logging, and show me the code that does the logging in the DB layer.

karangupta982:
Thanks @otoolep, I looked further into the db/ layer and the SQLite driver to understand how query logging could fit into the current architecture.

The rqlite SQLite driver already exposes sqlite3_trace_v2() through SQLiteConn.SetTrace(). The existing ConnectHook in db/driver.go also gives us a place to configure each SQLite connection, which would allow tracing to work across both the read/write connection and the read-only connection pool.

There are two things I found with this approach:

1. SetTrace() is currently behind the sqlite_trace build tag, so using it would require enabling that tag in the rqlite build configuration. SQLite tracing operates at the connection level, so it would also see SQL generated internally by rqlite, such as PRAGMAs, rather than only the user queries.

2. The other option is to implement the logging directly in the db/ execution layer. The existing executeStmtWithConn() and queryStmtWithConn() paths already measure execution time, and these cover the read and write query paths. This would keep the change entirely within db/ and would naturally log the SQL statements coming through rqlite's query execution layer.

I would go towards using sqlite3_trace_v2() because it is the SQLite-supported tracing mechanism and gives us SQL execution information at the database level, but I wanted to confirm the build-tag direction with you before designing the DB configuration and implementation around it.

Before I finalize the DB-level design, which approach would you prefer?

A. Use the existing sqlite3_trace_v2() / SetTrace() support, including enabling the sqlite_trace build tag, w.hich would be implemented outside db package.

B. Implement query logging directly in the Go db/ execution layer without using SQLite tracing.

I will keep this PR limited to the db/ package and its tests as you requested. No rqlited or store configuration will be added at this stage.

otoolep:
Agreed, `sqlite3_trace_v2` looks most promising. Presumably we're only interested in callbacks of type `SQLITE_TRACE_STMT`. 

I have no objection to enabling the build tag as needed, since I maintain a fork of the underlying SQLite driver. Before we go down this path, I would like to ensure we understand  `sqlite3_trace_v2` fully. If I was doing this, I would write a quick prototype Go program on top of the mattn library, enabling the build. Send in some SQLite statements, and see how it looks.

WDYT?

karangupta982:
Thanks @otoolep for the direction, I built the prototype using github.com/rqlite/go-sqlite3 v1.49.0 with -tags sqlite_trace. I found:

SQLITE_TRACE_STMT gave
```
Event: STMT
SQL:   INSERT INTO t (name, val) VALUES (?, ?)
ExpandedSQL: INSERT INTO t (name, val) VALUES ('alice', 42)
Duration: 0 ns
```
STMT fires when the statement starts, so it gives us the SQL but not the execution duration.

SQLITE_TRACE_PROFILE gave
```
Event: PROFILE
SQL:   (empty)
Duration: 2000000 ns (2ms)
```
PROFILE fires after the statement finishes, so it gives us the execution duration but not the SQL.

The two events have the same (ConnHandle, StmtHandle), so we can associate the STMT event with its corresponding PROFILE event.

Conclusion: If we want to log SQL and duration together, we need both events. STMT alone does not provide the duration, and PROFILE alone does not provide the SQL.

What also observed:
1. Internal PRAGMAs such as PRAGMA journal_mode and PRAGMA busy_timeout appear as normal trace events.
2. Trigger execution can produce comment-style events such as -- TRIGGER trg_after_insert.
3. Statements with syntax errors do not produce trace events, while constraint violations do.
4. Installing the trace through ConnectHook works for both the write connection and the pooled read-only connections.

Proposed implementation:
Based on these results, my current plan would be:
1. Subscribe to STMT and PROFILE both.
2. Capture the SQL from STMT and associate it with the corresponding PROFILE event.
3. Use the duration provided by SQLite for the initial implementation.
4. Log all trace statements initially, including PRAGMAs and trigger events. This keeps the DB logger generic and avoids making assumptions about which statements should be considered internal. Filtering can be added later as a configurable option if the output proves too noisy.

One question before I finalize QueryLogConfig:
My preference is to log everything initially, including internal PRAGMAs and trigger events, rather than filtering them in the DB layer.

Does that match your expectation, or would you prefer the DB layer to filter these statements from the beginning?

Please confirm if this direction looks right to you. Once confirmed, I would like to propose the detailed implementation plan for the DB layer, including the QueryLogConfig design, trace callback handling, statement - profile pairing, and tests, so that the complete approach is clear before implementation.

otoolep:
OK, using the Trace functions sounds good. What you outlined makes sense to me. However I will need to enable that build flag on my fork, and then update top of tree for rqlite.

One thing you could start thinking about is how to code this so it's testable and self-contained. A model I used to implement CDC (which also involves hooks) is here: https://github.com/rqlite/rqlite/blob/master/db/cdc.go

I would expect a similar type, this time with hooks for Tracing and Profiling. That would be the core of the work, which is then integrated into the db layer in a neat manner. So I would start thinking about how the db layer should be modified.

PR #2740, opened 2026-08-17
Fix database loss on failed swap (#2103)

### LINKED ISSUE #2103: Swapping a database should not delete the files unless swapping worked

karangupta982:
Closes #2103

### Changes

- Move the existing database files to a temporary path before swapping.
- Use a unique temporary path to avoid conflicts with existing swap files.
- Restore the original database files if the swap fails.
- Reopen the original database and recreate the checkpoint manager after rollback.
- Create the new checkpoint manager before making the new database active.
- Added tests for failed swaps, stale swap files, and WAL sidecar files.

### Validation

- `go test ./db`
- `go test ./...`

### DIFF
--- a/db/swappable_db.go
+++ b/db/swappable_db.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"path/filepath"
 	"regexp"
 	"sync"
 	"time"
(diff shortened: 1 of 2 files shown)


otoolep:
I'm going to hold off on this, I'm not sure it's worth the complexity. A failure to swap a database may need to be handled differently.

PR #2741, opened 2026-08-24
Feat-447: initial query logging configurations

karangupta982:
fix: #2739

Implements the initial query logging infrastructure using SQLite sqlite3_trace_v2.

This follows the self-contained hook pattern discussed in db/cdc.go:

1. Adds QueryLogConfig and QueryLogger as the core query logging component.
2. Uses SQLITE_TRACE_STMT to capture SQL and SQLITE_TRACE_PROFILE to capture execution duration.
3. Pairs STMT/PROFILE events using (ConnHandle, StmtHandle).
4. Integrates the logger through ConnectHook so tracing is installed on each new SQLite connection.
5. Adds unit tests for the tracing/pairing logic and integration tests against a real SQLite database.
6. Keeps the initial implementation intentionally minimal: no CLI/store configuration, filtering, or slow-query thresholds yet.

All tests pass with the sqlite_trace build tag.

This is intended as the base implementation for review before adding any further configuration or functionality.


karangupta982:
Hi @otoolep , tests are failing because the rqlite build is not compiling the SQLite driver with sqlite_trace; the new files therefore cannot see TraceInfo, TraceConfig, SetTrace, etc. As discussed, this appears to be the build-tag change you mentioned you would make on your fork.

db/query_log.go
```diff
@@ -0,0 +1,100 @@
+package db
+
+import (
+	"fmt"
+	"log"
+	"sync"
+
+	"github.com/mattn/go-sqlite3"
```
otoolep, line 8: Use the rqlite fork, like the rest of the db package.

db/query_log.go
```diff
@@ -0,0 +1,100 @@
...
+// both carry the same pair.
+type traceKey struct {
+	ConnHandle uintptr
+	StmtHandle uintptr
+}
+
+// QueryLogger is a self-contained query logging component that receives
+// SQLite trace_v2 events (STMT and PROFILE) and produces log lines
+// containing the SQL statement and its execution duration.
+// It follows the same design pattern as CDCStreamer: a dedicated type
```
otoolep: Remove these last three lines. Reference to other concrete only clutters the code, someone reading this has no idea which came first, nor do they care.

(I'm guessing an agent wrote these comments).

db/query_log.go
```diff
@@ -0,0 +1,100 @@
...
+		ql.mu.Lock()
+		sql, ok := ql.pending[key]
+		if ok {
+			delete(ql.pending, key)
+		}
+		ql.mu.Unlock()
+
+		if !ok {
+			// PROFILE without a preceding STMT — shouldn't happen normally,
+			// but be safe and skip.
```
otoolep: Emit a log message. We need someway of knowing this happened as it could indicate a bug.

db/query_log.go
```diff
@@ -0,0 +1,100 @@
...
+			return 0
+		}
+
+		durationMs := info.RunTimeNanosec / 1_000_000
```
otoolep: For clarify convert durationMS to a time.Duration object first, and then operate on that.

db/query_log_driver.go
```diff
@@ -0,0 +1,47 @@
...
+	"github.com/mattn/go-sqlite3"
+)
+
+// NewQueryLogDriver returns a new driver with query logging enabled.
+// It registers a SQLite3 driver that installs the given QueryLogger's
+// TraceHook on every new connection via ConnectHook. The driver also
+// disables checkpoint-on-close (same as DefaultDriver).
+// name must be unique across the process. If a driver with the given
+// name already exists, a panic will occur — same behavior as NewDriver.
+func NewQueryLogDriver(name string, ql *QueryLogger) *Driver {
```
otoolep: Use Once pattern like all other drivers so you don't get a panic.

This is a reasonable start, but I may rework the driver code more generally. We can proceed for now.

db/query_log_integration_test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	_, err = db.ExecuteStringStmt("INSERT INTO t (name) VALUES ('bob')")
+	if err != nil {
+		t.Fatalf("INSERT failed: %s", err)
+	}
+	rows, err := db.QueryStringStmt("SELECT name FROM t")
+	if err != nil {
+		t.Fatalf("SELECT failed: %s", err)
+	}
+	// Verify the query actually worked
```
otoolep: Check for the actual expected counts, not that they are not zero.

db/query_log_integration_test.go
```diff
@@ -0,0 +1,183 @@
...
+		t.Fatalf("failed to open database: %s", err)
+	}
+	defer db.Close()
+
+	_, err = db.ExecuteStringStmt("CREATE TABLE t (id INTEGER PRIMARY KEY, val INTEGER)")
+	if err != nil {
+		t.Fatalf("CREATE TABLE failed: %s", err)
+	}
+
+	// Execute multiple statements in a transaction
```
otoolep: This comment is incorrect, it's a bulk request, not a transaction.

db/query_log_test.go
```diff
@@ -0,0 +1,318 @@
...
+		StmtHandle:    0x2,
+		StmtOrTrigger: "SELECT 1",
+	})
+	ql.TraceHook(sqlite3.TraceInfo{
+		EventCode:      sqlite3.TraceProfile,
+		ConnHandle:     0x1,
+		StmtHandle:     0x2,
+		RunTimeNanosec: 5_000_000,
+	})
+	// No panic = pass.
```
otoolep: Comment is redundant.

otoolep (commented):
Seems reasonable, is generally the shape I would expect -- thanks.

The number of drivers is starting to proliferate, I will need to think about consolidating. However we can do that after the merge. The idea is to get the driver in first, all tests passing, and then we can move to integrating with the main code base.

otoolep:
Merge master into your branch.

db/query_log.go
karangupta982: db.go, driver.go, and state.go uses the same path. Replace directive in go.mod can handle this.

otoolep: Ah, OK.

otoolep:
Thanks, I will merge this now. This will allow you to show how this would be integrated with the actual DB code.

The Driver stuff needs work, I need to think about how to consolidate the various drivers.

karangupta982:
Thanks @otoolep  I will now work on integrating the query logger with the actual DB code, will share the implementation plan for verification.

otoolep:
I suspect it will go like this:
- We will use a dedicated Driver (the one you created)
- The Store instantiates the Driver.
- Open the database as normal with that driver.

This means the Store must be passed query logging controls if set. We just do that much, but do not enable logging. That is one PR. The second, and, last PR will be the addition of the command line flag allowing operators to enable.

To give a sense where this is going, the flag will be something like this:

```
-query-log <path>
```

Where path will be a path to a JSON file with various options (mostly log only if greater than a certain time, whether to log expanded SQL). However if `<path>` is the special value `stdee` then logging just goes to `stderr` with all queries logged. If not set, then there is no query logging.

That gives you a sense of the config that needs to be passed to the DB layer, and driver. The driver needs to be more sophisticated.

karangupta982:
Thanks @otoolep for the direction. I have raised the second PR.

Before starting the third PR, I would like to work on addressing the sync.Once limitation in QueryLogDriver, so that multiple calls with different QueryLogger instances work correctly, if you are okay with that.

Once that is addressed, I will start working on the third PR as discussed.

karangupta982:
Hi @otoolep ,I understand the overall direction for third PR. Before I start implementing, I just want to confirm a few things so I don't make the wrong assumptions:

1. confirming , should the special value be exactly stdee as written, or did you mean stderr or a typing error? My understanding is that this special value means logging all queries directly to stderr, without reading a JSON configuration file.

for a json path, if the file doesn't exist, the JSON is invalid, or the configuration is empty or incomplete, should we return a configuration error? My preference would be to return an error for an invalid configuration rather than silently fall back to logging everything to stderr.

2. for the JSON configuration, I understand the main options to be the minimum query duration and whether expanded SQL should be logged. If there are no fixed field names yet, is it OK if I keep it simple like:
```
{
  "duration": "100ms",
  "expanded_sql": true
}
```
and you can review/change the names or format if needed?

3. for the driver, I think we also need to address the current sync.Once in QueryLogDriver(). At th moment, sync.Once means the SQLite driver is registered only on the first call to QueryLogDriver(), and the ConnectHook including the QueryLogger passed to it, is fixed at that first registration. So if QueryLogDriver() is called again with a different QueryLogger, the newly supplied logger would not be used by the registered driver.

since third PR will have configurable query logging, and we already have DefaultDriver, NewDriver, CheckpointDriver, and QueryLogDriver, I think we should address this as part of making the driver more sophisticated. My initial thought is to avoid tying the registered driver's ConnectHook to the first QueryLogger, and instead have the driver configuration determine the appropriate behavior for each database. Does this sound like the direction you had in mind, or would you prefer a different approach for consolidating the drivers?

otoolep:
>confirming , should the special value be exactly stdee as written, or did you mean stderr or a typing error? My understanding is that this special value means logging all queries directly to stderr, without reading a JSON configuration file.

Actually, yeah, it should be `stderr`. In that case just log to standard error, like it does today.

>My preference would be to return an error for an invalid configuration rather than silently fall back to logging everything to stderr.

Me too. If there is a problem with the JSON file then exit. Do not fall back. The Config structure looks fine as a starting point.

>So if QueryLogDriver() is called again with a different QueryLogger, the newly supplied logger would not be used by 
the registered driver.

I see. You are right, though it's probably not an issue in practise, it's not great.

Yes, the Drivers need to be cleaned up. In fact, perhaps we should do that consolidation next, before anything else. Are you interested in doing that?

karangupta982:
Hi @otoolep , Yes, I am interested in doing the consolidation.

I went through the current drivers and their callers. My understanding is that the different driver configurations currently have separate registration and ConnectHook paths, which makes it difficult to compose features such as extensions and query logging.

My plan:

1. Introduce a common driver construction path where the ConnectHook is built by composing the needed behaviors (checkpoint config, tracing, etc.) rather than having separate hooks per driver.
2. Extensions stay as part of the SQLiteDriver config and can coexist with other features. 
3. Rework the driver registration so the configuration is not tied to the first QueryLogDriver() call, addressing the current sync.Once limitation.
4. Keep the existing driver behavior covered by tests and add coverage for combinations such as extensions + query logging...

I will keep the -query-log flag, JSON configuration, duration filtering, expanded SQL, and stderr handling for the following query-logging PR.

otoolep:
Sounds reasonable, show me what you have in mind.

PR #2745, opened 2026-08-27
feat-447: integrate query logger with store

### LINKED ISSUE #447: Support slow query log

karangupta982:
Fix: #447

Integrates the existing query logging support with the Store.

### Changes

- Add `QueryLogConfig` to `store.DBConfig`.
- Configure the Store to use `QueryLogDriver` when query logging is configured.
- Add Store-level tests covering:
  - Query logging disabled.
  - Query logging enabled and SQL statements being logged.

As discussed in #2741

### DIFF
--- a/store/db_config.go
+++ b/store/db_config.go
@@ -1,9 +1,8 @@
 package store
 
 import (
+	"log"
 	"path/filepath"
-
-	sql "github.com/rqlite/rqlite/v10/db"
 )
 
 // DBConfig represents the configuration of the underlying SQLite database.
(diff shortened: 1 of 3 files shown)


store/db_config.go
```diff
@@ -1,6 +1,10 @@
 package store
 
-import "path/filepath"
+import (
+	"path/filepath"
+
+	sql "github.com/rqlite/rqlite/v10/db"
```
otoolep: Not this pattern. If the Config object needs new fields then they should called called out explicitly using types from the standard library only, such as path, interval, and whether fully logging should be required?

Why? No reason, but that is the pattern that is there already and I want to keep it. Let's see how that goes. If it's too unwieldy, perhaps I will change my decision.

PR #2749, opened 2026-08-30
fix(snapshot): skip no-op rename in reap plan

### LINKED ISSUE #2746: Snapshot reaping can (in rare cases) permanently prevent a node from restarting
Potential bug found on v10.2 via Antithesis, on a 4-node cluster (3 voters + 1 read-only) built from tip of `master` (e1b34edc) running inside Docker containers on Linux.

In rare cases, a node can write a snapshot-maintenance plan to disk that can never succeed. Once that happens the node exits on every subsequent start, and no amount of restarting recovers it. The bad plan survives reboots because it is stored in the node's data directory, which is the same mechanism that normally makes snapshot maintenance crash-safe.

## Impact

***Liveness***: **A node goes down and stays down. Restarting it does not help, and it never rejoins the cluster on its own.**

The node exits during startup, before it begins serving. Any subsequent restart attempt sees it exit again immediately, in an unbounded loop. Nothing about the node's state improves between attempts, because the failure happens while reading data that is already on disk. Recovery would require manual intervention on the node's data directory.

The bug likely affects all v10.x releases (the root cause may have been added in 2b568e85d, see below).

## Triggers

Context: rqlite periodically compacts its snapshots — folding accumulated changes into the base snapshot and discarding what is no longer needed (see [`snapshot/DESIGN.md`](https://github.com/rqlite/rqlite/tree/master/snapshot/DESIGN.md)). That compaction is planned first and subsequently executed, with the plan written to disk so an interrupted compaction can be resumed after a crash.

The bug needs all of the following to hold at the moment a compaction is planned:

- The most recent base snapshot has *no newer incremental snapshots on top of it*. This is the state a node is in immediately after receiving a snapshot from the leader, which happens when a node falls far enough behind that the leader can no longer send it individual log entries.
- That base snapshot still carries data to fold in, so the compaction has work to do and reaches its final step.
- There is at least one older snapshot to discard, so compaction runs at all.
- The clock reading taken when the compaction is planned matches the one taken when that snapshot was originally named — see Root Cause. In practice this means either **less than a millisecond elapsed between the two**, or **the system clock was stepped backwards** in between (for example by an NTP correction or a VM migration), since the reading used is wall-clock time and is not monotonic.

Once the plan is written, the node fails on the next restart and on every restart after that.

## Symptom

While running, repeatedly:

```
reap failed: executing reap plan: rename <dir> <dir>: file exists
```

On every subsequent start, fatally:

```
failed to open store: failed to create snapshot store: executing reap plan: executing reap plan: rename <dir> <dir>: file exists
```

Note that the source and destination paths in the message are identical. While a POSIX rename of this kind would be a silent no-op, the Go standard library's [`os.Rename()` has a different behavior of returning an error](https://pkg.go.dev/os#Rename).

## Sample logs

From the affected node. The snapshot directory name is identical on both sides of the rename throughout:

```
[snapshot-store] 2026/08/26 21:46:40 reap failed: executing reap plan: rename /rqlite/file/data/wsnapshots/17-1039-1787780800033 /rqlite/file/data/wsnapshots/17-1039-1787780800033: file exists
[rqlited]        2026/08/26 21:46:42 failed to open store: failed to create snapshot store: executing reap plan: executing reap plan: rename /rqlite/file/data/wsnapshots/17-1039-1787780800033 /rqlite/file/data/wsnapshots/17-1039-1787780800033: file exists
[snapshot-store] 2026/08/26 21:47:18 reap failed: executing reap plan: rename /rqlite/file/data/wsnapshots/17-1039-1787780800033 /rqlite/file/data/wsnapshots/17-1039-1787780800033: file exists
[snapshot-store] 2026/08/26 21:47:42 reap failed: executing reap plan: rename /rqlite/file/data/wsnapshots/17-1039-1787780800033 /rqlite/file/data/wsnapshots/17-1039-1787780800033: file exists
[rqlited]        2026/08/26 21:47:45 failed to open store: failed to create snapshot store: executing reap plan: executing reap plan: rename /rqlite/file/data/wsnapshots/17-1039-1787780800033 /rqlite/file/data/wsnapshots/17-1039-1787780800033: file exists
```

The failure recurs identically across restarts, minutes apart, on the same snapshot ID.

## Root cause

A reap plan can contain a rename whose source and destination are the same directory. `os.Rename` fails with `EEXIST` when the destination is an existing directory — including when it is the source — and nothing in the plan, the executor or the resume path handles that case.

**Where the two names come from.** Snapshot directories are named by `snapshotName(term, index)` at: 
https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/store.go#L960-L964

Thus function is called in exactly two places for this bug to trigger. `Store.Create` uses it to name a snapshot when one is created, and `Store.reapInternal` uses it again at the final step of a compaction to name the result:

https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/store.go#L620-L628

**Why they can be equal.** When there is no incremental snapshot newer than the base one, `newest` resolves to `full` — the very snapshot being renamed. `newID` is then rebuilt from that snapshot's own term and index, so it agrees with the existing directory name in every field except the trailing timestamp. There are two ways that the timestamp component `msec` can also be equal:

1. **Less than a millisecond elapsed** between naming the snapshot and planning the compaction. The value is a millisecond quotient, so the two nanosecond readings differ while the quotient does not.
2. **The clock moved backwards.** `UnixNano` is wall-clock time and carries no monotonic guarantee, so an NTP step or a VM migration can make a later call return an earlier reading. This requires no sub-millisecond timing at all, and can reproduce the earlier value across an arbitrarily long interval.

[`AddRename`](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/plan/plan.go#L139-L145) then appends a rename with `Src == Dst` (persisted to the plan),

**Why it creates a permanent issue.** 

- `Executor.Rename` is documented as idempotent, but only for the case where the source is missing and the destination is present. It inspects `os.IsNotExist`, and `EEXIST` is not that, so the error propagates.

  https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/plan/executor.go#L27-L41)
- `Checker.RenameDone`, used by `check()` to detect a plan that already ran to completion, returns false while the source exists. With `Src == Dst` the source always exists, so the plan is never recognized as complete.
  https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/plan/checker.go#L25-L37
- The plan file is removed only after a successful execution, so it is never removed.

Since the source is also the destination, it exists on every attempt, and the plan fails identically forever. `NewStore` resumes the persisted plan through `check()` before the node starts serving, so the failure is fatal at startup rather than merely noisy at runtime.

## Potential fix

In `snapshot/store.go`, do not plan a rename that is a no-op:

```go
 		// 7. Rename to new snapshot name. The end result of the Reaping process
 		// will be a new full snapshot with a new ID. That ID is generated from
 		// the newest snapshot's index and term, and the current timestamp.
+		// Skip the rename if it would be a no-op.
 		finalDir := filepath.Join(s.dir, newID)
-		p.AddRename(full.path, finalDir)
+		if full.path != finalDir {
+			p.AddRename(full.path, finalDir)
+		}
```

This prevents the bad plan from ever being written. 

## Repro / Validation

In my subsequent testing with Antithesis, the above fix made the failure go away. I don't know if this is the logically correct solution though. Happy to validate any other fix.

It is also possible to test this locally with a unit test [using `synctest` to freeze the clock](https://pkg.go.dev/testing/synctest#hdr-Time) between `snapshotName` calls.

nityanand123gupta:
Fixes #2746.

**Disclosure:** I used a coding agent (Claude) to investigate and implement this fix, under my direction and review. The issue itself already contained an exceptionally thorough root-cause analysis and a validated Antithesis repro from the reporter — I read and understood that analysis, traced the actual code at the referenced locations to confirm it, and applied their suggested fix rather than treating their write-up as a black box.

## Understanding of the bug

`Store.reapInternal`'s compaction step always plans a final rename of the full snapshot's directory to a newly computed name (`newID`, derived from `newest`'s term/index plus the current millisecond timestamp). When there's no incremental snapshot newer than `full`, `newest == full`, so `newID` differs from `full`'s existing directory name only in its timestamp component. If that timestamp happens to collide with the one already in `full`'s name (sub-millisecond timing, or a backward clock step), the planned rename has `Src == Dst`.

`os.Rename` on some platforms (documented as reproducing on Linux) returns `EEXIST` when the destination already exists — including when it's the same path as the source. Since the rename is persisted to the on-disk plan before execution, and `Checker.RenameDone` can never see the source disappear when `Src == Dst`, a plan containing this rename can never be recognized as complete. Because the plan file is only removed after a successful execution, the failure is replayed identically on every subsequent restart, permanently preventing the node from starting.

## Fix

Skip adding the rename to the plan at all when `full.path == finalDir` (`snapshot/store.go`), matching the reporter's suggested fix.

## Testing

Added `Test_Store_ReapNoOpRename` in `snapshot/store_test.go`, which sets up the exact trigger conditions (a full snapshot carrying its own bundled, not-yet-checkpointed WAL — the state a node is in right after receiving a snapshot from the leader — with no incremental on top of it, and an older snapshot for the reap to discard) and pins the full snapshot's ID to the current millisecond so the recomputed name collides with it. The test asserts the reap completes, produces valid queryable data, and that no `Src == Dst` rename is ever left in a persisted plan.

**Caveat I want to be upfront about:** I developed and ran this on Windows, where I confirmed directly that `os.Rename(path, path)` on an existing directory returns `nil`, not an error — so I could not locally reproduce the pre-fix failure via an actual failing rename. The test therefore can't turn red on this platform even without the fix; I verified this by reverting only the `store.go` change and confirming the test still passes locally. What it does verify, on any platform: the reap completes correctly and produces valid data under the exact conditions described in the issue. I'm relying on the reporter's own Antithesis-validated confirmation (mentioned in the issue) that removing the redundant rename makes the real EEXIST failure go away on the platform where it reproduces, and expect CI here to exercise that path properly if it runs on Linux.

`go build ./...` and `go test ./snapshot/...` pass; `gofmt` and `go vet` are clean on the changed files.

### DIFF
--- a/snapshot/store.go
+++ b/snapshot/store.go
@@ -642,8 +642,20 @@ func (s *Store) reapInternal() (int, int, error) {
 		// 7. Rename to new snapshot name. The end result of the Reaping process
 		// will be a new full snapshot with a new ID. That ID is generated from
 		// the newest snapshot's index and term, and the current timestamp.
+		//
+		// Skip the rename if it would be a no-op. This happens when there are no
+		// incremental snapshots newer than full (so newest == full) and the
+		// millisecond-resolution timestamp component of newID happens to match the
+		// one already in full's name -- e.g. because less than a millisecond
+		// elapsed since full was named, or the wall clock stepped backwards. In
+		// that case os.Rename fails with EEXIST since source == destination, and
+		// since the rename is persisted in the plan, every future restart replays
+		// the same failing rename forever (see checker.RenameDone, which never
+		// considers a Src == Dst rename done).
 		finalDir := filepath.Join(s.dir, newID)
-		p.AddRename(full.path, finalDir)
+		if full.path != finalDir {
+			p.AddRename(full.path, finalDir)
+		}
 	}
 
 	// Persist the plan to disk for crash recovery.
--- a/snapshot/store_test.go
+++ b/snapshot/store_test.go
@@ -984,6 +984,90 @@ func Test_Store_Reap(t *testing.T) {
 
 }
 
+// Test_Store_ReapNoOpRename is a regression test for the case where the reap
+// plan's final rename is a no-op (source and destination are the same
+// directory). This happens when the newest snapshot has no incremental
+// snapshots on top of it (so it plans to rename itself to itself) and the
+// millisecond-resolution timestamp component of the recomputed name happens
+// to match the one already in its own name -- which is guaranteed here since
+// we give it an explicit ID and, within a single goroutine with no blocking
+// calls, consecutive time.Now() calls return the same instant.
+//
+// Before the fix, planning this rename unconditionally meant the persisted
+// plan always contained a Src == Dst rename. On platforms where renaming a
+// directory onto itself returns an error (observed on Linux, though not
+// reproducible with os.Rename on Windows, which is why this test asserts on
+// the planned operations rather than a platform-specific rename failure),
+// that made the reap plan permanently un-completable: Checker.RenameDone
+// never sees the source disappear, so a resumed plan fails identically on
+// every restart.
+func Test_Store_ReapNoOpRename(t *testing.T) {
+	dir := t.TempDir()
+	store, err := NewStore(dir)
+	if err != nil {
+		t.Fatalf("Failed to create new store: %v", err)
+	}
+	defer store.Close()
+
+	// An older snapshot for the reap to discard.
+	createSnapshotInStore(t, store, "2-1017-1704807719996", 1017, 2, 1, "testdata/db-and-wals/backup.db")
+
+	// The newest snapshot already has its own bundled, not-yet-checkpointed WAL
+	// file -- the state a node is in immediately after receiving a snapshot
+	// from the leader -- and no incremental snapshot follows it. Its ID's
+	// timestamp component is fixed to whatever time.Now() yields right now, so
+	// that the ID snapshotName() recomputes during reap collides with it.
+	now := time.Now()
+	msec := now.UnixNano() / int64(time.Millisecond)
+	fullID := fmt.Sprintf("2-1131-%d", msec)
+	createSnapshotInStore(t, store, fullID, 1131, 2, 1, "testdata/db-and-wals/backup.db", "testdata/db-and-wals/wal-00")
+
+	n, c, err := store.Reap()
+	if err != nil {
+		t.Fatalf("Failed to reap snapshots: %v", err)
+	}
+	if exp, got := 1, n; exp != got {
+		t.Fatalf("Expected %d snapshots reaped, got %d", exp, got)
+	}
+	if exp, got := 1, c; exp != got {
+		t.Fatalf("Expected %d checkpoints made, got %d", exp, got)
+	}
+
+	snaps := mustListSnapshots(t, store)
+	if exp, got := 1, len(snaps); exp != got {
+		t.Fatalf("Expected %d snapshot in destination store, got %d", exp, got)
+	}
+
+	// Confirm the persisted plan (if any is still on disk) never contains a
+	// Src == Dst rename, regardless of what the underlying OS does with one.
+	if p, err := plan.ReadFromFile(store.reapPlanPath); err == nil {
+		for _, op := range p.Ops {
+			if op.Type == plan.OpRename && op.Src == op.Dst {
+				t.Fatalf("plan contains a no-op rename: %+v", op)
+			}
+		}
+	}
+
+	// The reap must actually have completed and produced valid, queryable data,
+	// not merely returned without error.
+	_, rc, err := store.Open(snaps[0].ID)
+	if err != nil {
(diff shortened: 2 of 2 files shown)


otoolep:
Thanks for your contribution, I will review.

otoolep:
I'm not convinced this is the right solution. It's a band aid. The real fix is to change the naming scheme so this collision simply isn't possible.

I will continue discussion on #2746.

otoolep:
Thanks for your contribution. I'm closing this for now, as I am not sure it's the right path. It may be the most practical path, if it is, we will implement. But for now, I need to think about it.

PR #2755, opened 2026-08-31
db: replace breaking-pragma regexes with ASCII prefix scanning

zxysilent:
## Motivation

`IsBreakingPragma` runs against **every SQL statement** passing through the Store's Raft gate (`PragmaCheckRequest.Check`). The old implementation kept five compiled regexes in a map and matched each statement against up to all five of them (map iteration order is random, so worst case is common).

## Change

Replace the regex map with a zero-allocation ASCII byte scanner that parses the statement prefix — `PRAGMA` keyword, optional schema qualifier (`main.`), pragma name — and matches against the known-breaking set.

### Semantics preserved

- `journal_mode`, `wal_autocheckpoint`, `synchronous`, `query_only`:   breaking only in assignment form (`name = value`); the bare query form   (e.g. `PRAGMA journal_mode`) passes through, as before.
- `wal_checkpoint`: breaking in any form (bare name, `=value`, `(arg)`).
- Case-insensitive, leading whitespace tolerated, schema qualifier   (must be non-empty) supported — all matching the original regexes.

### One deliberate fix

The original `wal_checkpoint` pattern had no word boundary, so it matched any pragma whose name *starts with* `wal_checkpoint` (e.g. `PRAGMA wal_checkpoint2`). The scanner requires a full name-token match.

Verified against 55 test cases covering all existing tests plus edge cases (empty qualifier `PRAGMA .journal_mode=1`, name boundaries, query forms).

## Benchmarks

Hygon C86 7265, `0 B/op, 0 allocs/op` for all variants:

| workload        | old (regex map) | new (scanner) | speedup |
|-----------------|-----------------|---------------|---------|
| mixed input     | ~9,800 ns/op    | ~290 ns/op    | ~34×    |
| non-PRAGMA path | —               | ~22 ns/op     | —       |

Also removes the package's `regexp` import and the one-time init cost of compiling five regexes (~65 µs, 37 KB, 285 allocs).

`DESIGN.md` updated: describes the regex-free check and now includes `query_only`, which the doc had omitted.

### DIFF
--- a/db/state.go
+++ b/db/state.go
@@ -12,7 +12,6 @@ import (
 	"net/url"
 	"os"
 	"path/filepath"
-	"regexp"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


otoolep:
Thanks for your contribution -- I will take a look.

otoolep:
I need the benchmark code added to the PR. Can you do that @zxysilent ?

db/state.go
```diff
@@ -127,25 +126,100 @@ func SynchronousModeFromInt(i int) (SynchronousMode, error) {
...
+		if b >= 'a' && b <= 'z' || b >= 'A' && b <= 'Z' || b >= '0' && b <= '9' || b == '_' {
+			i++
+			continue
+		}
+		break
+	}
+	return s[:i], s[i:]
+}
+
+func equalASCIIFold(a, b string) bool {
```
otoolep: Same functionality as https://pkg.go.dev/strings#EqualFold

db/state.go
```diff
@@ -127,25 +126,100 @@ func SynchronousModeFromInt(i int) (SynchronousMode, error) {
...
+	if len(name) == 0 {
+		return false
+	}
+
+	// wal_checkpoint is breaking in any form (bare name, =value, or (arg));
+	// the others only when used in the assignment form (name = value).
+	if equalASCIIFold(name, "wal_checkpoint") {
+		return true
+	}
+	if equalASCIIFold(name, "journal_mode") ||
```
otoolep: This should be moved to a `var` for easier maintenance.

otoolep:
>Verified against 55 test cases covering all existing tests plus edge cases (empty qualifier PRAGMA .journal_mode=1, name boundaries, query forms).

This implies that you ran tests that are not included in the PR. Can you add any new tests you ran?

otoolep:
Thanks for the updates, I will take a look.

otoolep:
Thanks for your contribution.

PR #2756, opened 2026-08-31
Store index and term in fingerprint file

### LINKED ISSUE #2747: Fast restart can adopt the index of a snapshot that was never restored, thereby hiding committed rows
A potential safety issue found via Antithesis testing at upstream master [`e1b34ed`](https://github.com/rqlite/rqlite/commit/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f), observed on a read replica in a 4-node cluster under network partitions and node kills. rqlite references below are pinned to that commit; `hashicorp/raft` references to v1.7.3, the version in `go.mod`.

When a node restarts, `Store.Open` may take the [*fast-restart*](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L703) path: rather than restoring the SQLite file from a snapshot, it keeps the file already on disk after [validating](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L651) it against the [`clean_snapshot` fingerprint](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L640). On that path it [adopts its `fsm_index` and `db_applied_index`](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L706-L712) from [`LatestIndexTerm()`](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/store.go#L336-L353), the index and term of the newest snapshot in the snapshot *directory*.

Nothing ties those two things together. The [fingerprint records the size, mod time and CRC32 of the database file](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/file_fingerprint.go#L12-L16); it carries no Raft index. So if a snapshot reaches the snapshot directory but is never ~applied~ restored to the FSM, the node comes back up advertising that snapshot's index over a database that has never seen it.

## Impact

**Safety**: The node reports `fsm_index` and `db_applied_index` values it does not hold, and every row committed in the falsely claimed range is missing from its database.

- Silent data loss on local reads. Reads served by that node at `level=none` return incomplete results with no error and no staleness signal. `freshness` does not help: it bounds the time since the node last heard from a leader, and this node still hears from one.
- Log replay can never repair the missing range. Raft's `lastApplied` and `lastSnapshot` both come up at the adopted index, so the entries below it are never re-applied, and the node's own snapshotting reports `nothing new to snapshot` forever. A later full snapshot install from the leader would repair the database, but the leader only sends one when its own log no longer reaches back to the node's claimed index. When the leader can still replicate from its log, it does: the node applies entries above the gap and the rows below the adopted index are lost for good, with every health signal green.
- A voter in this state would break durability. The observation was on a read replica, where the damage is confined to reads served by that node, but the same code path runs on voters. Such a node still counts toward quorum and can be elected leader.

## Triggers

- The node has a valid `clean_snapshot` fingerprint, so restart takes the fast-restart path.
- Raft fully persisted a snapshot into the snapshot directory but *never ~applied~ restored it to the FSM*, and that snapshot is newer than the state the fingerprint describes.
- The node restarts.

The middle condition is easier to reach than it looks. `hashicorp/raft`'s `installSnapshot` [calls `sink.Close()`, which makes the received snapshot durable and visible, before handing it to the FSM to restore](https://github.com/hashicorp/raft/blob/v1.7.3/raft.go#L1905-L1913). Anything that prevents or interrupts the restore leaves exactly this state behind, and a crash in that window is sufficient on its own.

It can also happen with no crash at all, which is how it was observed here. [`Sink.Close()` signals the reaper](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/sink.go#L254-L259), which [takes the snapshot store's write lock](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/store.go#L794); Raft then immediately [opens the snapshot it just installed](https://github.com/hashicorp/raft/blob/v1.7.3/fsm.go#L190-L193), and `snapshot.Store.Open` [fails outright against a held write lock rather than waiting](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/snapshot/store.go#L385-L387). The install fails as a direct consequence of its own `sink.Close()` having woken the reaper. Frequent snapshotting makes this much easier to hit; the reproduction ran with `-raft-snap=50 -raft-snap-int=2s`.

The failed install is by itself harmless: Raft [declines to advance `lastApplied` and responds unsuccessfully](https://github.com/hashicorp/raft/blob/v1.7.3/raft.go#L1924-L1931), and the leader retries. The damage is done only by the next restart.

## Symptom

The install fails, leaving the snapshot in the directory ~unapplied~ without the FSM being restored:

```
[raft] 2026/08/27 17:01:31 [ERROR] failed to restore snapshot: error="failed to open snapshot 19-612-1787850084136: acquiring read lock: MSRW conflict owner: reap"
```

The node then restarts and adopts that snapshot's index anyway:

```
[store] 2026/08/27 17:01:45 1 preexisting snapshots present
[store] 2026/08/27 17:01:45 detected successful prior snapshot operation, skipping restore
[store] 2026/08/27 17:01:45 clean snapshot check CRC32 matched, calculation took 879.228µs
```

From here it is pinned. The leader's append probes fail because the node's log now ends at the adopted index, and the node's own snapshot attempts find nothing new:

```
[raft] 2026/08/27 17:01:49 [WARN]  failed to get previous log: previous-index=949 last-index=612 error="log not found"
[raft] 2026/08/27 17:01:49 [ERROR] failed to take snapshot: error="nothing new to snapshot"
[raft] 2026/08/27 17:01:53 [ERROR] failed to take snapshot: error="nothing new to snapshot"
```

`nothing new to snapshot` then repeated for the rest of the run, across two further restarts that each re-adopted the same unapplied snapshot. Throughout, rows that the cluster had acknowledged and that were present on every voter returned nothing from this node, while its `/status` reported an index past them.

## Root cause

The tail of the fast-restart branch in `Store.Open`:

https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L703-L712

The branch is entered on the strength of the `clean_snapshot` fingerprint, but [`FileFingerprint`](https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/file_fingerprint.go#L12-L16) is only `ModTime`, `Size` and `CRC32`. It can answer *"is this the same file I last fingerprinted?"* but not *"which Raft index does this file correspond to?"*. The code answers the second question with `LatestIndexTerm()`, a different question about a different object, and the two agree only by convention.

The convention normally holds because `fsmRestore` invalidates the fingerprint, swaps the database, then re-fingerprints:

https://github.com/rqlite/rqlite/blob/e1b34edc2b572ed7ec02cac4030ebad57d57cb1f/store/store.go#L2791-L2803

But that only runs if `fsmRestore` runs. When the install fails before rqlite's `Restore` is ever entered (as above, [inside Raft's FSM runner at `snapshots.Open`](https://github.com/hashicorp/raft/blob/v1.7.3/fsm.go#L190-L193)), the fingerprint survives unchanged, still describing the old database, while the directory's newest snapshot has moved on.

`NoSnapshotRestoreOnStart = true` on the same path is what makes it permanent: on the next startup Raft [skips the FSM restore](https://github.com/hashicorp/raft/blob/v1.7.3/api.go#L678-L680) but [still sets `lastApplied` and `lastSnapshot` from the newest snapshot in the directory](https://github.com/hashicorp/raft/blob/v1.7.3/api.go#L639-L648), so the local log is never replayed and no new snapshot is ever built that would expose the discrepancy.

## Proposed fix

Decline the fast path whenever the snapshot directory has moved past the database file. That requires a fingerprint that knows which snapshot its file corresponds to:

```go
type FileFingerprint struct {
        ModTime time.Time `json:"mod_time"`
        Size    int64     `json:"size"`
        CRC32   uint32    `json:"crc32,omitempty"`
        Index   uint64    `json:"index,omitempty"`
        Term    uint64    `json:"term,omitempty"`
}
```

otoolep:


otoolep:
This should be the fix for issue #2747, and is built on a bunch of others preparatory changes that went in earlier.

This PR is not yet complete as it needs unit tests.

otoolep:
Unit test added.

PR #2758, opened 2026-09-02
fix-447:Consolidate SQLite driver configuration

karangupta982:
## Description

Consolidates SQLite driver configuration into a single `DriverConfig` and `NewDriverFromConfig` path.

Previously, features such as SQLite extensions, checkpoint-on-close behavior, and query logging were configured through separate drivers, which meant their `ConnectHook`s could override each other.

This change allows these connection-level behaviors to be composed through a single `ConnectHook`, so features such as extensions and query logging can coexist.

### Changes

- Add `DriverConfig` for SQLite driver configuration.
- Add `NewDriverFromConfig` to register a driver from a composed configuration.
- Consolidate checkpoint and query-tracing setup into `buildConnectHook`.
- Update `Store` to create a single configured driver when extensions or query logging are enabled.
- Remove the separate query-log driver implementation.
- Update and add driver tests for the new configuration path.
- Keep `ForeignKeyDriver` unchanged.

Follow up fix discussed in #2741


db/driver.go
```diff
@@ -26,13 +26,43 @@ const (
...
+}
+
 // Driver is a Database driver.
 type Driver struct {
 	name       string
 	extensions []string
 	chkOnClose CnkOnCloseMode
 }
 
+// Registers a new SQLite driver under name using cfg to
```
otoolep: Comment should start with NewDriverFromConfig as per Go style.

otoolep (commented):
This looks good -- thanks. One minor change and then I will merge.

otoolep:
@karangupta982 -- can you update the changelog too. You can say thanks to yourself. :-)

Add it to the Implementation section.

karangupta982:
Thanks @otoolep ,

While updating CHANGELOG.md i ran in conflict with this line:

[PR #2756](https://github.com/rqlite/rqlite/pull/2756): Store Index and Term in Fingerprint file and check during start-up. Fixes issue [#2747](https://github.com/rqlite/rqlite/issues/2747). Thanks @rohanpadhye

So i added both of these, please let me know if there is any issue with this.

otoolep:
Thanks for your contribution @karangupta982

PR #2760, opened 2026-09-04
Plumb Term and Index through during restore

### LINKED ISSUE #2759: Failed snapshot restore can duplicate committed rows and leave nodes diverged
Testing with Antithesis on v10.3 (6144b4c5) revealed another potential safety issue (unrelated to #2747).

**AI Disclaimer**: I used Claude Fable 5 with the Antithesis skills to setup the workload, to triage the bug, and analyze the root cause (including multiple independent agent reviews to confirm the mechanism). The issue report was drafted by Claude and hand-edited by me. I have personally checked every claim and line of code referenced here.

## Impact

***Safety*** (Correctness):
- **A committed write can be applied to the database more than once** (e.g., one `INSERT` committed; two rows read back in `SELECT`)
- **Replicas diverge for the same committed index** (leader and follower disagree on data returned to client)

## Triggers and Provenance

Triggers for the issue to manifest:

- A Raft follower node that's fallen behind catching up by installing a snapshot received from the leader.
- On the follower node, after swapping the database: a race condition with the snapshot reaper, or I/O errors in any post-swap disk operation (no crash needed)

Note that this is not *caused* by the recent fixes to #2747, though b3d61e13 added one post-swap I/O operation. The race condition we observed became possible after a4d7d6db (June 2026, which released the snapshot-store lock before the post-swap scan), and other I/O errors could have caused this corruption since 6b823052 (Jan 2024, which introduced the first post-swap fail-able operation).
## Summary

When a Raft follower installs a snapshot from the leader, as part of [`fsmRestore()`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2784) rqlite replaces the follower's SQLite database file on disk — [`db.Swap()`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2820) closes the current database, deletes its files, renames the new one into place, and reopens; there is no undoing it — [and *then* does several things that can fail](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2830-L2850). By "fail" here I mean returning an error from `fsmRestore()`, not a node crash (the code is actually robust to crashes). For example, the failure we observed in the Antithesis run: one of the steps between `db.Swap()` and the restore being deemed successful scans the snapshot directory, the background snapshot reaper rewrites that same directory concurrently, and an unlucky interleaving makes the scan return an error.

When that happens, the database has already been replaced but the node's applied-index bookkeeping has not: [`fsmIdx` and `dbAppliedIdx` have not been updated](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2842-L2845), and [because Raft considers `FSM.Restore` to have failed, it does not advance `lastApplied` either](https://github.com/hashicorp/raft/blob/v1.7.3/raft.go#L1923-L1934).

The follower is now holding **new data with an old index**. It potentially re-applies log entries whose effects are already baked into the newly installed database. Those entries are not idempotent, so a re-application can silently duplicate rows. The node's content silently diverges from the rest of the cluster, and it can then be elected leader and push its corrupted database to everyone else.

## Why FSM Restore Failures (not Crashes) are Problematic

[`Store.fsmRestore`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2784) is rqlite's implementation of `raft.FSM.Restore`. Raft calls it with a stream containing the leader's snapshot.

| # | step | line | can it fail? |
|---|---|---|---|
| 1 | Write the snapshot stream to a scratch file | [2803](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2803) | yes |
| 2 | Close the stream reader (**this also releases a lock on the snapshot store**) | [2811](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2811) | — |
| 3 | Delete the "clean snapshot" marker that makes fast restarts possible | [2817](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2817) | yes |
| 4 | **`db.Swap(tmpPath, …)` — replace the live database** | **[2820](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2820)** | yes; **irreversible** |
| 5 | Look up which index/term to adopt, by scanning the snapshot directory | [2830](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2830) | **yes** |
| 6 | Write a new fingerprint file recording the database's identity | [2838](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2838) | **yes** |
| 7 | `fsmIdx.Store(li)`, `fsmTerm.Store(tm)`, `dbAppliedIdx.Store(li)` and their signals | [2842-2846](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2842-L2846) | no |
| 8 | Record the database's modification time | [2847](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2847) | **yes** |

Step 4 is the point of no return. Steps 5, 6 and 8 all come after it, all return errors, and there is no rollback and no fail-stop.

### Sample failure - Race with the concurrent snapshot reaper

Step 5 calls this — `cat.Scan` is a raw `os.ReadDir` walk, no lock:

https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/state.go#L76-L87

**Aside**: the snapshot store has its [own method](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/store.go#L338-L353) that does the same thing correctly, wrapping the scan in `mrsw.BeginRead()`. Same logic, one takes the lock and the other does not. The startup path [calls the locked one](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L651). The restore path [calls the unlocked one](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2830).

I assume the lock exists because the snapshot *reaper* concurrently compacts old snapshots: renaming WAL files between directories, checkpointing them, deleting directories, rewriting `meta.json`, renaming the survivor ([`snapshot/plan/executor.go`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/plan/executor.go)). The [`reapLoop`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/store.go#L774) background goroutine is woken through `reapCh` by `Sink.Close`, and it holds [`mrsw.BeginWriteBlocking("reap")`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/store.go#L794) while it works. Every intermediate state of that plan is a directory an unlocked scan can read. And [the scan](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/snapshot.go#L470-L498) is strict: one directory it cannot parse [fails the whole thing](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/snapshot/snapshot.go#L483-L486).

In a 30-minute Antithesis run of a 4-node cluster (3 voters + 1 read replica) with fault injection turned on, this produced **94 failed restores** across all four nodes, all of them "a file that was there a moment ago is not now":

```
38  missing data file in snapshot directory
26  data-NNN.wal ... is not a valid SQLite WAL file
10  reading meta.json: open ... no such file or directory
 6  loading CRC32 for data-NNN.wal ...: sidecar ... no such file
 5  loading CRC32 for data.db ...: sidecar ... no such file
 5  NNN-NNN.wal ... is not a valid SQLite WAL file
 3  data.db ... is not a valid SQLite database file
 1  loading CRC32 for NNN-NNN.wal ...: sidecar ... no such file
```

(`IsValidSQLiteFile` and `IsValidSQLiteWALFile` return `false` when `os.Open` fails, so "invalid file" here means "unlinked between the glob and the open".)

Note that simply switching to the locked scan in Step 5 does not solve the issue, because a `MSRW conflict` error would also result in `FSM.Restore` failing. Ideally, the lock should be held across the entire restore operation.

### HashiCorp Raft assumes FSM Restore is atomic, and aborts on error

From `hashicorp/raft` v1.7.3, [`installSnapshot` in `raft.go`](https://github.com/hashicorp/raft/blob/v1.7.3/raft.go#L1923-L1934):

```go
// Wait for the restore to happen
if err := future.Error(); err != nil {
	r.logger.Error("failed to restore snapshot", "error", err)
	rpcErr = err
	return                                    // <-- bails out here
}

// Update the lastApplied so we don't replay old logs
r.setLastApplied(req.LastLogIndex)

// Update the last stable snapshot info
r.setLastSnapshot(req.LastLogIndex, req.LastLogTerm)
```

The comment states the invariant precisely: *"Update the lastApplied so we don't replay old logs."* On the error path that line is skipped. The same happens one level down in `runFSM`'s restore closure, which [returns before updating its `lastIndex`/`lastTerm`](https://github.com/hashicorp/raft/blob/v1.7.3/fsm.go#L205-L212).

Raft's contract here is reasonable. It assumes `FSM.Restore` is atomic: either the state machine took the snapshot or it did not. rqlite's `fsmRestore` breaks that assumption by half-taking it. Raft, believing nothing happened, leaves `lastApplied` where it was.

And rqlite's apply path has no idempotence guard: [`fsmApply`](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L2509) executes the entry's SQL unconditionally. So entries already baked into the installed snapshot get executed a second time.

### Crashes are safe, errors are not

Interestingly, the restore code is actually quite durable for node crashes, just not errors returned by the I/O operations. In case of crashes, `fsmRestore` deletes the fast-restart marker *before* it swaps the database (steps 3 and 4 above), and the comment says why: "Any existing SQLite file is about to be invalid, so mark that we can't fast-restart with it." On restart, `Store.Open` [bails out of the fast path the moment the marker is missing](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L637), and the fast path is [the only thing that suppresses Raft's own startup restore](https://github.com/rqlite/rqlite/blob/6144b4c536861b6be7ded7ae75b8bbce1a9dba82/store/store.go#L720) (`raftConfig.NoSnapshotRestoreOnStart = true`).

So a node that **dies** anywhere after step 3 comes back with no marker, Raft restores the newest snapshot into the FSM from scratch, and content and index agree again. That holds even for a crash inside `db.Swap` itself, which can leave no database file at all. We observed this: one node hit the failed restore, was killed by fault injection, restarted, rebuilt cleanly, and ended up the only *correct* node in the cluster.

## What the client sees

In our testing with a 4-node cluster: node `rqlite-1` failed a restore of snapshot `105-1178` (term 105, **index 1178**). Its database had been swapped; its applied index had not moved. It seemingly re-applied entry 1178. An `INSERT` executed twice inserts two rows.

**Invariant 1 broken — a committed write appears more than once.** A client wrote one row, the cluster acknowledged it once at index 1178, and the row exists twice on that node. A `SELECT COUNT(*)` for that key returns 2 on the affected node. Any read served by that node returns a duplicate, and any aggregate over that table is wrong.

**Invariant 2 broken — replicas that applied the same committed prefix hold different content.** In our tests, content hashes diverged after the failed restore. Then *it spread*: the corrupted node was elected leader shortly afterwards, and two other nodes subsequently adopted the corrupted data. 

## Suggested fixes

1. **Make `fsmRestore` atomic with respect to Raft's view.** Resolve the index and term *before* `db.Swap`, so that everything after the swap is bookkeeping that cannot fail.

2. **Take the index and term from the snapshot actually being restored** rather than from whatever is newest on disk. Raft's `FSM.Restore` is not handed the metadata, but rqlite already plumbs index and term through the snapshot sink, so the material exists.

otoolep:


otoolep:
Hmmm, need to look into the failing tests.

rohanpadhye:
What about `createSnapshotFingerprint`: can it fail? 

https://github.com/rqlite/rqlite/blob/48ab1eef2ce50cfd42d838841ccc2fa42ff9092a/store/store.go#L2835-L2845

otoolep:
>What about createSnapshotFingerprint: can it fail?

Of course it can, I started thinking about the restart case again and that it would cover me. Blindness.

I think the bullet proof fix for this is to something in the defer statement. Only if a `nil` err is about to be returned to the caller does the code swap the database. The alternative -- undo the swap if there is an error, well the undo itself could fail.

The other option, which I don't like is to hard exit on any error after the swap. This would force a node restart, but the Reap issue would be address.

otoolep:
Ah, but that won't work because I need to calculate the fingerprint. I'm thinking the hard exit might be the right here. If I can't create the fingerprint, bail. It really would indicate something has gone wrong.

otoolep:
I'm going to approach this in phases. Fix the core issue first, then make the system robust against other errors as a follow up.

PR #2761, opened 2026-09-04
Hard exit if FSM Restore fails post database-swap

otoolep:


otoolep:
cc @rohanpadhye

Not the most elegant, but there is precedent for this. There are a few existing places in the code that are unrecoverable, and rqlite aborts. Looks like we need another.

rohanpadhye:
Seems reasonable.

As not-a-go-programmer, I didn't even know one could inspect returned values and short-circuit this way, so I'd say quite elegant actually.

otoolep:
Yeah, that the error return value is **named** in the function signature allows the code to check its value in the `defer` handler.

To be clear, I wouldn't use this approach to handle errors that can happen in normal processing. It requires two things to be true:

- A unrecoverable error, something is fundamentally broken. A disk fills up, a critical file can't be written to disk, etc.
- Attempting to handle the error could leave the system in an even worse state.

Here is another example in the code: https://github.com/rqlite/rqlite/blob/c5b888f9125371017be8f18f81db9f08ec8c00e6/store/store.go#L2039

In this case failing to mark that a full snapshot is needed is bad. It means the entire Snapshotting process is blocked from that point on. And all this function does is write a file to disk, so if rqlite can't do that, we're dead.

It's a bonus if restarting the node may clear the condition, but that is not a required condition for this approach. It's about damage limitation.

PR #2768, opened 2026-09-06
feat/447-Add-configurable-slow-query-logging

### LINKED ISSUE #447: Support slow query log

karangupta982:
Adds configurable slow-query logging for SQL statements.

## Changes

- Adds the `-query-log` command-line flag.
- Supports disabling query logging with an empty value.
- Supports logging to stderr or a JSON configuration file.
- Adds configurable minimum query duration.
- Adds optional expanded SQL logging.
- Rejects invalid or incomplete query-log configuration.
- Ensures queries are logged when their duration meets the configured threshold.
- Adds tests covering configuration parsing and query-log behavior.

Closes: #447


cmd/rqlited/query_log.go
```diff
@@ -0,0 +1,125 @@
...
+
+import (
+	"encoding/json"
+	"fmt"
+	"log"
+	"os"
+	"time"
+)
+
+const (
```
otoolep, line 11: This functionality is important, but as much as possible should be moved on to the QueryLog type (as musch as possible). There shouldn't be this much of functionality in the `cmd` package.

cmd/rqlited/query_log_test.go
```diff
@@ -0,0 +1,225 @@
+package main
+
```
otoolep, line 2: This is kinda overkill. QueryLog configuration should be a pretty simple type, which load and loading methods on it.

db/query_log.go
```diff
@@ -13,6 +13,16 @@ import (
 type QueryLogConfig struct {
```
otoolep, line 13: Validation, if needed, is much better placed on the type itself. I know it's a judgement call, but there are too many free-floating functions to check command line-flags.

store/db_config.go
```diff
@@ -15,6 +16,15 @@ type DBConfig struct {
 
 	// Controls query logging. If nil, query logging is disabled.
 	QueryLogger *log.Logger
```
otoolep, line 18: Consolidate it. There should be single QueryLog config object, and validation methods should be on it.

otoolep:
This looks like the right direction, but it's the wrong approach stylistically. Use the type system as much as possible, not free floating validation functions. There is too much logic in the `cmd` package.

cmd/rqlited/query_log.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// queryLogDefaultMinDuration is applied when JSON config omits the "duration" field.
+	queryLogDefaultMinDuration = 10 * time.Second
+
+	// queryLogDefaultExpandedSQL is applied when JSON config omits the "expanded_sql" field.
+	queryLogDefaultExpandedSQL = true
+)
+
+// queryLogConfig holds the resolved query-log configuration ready for use by the Store.
+type queryLogConfig struct {
```
otoolep, line 23: This should be an exported type, with proper JSON exported fields, etc. A model to follow is CDC config. https://github.com/rqlite/rqlite/blob/master/cdc/config.go

query logging might need to its own package. If it doubt, start small.

otoolep:
Let me know if you've got any questions, happy to discuss more here if needed. But CDC config is of the shape I would prefer. Query Logging might be sophisticated and important enough, worthy of it's own package.

It's easy to go overboard here, I realise, but right now the presence of so much new code in `cmd/rqlited` tells us we're not there yet. I suggest we add a new package `db/querylog` and look into moving the code into there. What would I expect to be in there? I would expect:
- a type defining the query logging config, JSON attributes, validation.
- a second type, config by the config type, which has a function, suitable as a callback for SQLite (or as close as possible)
- tests

The DB drivers would then expect an instance of such a type. The command line flag would simply take a string, and that string is passed to code in `db/querylog`. It would contain the magic for figuring out if it's stderr, or a file. That's my sense anyway.

Sometimes these things only become apparent when we actually code, which is what is happening now. A sign we're on the right track is when we decide to add a new config param to query logging, and the only code that changes is in `db/querylog`.

karangupta982:
Thanks for the direction @otoolep , I have moved the query logging code into a new db/querylog package. The package now handles the configuration, validation, logger, and tests. cmd/rqlited only passes the flag value to querylog.New() and sends the logger to the store. store.DBConfig now keeps the logger directly. Please let me know if this structure looks right.

otoolep:
Thanks @karangupta982 -- I'll take a look at this in the near future, this looks better.

db/querylog/config.go
```diff
@@ -0,0 +1,139 @@
...
+//   - absent "expanded_sql" → default (true)
+//
+// Example JSON:
+//
+//	{"duration": "10s", "expanded_sql": true}
+type Config struct {
+	// Duration is the minimum execution duration before a statement is logged.
+	// Go duration syntax: "10s", "500ms", "1m". Defaults to DefaultMinDuration
+	// when absent. An empty string is rejected.
+	Duration *string `json:"duration"`
```
otoolep: Why isn't this  a `time.Duration`? CDC config uses actual time types. See https://github.com/rqlite/rqlite/blob/master/cdc/config.go#L86

db/querylog/config.go
```diff
@@ -0,0 +1,139 @@
...
+
+	// ExpandedSQL controls whether the expanded SQL (with bound parameters
+	// filled in) is preferred over the original statement text. Defaults to
+	// true when absent.
+	ExpandedSQL *bool `json:"expanded_sql"`
```
otoolep: Then make it a non-pointer bool and call it NoExpandedSQL. That's the style adopted in the rest of the codebase.

db/querylog/querylog.go
```diff
@@ -0,0 +1,130 @@
...
+	"sync"
+	"time"
+
+	"github.com/mattn/go-sqlite3"
+)
+
+// LoggerConfig holds the resolved runtime parameters used to construct a QueryLogger.
+// It is the constructor argument for NewQueryLogger, and is also used directly by
+// tests that need to build a logger without going through the flag/file path.
+type LoggerConfig struct {
```
otoolep: Why do we have two Config types? We shouldn't need two. CDC, which is much more complicated than this module, has a single config type. Or am I missing something?

otoolep (commented):
The presence of two Config objects is a sign this code isn't quite constructed properly yet.

karangupta982:
Hi @otoolep ,
You are right, one type is enough. I splited types because I was separating concerns, Config for the JSON loadable shape and LoggerConfig for the resolved runtime values, since the logger holds a real *log.Logger that doesn't come from JSON. But following the CDC pattern, I have merged them into a single Config, which is cleaner and removes the extra type.

Also switched MinDuration to time.Duration and renamed ExpandedSQL *bool to NoExpandedSQL bool as you suggested.
The CI failure appears unrelated to the querylog changes. It is Test_SingleNodeQueryTimeout, which looks like a timeout related failure.

otoolep:
Can you merge master into this branch to address the CI failure?

karangupta982:
Yes, Merged master into the branch and all checks are green now. Let me know if anything else is needed.

db/querylog/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+	stderrValue = "stderr"
+)
+
+// Config holds the query-log configuration. It is both the JSON config shape
+// and the argument to NewQueryLogger.
+type Config struct {
+	// MinDuration is the minimum query duration to log. Zero logs everything.
+	// If unset, DefaultMinDuration is used.
+	MinDuration time.Duration `json:"min_duration"`
```
otoolep, line 24: Without `omitempty` you can't parse a Config object without it. Is that on purpose?

db/querylog/config.go
```diff
@@ -0,0 +1,82 @@
...
+
+	// NoExpandedSQL uses the original SQL text instead of expanded SQL.
+	NoExpandedSQL bool `json:"no_expanded_sql,omitempty"`
+
+	// Logger is the log destination. Not read from JSON; nil disables logging.
+	Logger *log.Logger `json:"-"`
```
otoolep, line 30: Why is this in the Config? The logger should be in the QueryLogger type.

otoolep:
I took a quick look, but this still looks a bit confusing so I stopped. New() returns a new Config, but NewQueryLogger returns a new Query Logger. But the package is called "query log".

This is not consistent with Go style. Again, look at the `cdc` package. There is a `NewConfig` function and a `NewService` function.

I would expect the `query_log` package to have the following functions (something like this):

```
- DefaultConfig(....) // Default config
- NewConfigFromFile(.....) // Config from file, or perhaps a io.Reader
- New(config) // Return actually logger, so that the code is logger := query_log.New(config)
```

karangupta982:
Hi @otoolep , I have refactored the query logging implementation following the cdc package pattern:

- DefaultConfig() for the default configuration
- NewConfig(string) for loading the configuration and handling stderr/file input
- New(config) for creating the QueryLogger

The runtime logger is now separate from the configuration, and the query logging logic has been moved into the new db/querylog package. I’ve also updated the DB driver and tests accordingly. All tests are passing.

otoolep:
Thanks, this looks pretty good. I will merge in the near future, but may revert the flags changes. I may not yet want to expose it just yet.

otoolep:
@karangupta982 - if you want to push up this diff, but just remove the command line flag changes we can merge. Once it's not exposed to users we can merge and continue to build and test it.

Once we're eventually happy, then we expose the flags.

otoolep:
BTW, if you're interested in writing more code, perhaps you add support for a `hide: true` flag to the flag generation tool. See https://github.com/rqlite/flagforge. E.g.

```
[[flags]]
name = "HTTPAddr"
cli = "http-addr"
type = "string"
default = "localhost:4001"
short_help = "HTTP server bind address"
section = "HTTP API"
hide = true
```

If set, then the config remains in place, it just doesn't generate the Go or HTML output. And then we the time comes to enabling the feature, we simply flip the flag and run the tool again.

otoolep:
Thanks for your contribution.

karangupta982:
Hi @otoolep , apologies for the delay! I was traveling for a few days and could not get to this sooner. BTW, I have raised the PR(https://github.com/rqlite/flagforge/pull/5) to add hide support in the flagforge repo.

PR #2769, opened 2026-09-06
http: allow multiple bind addresses

### LINKED ISSUE #2744: Allow binding of multiple interfaces (but not 0.0.0.0)
**Is your feature request related to a problem? Please describe.**

I want to have rqlite listening on my tailscale interface ip (so that other hosts can securely get to the cluster) and localhost (so local queries dont route through tailscale layer)

i can only choose 0.0.0.0 to get both. (I tried chaining more http-addr and seems last wins)

but if I do 0.0.0.0 then I also get my public internet gateway IP and then I have to do more stuff to prevent outsiders from access to the cluster.

**Describe the solution you'd like**

allow chaining of http-addr on command line; environment variables can be a space/comma seperated list

**Describe alternatives you've considered**

I build a firewall rule on each node as a workaround.  Right now this is what I am relying on since my firewall is deny-all and tailscale works around that natively --- but I worry that I might one day forget and when I change firewall to allow all suddenly my rqlite cluster is a security hole that I had no idea was part of the attack surface.

The other idea I had was to listen to localhost only and do caddy tricks to make it work (adds DNS complexity and a lot of other stuff for a workaround)

**Additional context**

keydb has a simlar issue and in their conf file you can do "bind {tailscale_ip} localhost" and it does the right thing.

atirna:
## Why

Binding `-http-addr` to `0.0.0.0` exposes every interface when a node only needs selected ones, such as Tailscale and localhost.

## Changes

- Allow `-http-addr` to take a comma-delimited list of bind addresses.
- Keep one HTTP handler and service state across those listeners.
- Require a single explicit `-http-adv-addr` when more than one bind address is configured.

## Verification

- `go build ./...`
- `go test ./...`
- `RQLITED_PATH=<rqlited> python3 -m unittest single_node.TestMultipleHTTPAddresses`

Fixes #2744

### DIFF
--- a/cmd/rqlited/config_flags.go
+++ b/cmd/rqlited/config_flags.go
@@ -18,7 +18,7 @@ type Config struct {
 	ShowVersion bool
 	// Unique ID for node. If not set, set to advertised Raft address
 	NodeID string
-	// HTTP server bind address. To enable HTTPS, set X.509 certificate and key
+	// Comma-delimited HTTP server bind addresses. To enable HTTPS, set X.509 certificate and key
 	HTTPAddr string
 	// Advertised HTTP address. If not set, same as HTTP server bind address
 	HTTPAdv string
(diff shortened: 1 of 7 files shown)


otoolep:
Thanks for the contribution, but I am taking a very different approach to delivering this feature -- more fundamental refactoring.

atirna:
makes sense, a fundamental refactor is the better call for this one. thx for looking.

PR #2771, opened 2026-09-09
Refactor busy timeout methods to separate read-write and read-only functionality

zxysilent:
The current `BusyTimeout()` method reads the `busy_timeout` value from both the read-write and read-only database connections. However, checkpoint operations only need the read-write timeout, so querying the read-only connection introduces an unnecessary connection acquisition and PRAGMA query.

This change splits the timeout lookup into `RWBusyTimeout()` and `ROBusyTimeout()`:

- Checkpoint paths now read only the RW timeout, avoiding unnecessary RO connection access.
- The existing `BusyTimeout()` method is preserved for backward compatibility.
- Callers can now explicitly query the timeout for either the RW or RO connection.
- Existing timeout configuration, checkpoint behavior, and timeout restoration remain unchanged.
- Tests cover the RW-specific, RO-specific, and backward-compatible APIs.

This reduces unnecessary database operations during checkpointing while preserving the existing public API.

### DIFF
--- a/db/db.go
+++ b/db/db.go
@@ -733,13 +733,30 @@ func (db *DB) SetBusyTimeout(rwMs, roMs int) (err error) {
 	return nil
 }
 
-// BusyTimeout returns the current busy timeout value.
+// RWBusyTimeout returns the current busy timeout value for the read-write
+// database connection.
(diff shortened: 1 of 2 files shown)


otoolep:
Thanks for your contribution.

PR #2772, opened 2026-09-09
Don't delete the old database until the new one is actually open

goingforstudying-ctrl:

Depends on #2773, which contains the standalone `StashFiles`, `PopFiles`, and `RemoveStashedFiles` functions requested in review. This branch includes those commits through a merge; the foundation should be reviewed and merged first.

`Swap` currently deletes the original SQLite database before renaming and opening the replacement. A failed rename or open can therefore lose the original files. This change stashes the database and its WAL/SHM files until the replacement and its checkpoint manager are ready.

On a failed replacement, the rollback restores the original files and reopens the same `SwappableDB` with its original driver, foreign-key/WAL settings, read-only connection limit, synchronous mode, and a matching checkpoint manager. Stash conflicts preserve existing backup files. Rollback, reopen, or cleanup failures during rollback are included in the returned error; files are not moved if closing a replacement fails. Once installation succeeds, a failure to remove the stash is logged and does not trigger rollback of a partially deleted stash. Such committed leftovers are tracked separately and cleanup is retried on the next swap before closing the active database; stashes retained after a failed rollback are never automatically discarded by this retry.

Store startup removes stashed files after its existing clean-snapshot/Raft startup decision. It does not infer that a swap completed just because a main database file exists. The boot/ReadFrom path clears its CDC registration flag on failure as well as success, so hooks are reinstalled on the next write after rollback. The existing hard exit after a failed restore is retained: this does not guarantee recovery from arbitrary I/O or close failures, preserve connection-local hooks, or provide an atomic crash-recovery transaction.

Tests cover successful WAL swaps, preserved connection limits, real malformed SQLite replacements, unavailable drivers, stash conflicts, self-swaps, explicit rollback cleanup errors, and continued reads/writes/checkpointing through the same wrapper after recoverable failures. Store tests cover cleanup both when reusing a validated database and rebuilding one, and actual CDC events before and after a failed boot swap. Additional tests cover retrying a rolled-back swap and recovering after a committed-stash cleanup obstruction is removed. The cross-filesystem rename test is retained for Linux and is skipped locally because macOS has no `/dev/shm`.

Validation on macOS arm64, Go 1.26.4:
- `GOMAXPROCS=2 go test -p 2 ./db ./store -count=1` passed at 8e11c2d6, before the final CDC/cleanup-retry/synchronous-mode follow-up.
- `GOMAXPROCS=2 go test -p 2 ./db -count=1` on the final implementation.
- `GOMAXPROCS=2 go test -race -p 2 ./db ./store -run 'Test_(Swap|SingleNodeBoot|StoreCDC|CreateDBOnDisk)' -count=1` on the final implementation.
- `GOMAXPROCS=2 go test -race -p 2 ./db -run 'Test_(Swap|Stash|MoveSQLite|RemoveStashed)' -count=1`
- `GOMAXPROCS=2 go vet -p 2 ./db ./store`
- `git diff --check`

Related: #2103. This does not revert the failure handling introduced by #2761.


otoolep:
Did an Agent provide this analysis? It's OK, but please declare it if so.

db/swappable_db.go
```diff
@@ -58,29 +64,106 @@ func (s *SwappableDB) Swap(path string, fkConstraints, walEnabled bool) error {
...
+// swapAsidePath returns the path the database files are moved to while a swap is
+// in progress.
+func swapAsidePath(dbPath string) string {
+	return dbPath + ".swap-old"
+}
+
+// moveFilesAside moves the SQLite database at dbPath, along with any WAL and SHM
+// files, so that their names begin with asidePath. Stale files at asidePath are
+// removed first. If any move fails, moves already made are rolled back.
+func moveFilesAside(dbPath, asidePath string) error {
```
otoolep: OK, I see what you're doing. 

This function is specific to SQLite data files. We should make it an exported function like `RemoveFiles` and unit test it separately. What you're proposing now looks like it will work, but it's made up of standalone functions with weak connections in the code.

I suggest we call the required functions, add both to state.go:

```Go
StashFiles(dbPath string) error

PopFiles(dbPath string) error
```
If `PopFiles` is called and there are no files to pop, it probably returns without doing anything, and no error. I think. I also think they share the same suffix pattern, that it is not passed in. I think.

(Using names inspired by Git).

otoolep: Add those to `state.go` along with unit tests. Write a separate PR, and we merge it. Once it's merged, this fix should be cleaner. If you disagree, let me know.

otoolep:
We might also need `RemoveStashedFiles(dbPath string)` which is called by the Store on startup.

otoolep:
And once we finally are sure that a failed swap, i.e. one that returns an error. leaves everything as it was before we can revert this hard exit: https://github.com/rqlite/rqlite/pull/2761/

goingforstudying-ctrl:
Yes, this is an AI-assisted contribution. The account owner is a human developer who uses coding agents to help with analysis and programming. This reply is also being posted with AI assistance on their behalf.

db/swappable_db.go
goingforstudying-ctrl: Split this out into #2773 as requested: `StashFiles`, `PopFiles`, and `RemoveStashedFiles` are in `db/state.go`, use a fixed `.stash` naming scheme, and have separate unit tests. `PopFiles` with no stash is a no-op. Existing destination files are not overwritten; failed moves attempt rollback and report rollback errors too.

This PR now uses those helpers through [3371fb66](https://github.com/rqlite/rqlite/commit/3371fb66b12a49090587646d154a4609d133e342), including Store startup cleanup. #2773 should merge first. Tests cover partial file sets, real SQLite WAL data, failed swaps and rollback errors, continued use/retry of the same wrapper, synchronous-mode preservation, and CDC events after a failed boot swap. The final local `db` suite, targeted `db`/`store` race tests, and `go vet` pass.

The hard exit from #2761 is still in place. These changes do not guarantee that every error leaves all state unchanged, or make the file renames atomic across a crash.

PR #2773, opened 2026-09-09
Add helpers for stashing SQLite database files

goingforstudying-ctrl:

This separates the SQLite file operations requested in https://github.com/rqlite/rqlite/pull/2772#discussion_r3968456941 into `db/state.go`: `StashFiles`, `PopFiles`, and `RemoveStashedFiles`, with a shared `.stash` naming scheme.

The helpers move the database and any WAL/SHM sidecars that exist. Pop with no stash is a no-op. Neither stash nor pop overwrites existing destination files, including sidecars without a corresponding source, so callers explicitly decide when to discard files. A failed rename rolls back earlier moves and reports rollback failures instead of hiding them. Callers must close the database and exclude concurrent access; the three renames do not constitute an atomic or crash-durable transaction.

Tests cover all file-presence combinations, destination conflicts, invalid source types, partial moves and rollback failures, idempotent removal, and round trips of real SQLite databases in DELETE and WAL modes with uncheckpointed WAL data.

Validation on macOS arm64, Go 1.26.4:
- `GOMAXPROCS=2 go test -p 2 ./db -run 'Test_(Stash|RemoveStashed|MoveSQLite)' -count=1`
- `GOMAXPROCS=2 go test -p 2 ./db -count=1`
- `git diff --check`

This PR only adds the independently tested primitives. The swap integration remains in #2772 and depends on this PR.


db/state.go
```diff
@@ -509,6 +509,81 @@ func RemoveFiles(path string) error {
...
+// StashFiles moves the SQLite database file and any associated WAL and SHM files
+// aside to dbPath + ".stash" (with "-wal" and "-shm" appended for the sidecars).
+// Missing files are ignored. An existing stash is never overwritten.
+//
+// The caller must close the database and prevent concurrent access to the files.
+// The renames are not atomic as a group or guaranteed durable across a crash. If
+// a rename fails, earlier renames are rolled back; rollback errors are returned
+// along with the original error, and files may then remain at either location.
+func StashFiles(dbPath string) error {
+	return moveSQLiteFiles(dbPath, dbPath+stashedFilesSuffix, os.Rename)
```
otoolep, line 525: Why go to the trouble of passing a function to `moveSQLiteFiles`?

db/state.go
```diff
@@ -509,6 +509,81 @@ func RemoveFiles(path string) error {
...
+// database and its WAL and SHM files untouched. Missing stash files are ignored.
+func RemoveStashedFiles(dbPath string) error {
+	return RemoveFiles(dbPath + stashedFilesSuffix)
+}
+
+func moveSQLiteFiles(from, to string, rename func(string, string) error) error {
+	suffixes := []string{"", "-wal", "-shm"}
+	var present []string
+	for _, suffix := range suffixes {
+		info, err := os.Lstat(from + suffix)
```
otoolep, line 546: Use utils in `internal/fsutil` package

db/state.go
```diff
@@ -509,6 +509,81 @@ func RemoveFiles(path string) error {
...
+		if os.IsNotExist(err) {
+			continue
+		}
+		if err != nil {
+			return err
+		}
+		if !info.Mode().IsRegular() {
+			return fmt.Errorf("not a regular SQLite file: %s", from+suffix)
```
otoolep, line 554: This error message is not correct. `IsRegular` tells you nothing about whether it's a SQLite file. If you want to perform this check -- and it's reasonable -- use the functions in the DB package that already do this.

db/state.go
```diff
@@ -509,6 +509,81 @@ func RemoveFiles(path string) error {
...
+		present = append(present, suffix)
+	}
+	if len(present) == 0 {
+		return nil
+	}
+
+	// Check every destination, even if its source is absent, to avoid combining
+	// a database with WAL or SHM files from a different database.
+	for _, suffix := range suffixes {
+		if _, err := os.Lstat(to + suffix); err == nil {
```
otoolep, line 565: Use fsutl.

db/state_test.go
```diff
@@ -965,3 +966,202 @@ func mustGzip(dst, src string) {
 		panic(err)
 	}
 }
+
+func Test_StashPopFiles(t *testing.T) {
+	for mask := 0; mask < 8; mask++ {
```
otoolep, line 971: What is this all about?

db/state_test.go
```diff
@@ -965,3 +966,202 @@ func mustGzip(dst, src string) {
 		panic(err)
 	}
 }
+
+func Test_StashPopFiles(t *testing.T) {
```
otoolep, line 970: Add a short GoDoc explaining what each test is checking.

db/state.go
goingforstudying-ctrl: Removed the parameter. It was only there so the rollback test could inject failures; moveSQLiteFiles now uses a package-level renameFile variable that defaults to os.Rename and the test overrides it. Pushed in 09d874f.

db/state.go
goingforstudying-ctrl: The db validators are in now: IsValidSQLiteFile for the database file, IsValidSQLiteWALFile for the WAL. There is no WAL-index validator in the db package, so the -shm path keeps the regular-file check only, with honest messages. Empty WALs still pass, since a truncating checkpoint leaves one behind. Pushed in 09d874f.

db/state_test.go
goingforstudying-ctrl: That's a bitmask loop over the three SQLite files: bit i selects db, wal, or shm, so all 8 file combinations get a subtest. I renamed the subtests (db+wal, wal+shm, none, ...), added a comment explaining the mask, and wrote GoDoc for each test. Pushed in 09d874f.

otoolep: Too complicated. Turn it into a table-driven tests.

Tests have to be simple or you risk them not working, or testing the wrong thing.

goingforstudying-ctrl: Rewrote the stash tests as table-driven cases and dropped the bitmask-name helper. Pushed in 7a327ff.

PR #2775, opened 2026-09-11
Vendor BoltDB raft store into internal/raftdb with faster MsgPack codec

zxysilent:
Copy github.com/rqlite/raft-boltdb/v2 into internal/raftdb so the MsgPack hot path can be optimized, and switch store/log to use it.

The encode/decode helpers previously built a fresh codec.MsgpackHandle and wrapped the payload in a bytes.Buffer on every call, paying the per-type reflection cost for each raft log stored or retrieved. A single package-level handle is shared instead (safe for concurrent use: the handle's lazily-built codec function cache is copy-on-write behind a mutex, published via atomic.Value), and encoding/decoding goes through codec.NewEncoderBytes/NewDecoderBytes, which work zero-copy directly on the byte slice.

Also remove the go-metrics instrumentation copied from upstream: rqlite never installs a go-metrics sink (observability goes through expvar/OTLP), so every MeasureSince/AddSample call in GetLog/StoreLogs was pure overhead - a []string key allocation plus time.Now per call.

Switch go-msgpack v1 to v2, matching hashicorp/raft v1.7.x. This is required for correctness, not just consistency: v2's decoder understands the legacy time.Time encoding (BinaryMarshaler form) emitted by go-msgpack v0.5.5, which is the format found in raft.db files written by older rqlite releases. The v1 decoder rejects that format. Encoding is byte-identical between v1.1.5 and v2.1.5, so the write path is unchanged.

On-disk format is unchanged and byte-for-byte identical to upstream (cross-encode/decode tests assert this), so existing raft.db files remain fully readable. Upstream github.com/rqlite/raft-boltdb/v2 is no longer a dependency.

Benchmarks (linux/amd64, Hygon C86 7265, benchtime=3s):

| Benchmark     | upstream ns/op | this branch ns/op | speedup          | upstream B/op | branch B/op | B/op delta | upstream allocs | branch allocs | allocs delta |
| ------------- | -------------: | ----------------: | ---------------- | ------------: | -----------: | ---------- | --------------: | ------------: | -----------: |
| encodeMsgPack |           3324 |              1459 | **2.3x (-56%)**  |          1584 |        1296 | **-18%**   |              22 |             5 |   **-77%**   |
| decodeMsgPack |           4512 |              1818 | **2.5x (-60%)**  |          1792 |         744 | **-58%**   |              24 |             6 |   **-75%**   |

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,13 +9,13 @@ require (
 	github.com/aws/aws-sdk-go-v2/feature/s3/manager v1.22.44
 	github.com/aws/aws-sdk-go-v2/service/s3 v1.107.3
 	github.com/hashicorp/go-hclog v1.6.3
+	github.com/hashicorp/go-msgpack/v2 v2.1.5
 	github.com/hashicorp/raft v1.7.3
 	github.com/klauspost/compress v1.19.2
 	github.com/mattn/go-sqlite3 v1.14.50
 	github.com/mkideal/cli v0.2.7
 	github.com/mkideal/pkg v0.1.3
 	github.com/peterh/liner v1.2.2
-	github.com/rqlite/raft-boltdb/v2 v2.0.0-20230523104317-c08e70f4de48
 	github.com/rqlite/rqlite-disco-clients v0.0.0-20250205044118-8ada2b350099
 	github.com/rqlite/sql v0.0.0-20260224021119-1b2524a41372
 	go.etcd.io/bbolt v1.5.0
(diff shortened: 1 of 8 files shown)


otoolep:
Thanks for your contribution. Unless this actually makes a difference to throughput I may not merge it. Vendoring means I have to pick up changes manually to code.

fsync dominates rqlite performance. This probably won't help with that.

PR #2779, opened 2026-09-13
db: only dump the indexes and triggers of the selected tables

### LINKED ISSUE #2777: db: dump of selected tables cannot be restored
Up front: I worked this out with a coding agent. Happy to reshape anything if you want it
done differently.

A SQL dump taken with a table filter can't be loaded back into an empty database. The dump
carries the indexes, triggers and views of every other table, so it dies on the first
object that points at a table that isn't in the dump.

Repro:

```sql
CREATE TABLE t1 (id INTEGER PRIMARY KEY, name TEXT);
CREATE TABLE t2 (id INTEGER PRIMARY KEY, v TEXT);
CREATE TABLE t3 (id INTEGER PRIMARY KEY, w TEXT);
INSERT INTO t1(id,name) VALUES(1,'a');
CREATE INDEX idx_t1_name ON t1(name);
CREATE INDEX idx_t2_v ON t2(v);
CREATE INDEX idx_t3_w ON t3(w);
CREATE TRIGGER trg_t1 AFTER INSERT ON t1 BEGIN UPDATE t1 SET name='x' WHERE id=NEW.id; END;
CREATE TRIGGER trg_t2 AFTER INSERT ON t2 BEGIN UPDATE t2 SET v='y' WHERE id=NEW.id; END;
CREATE VIEW v_t2 AS SELECT id, v FROM t2;
CREATE VIEW v_t1_t2 AS SELECT a.id FROM t1 a JOIN t2 b ON a.id=b.id;
```

`curl -G 'localhost:4001/db/backup?fmt=sql&tables=t1'` writes this (SQL format):

```sql
PRAGMA foreign_keys=OFF;
BEGIN TRANSACTION;
CREATE TABLE t1 (id INTEGER PRIMARY KEY, name TEXT);
INSERT INTO "t1" VALUES(1,'a');
CREATE INDEX idx_t1_name ON t1(name);
CREATE INDEX idx_t2_v ON t2(v);
CREATE INDEX idx_t3_w ON t3(w);
CREATE TRIGGER trg_t1 AFTER INSERT ON t1 BEGIN UPDATE t1 SET name='x' WHERE id=NEW.id; END;
CREATE TRIGGER trg_t2 AFTER INSERT ON t2 BEGIN UPDATE t2 SET v='y' WHERE id=NEW.id; END;
CREATE VIEW v_t2 AS SELECT id, v FROM t2;
CREATE VIEW v_t1_t2 AS SELECT a.id FROM t1 a JOIN t2 b ON a.id=b.id;
COMMIT;
```

Loading it into an empty database:

```
$ sqlite3 new.db < dump.sql
Parse error near line 6: no such table: main.t2
Parse error near line 7: no such table: main.t3
Parse error near line 9: no such table: main.t2
$ echo $?
1
```

The two views still get created on the way, so what's left is a partial schema pointing at
tables that aren't there. Through the API, `db.ExecuteStringStmt(dump)` comes back with one
response carrying `error = "no such table: main.t2"`.

The reason is that only the tables themselves are filtered (`DumpTablesReq`), while the
indexes, triggers and views are written from the whole `sqlite_master`. The code says it:

```go
// For indexes, triggers, and views, we could add more sophisticated filtering
// based on the table they relate to, but for now include all of them
```

What I'd expect from a filtered dump: only the objects that belong to the selected tables,
or the ones that can be built on top of them. An index or a trigger of a table nobody asked
for shouldn't be in there. A view should only be in there when everything it reads from is
in there too, including views that read from other views. A dump without a filter should
keep emitting everything, the way it does now.

The part I don't know how you'd want it solved is the views. Indexes and triggers can be
tied to a table with `sqlite_master.tbl_name`, views carry nothing like that. I can see two
ways: parse the view SQL with `github.com/rqlite/sql`, or let SQLite answer it (create the
selected tables in a scratch in-memory database, then create and probe each view, keeping
the ones that resolve). The first keeps everything on the current connection, the second
handles quoting, CTEs and subqueries without hand-rolling a parser.

Which one do you prefer? If either is fine, I'll send a PR with tests.

geiltonxavier:
Up front: this patch came out of a session with a coding agent.

Fixes #2777.

### What it does

An index or a trigger is only dumped when the table it belongs to was selected
(`sqlite_master.tbl_name`), so a dump of selected tables no longer carries the schema objects of
the whole database. Without a table filter nothing changes.

Views are written out the way they are today. That's the separate question from the issue, so this
change deliberately leaves the view statements alone.

### On the question in the issue

Filtering the indexes and triggers is enough for the filtered dump to load. SQLite resolves the
tables a view reads from when the view is used, not when it is created, so a view that reads from a
table the dump doesn't carry doesn't stop the dump from loading: it just leaves a view behind that
can't be queried. In the reproduction the load errors were lines 6, 7 and 9, the two indexes and
the trigger and never the two `CREATE VIEW` statements.

### Tests

`db/dump_table_selection_test.go`, with the schema from the issue:

- a dump of one table loads into an empty database and carries exactly its own index and trigger,
  its own table, and its rows;
- a dump of two tables carries the indexes and triggers of both;
- a dump of the whole database still carries every table, index, trigger and view, and still loads.

```
$ go test ./db/
ok  	github.com/rqlite/rqlite/v10/db	33.260s
$ go vet ./db/ && gofmt -l db/
(no output)
```

### Notes

Touches `db/` only.


geiltonxavier:
The only failing job is `end_to_end_auto_state`, and it looks unrelated to this change: it fails in 9.8s, before any of the tests in the job can run, and the same job also fails on #2778 while master is green. Nothing here touches the auto-backup path either (those tests do not use the `tables` parameter at all). Happy to re-run or dig in if you read it differently.

db/dump_schema.go
```diff
@@ -0,0 +1,90 @@
+package db
```
otoolep: I don't like this pattern of splitting into a distinct file (I see Agents do it all the time). It's not significant enough to warrant its own source file.

Instead Integrate the code into db.go, and put the new unit tests alongside the pre-existing "dump" unit test.

otoolep:
CircleCI is having problems, it's affecting all PRs. You can ignore that test error for now.

db/dump_table_selection_test.go
```diff
@@ -0,0 +1,167 @@
...
+
+	expTables := `[{"columns":["obj"],"types":["text"],"values":[["table:t1"],["table:t2"]]}]`
+	if got := schemaObjectsOfType(newDB, "table"); got != expTables {
+		t.Fatalf("tables of a two-table dump: got %s want %s", got, expTables)
+	}
+}
+
+// TestDumpAllTablesUnchanged checks that a dump made without naming any table is
+// untouched: it carries every table, index, trigger and view.
+func TestDumpAllTablesUnchanged(t *testing.T) {
```
otoolep: Surely this test was already present effectively.

geiltonxavier:
Both done in d03d5d3d: the code lives in db.go now (no separate file), and the tests sit next to Test_DB_Dump in db_common_test.go.

On the Views I took the third option: the doc comment on Dump now spells out that views are written out as they are, filtered or not, so a view that reads from a table the dump does not carry is created but cannot be queried, and that the dump itself still loads. If you would rather have that in the user docs, point me at it and I will send a change to rqlite.io.

Re-ran everything: go test ./db/ ok (31.3s), go vet and gofmt clean. Thanks for the note on CircleCI.

geiltonxavier:
Both review points are in 126eae82:

- the code lives in db.go now, no separate source file, and the tests sit next to Test_DB_Dump in db_common_test.go;
- you're right about that third test: Test_DB_Dump already covers the tables and their data in a full dump, so I cut it down to the only thing it wasn't covering - that an unfiltered dump still carries every index, trigger and view, since its fixture has none. Renamed the three new ones to Test_DB_* to match the file.

Re-ran: go test ./db/ ok (31.3s), go vet and gofmt clean.

db/db.go
```diff
@@ -1802,6 +1804,86 @@ func (db *DB) Dump(w io.Writer, tableNames ...string) error {
...
+	return objs, nil
+}
+
+// schemaStatements returns the statements that recreate the indexes, triggers and
+// views of the database. When tables are named, an index or a trigger is only
+// returned when the table it belongs to is one of them, so that a dump of
+// selected tables doesn't carry the schema of the whole database. Views are
+// returned as they are today: a view that reads from a table the dump doesn't
+// carry is still written out.
+func (db *DB) schemaStatements(ctx context.Context, conn *sql.Conn, tables []string) ([]string, error) {
```
otoolep: One more change. Since you've created a type, then use the type system.

```
type schemaObject struct

type schemaObjects []schemaObject

// Filter returns a new schemaObjects which only contains indexes, tables, and triggers related to the given tables.
func (s schemaObjects) Filter(tables []string) schemaObjects { ....
```

Then you can write targeted unit tests *for the filtering specifically*. You create `schemaObjects`, fill them with values, and ensure filter works.

Make sense?

db/db_common_test.go
```diff
@@ -2171,6 +2171,153 @@ func Test_DB_Dump(t *testing.T) {
...
+	expTables := `[{"columns":["obj"],"types":["text"],"values":[["table:t1"],["table:t2"]]}]`
+	if got := schemaObjectsOfType(newDB, "table"); got != expTables {
+		t.Fatalf("tables of a two-table dump: got %s want %s", got, expTables)
+	}
+}
+
+// Test_DB_DumpAllCarriesSchemaObjects checks that a dump made without naming any
+// table still carries every index, trigger and view. Test_DB_Dump already covers
+// the tables and their data, with a fixture that holds no schema objects at all.
+func Test_DB_DumpAllCarriesSchemaObjects(t *testing.T) {
```
otoolep: Why do we need this test? Was it not tested already?

db/db_common_test.go
```diff
@@ -2171,6 +2171,153 @@ func Test_DB_Dump(t *testing.T) {
...
+			t.Fatalf("loading dump: %s", r.GetError())
+		}
+	}
+	return newDB
+}
+
+// TestDumpSelectedTablesSchemaObjects checks that a dump of a single table loads
+// into an empty database, and that it only carries the indexes and triggers of
+// that table instead of those of the whole database.
+func Test_DB_DumpSelectedTablesSchemaObjects(t *testing.T) {
```
otoolep: What these tests are doing is checking the filter logic. But that logic is a type-level thing, not a database level thing. That's why the type approach is more powerful. It means simpler testing, but also clearer testing.

I agree you need a round-trip test to ensure that a filtered database dump does load correctly, but much of the code logic is testing *without* a database connection at all.

otoolep (changes requested):

otoolep:
You can merge master into your branch to fix that CI issue.

db/db.go
geiltonxavier: Makes sense, done in 9d521a8a. schemaObjects is a named type now, with Filter(tables []string) schemaObjects, and Dump writes objs.Filter(tableNames). The loader returns the named type too, so the filtering is pure and tested without a database.

One thing for later: if views ever get filtered by what they read from, that step cannot live in Filter, because deciding whether a view's inputs are present needs SQLite to answer. It would sit next to it.

db/db_common_test.go
geiltonxavier: Gone in 9d521a8a, you were right. That case is Filter with no tables returning everything, which the type test covers, and Test_DB_Dump already round trips a full dump.

db/db_common_test.go
geiltonxavier: Agreed, and that is what it is now. Test_SchemaObjects_Filter is pure and has one case per rule, and the database test is a single round trip (Test_DB_DumpSelectedTables) that proves a filtered dump loads and carries only the selected table's objects.

geiltonxavier:
Merged master in 6171579b. CircleCI is green now, including end_to_end_auto_state. AppVeyor is still pending, but it is pending on master's own commits too.

otoolep:
Thanks for your contribution.

otoolep:
I liked your blog post @geiltonxavier 

https://geiltonxavier.com/posts/open-source-in-the-age-of-ai/

geiltonxavier:
> I liked your blog post @geiltonxavier 
> 
> 
> 
> https://geiltonxavier.com/posts/open-source-in-the-age-of-ai/

Thanks, Philip, and thanks for reading the post. I believe this one can help more people who want to contribute to open source and are not sure how it works with AI. I think every step up to the merge has something that can help whoever reads it.

PR #2807, opened 2026-09-20
Snapshot IDs now contain generation

otoolep:


otoolep:
More work for https://github.com/rqlite/rqlite/issues/2746

PR #2812, opened 2026-09-21
Keep snapshot ordering correct when the clock moves backwards

### LINKED ISSUE #2809: Clock going backwards can affect correctness if new snapshot post reap has same term and index
Started with https://github.com/rqlite/rqlite/issues/2746

If the clock goes backwards *and* a reap happens which results in the same term and index, the snapshots will sort incorrectly.

goingforstudying-ctrl:

Fixes #2809.

Snapshot IDs embed a wall-clock millisecond timestamp, and `Less` orders snapshots with the same term and index by that field. If the clock moves backwards (NTP step, VM migration), a snapshot created after a reap — or one received afterwards at the same term and index — can get an earlier timestamp than the snapshot it replaces. The newer state then sorts as older, so `Newest`/`NewestFull`/`PartitionAtFull` would treat the stale snapshot as the store's latest state.

Following the approach outlined in the issue: when naming a new snapshot, find existing snapshots with the same term and index, take the largest timestamp T among them, and use the larger of (current time, T+1).

What changed:

- `SnapshotNamer.makeName` takes a minimum-millisecond floor; `MakeName` behaves exactly as before (no floor).
- `SnapshotNamer.makeNameForSet` derives the floor from a `SnapshotSet`: `WithTermIndex(term, index)` selects the matching snapshots, and the newest of them sets the floor.
- `reapInternal` reuses the set it already scanned; `Store.Create` scans the store under a read lock so the read cannot race with a reap. The generation-bump collision loop stays as a backstop.
- I left `Clone` as-is since it errors out on an ID collision rather than relying on sort order.
- DESIGN.md documents the new ordering guarantee.

Tests: new unit cases for the floor over in-memory sets, plus regression tests driving `Store.Create` and `Reap` with a fixed clock stepped backwards. The two existing NameCollision tests now expect the bumped timestamp. `go test -race ./snapshot/` passes.


otoolep:
I agree this is the right approach, but I've already started it. You can see the functionality here: https://github.com/rqlite/rqlite/pull/2810/changes#diff-5c6bcd0ff5f76d3d6e311122b52e890cff76233ee29e88a138f074bcc74518e1R382

If you want to generate a new PR using this function I just added, go for it. I think passing in a min msec is a reasonable approach, but use the function above to do it. The approach in this PR is too imperative, and is not thinking about solving the problem in terms of types.

goingforstudying-ctrl:
Reworked this PR in place rather than opening a new one. Now built on `WithTermIndex` in 7b9e928. `SnapshotNamer.makeNameForSet` takes a `SnapshotSet`, filters it with `WithTermIndex(term, index)`, and floors the new ID at the newest match's timestamp + 1, so the naming logic works through the set type instead of re-scanning the directory itself.

`reapInternal` reuses the set it already scans. `Store.Create` didn't have one, so it now scans under a read lock, which keeps the scan from racing the reaper. Min msec floor and generation-bump backstop are unchanged; `Clone` still errors on collision.

snapshot/snapshot_namer.go
```diff
@@ -24,14 +25,41 @@ func NewSnapshotNamer(nowFn func() time.Time) *SnapshotNamer {
...
 	return fmt.Sprintf("%d-%d-%d-%d", term, index, msec, gen)
 }
 
+// makeNameForSet returns a name for the Snapshot, as makeName does, but with
+// the minimum millisecond field derived from the given SnapshotSet: one more
+// than the millisecond field of the newest snapshot in the set sharing the
+// given term and index, if any. The returned name therefore sorts as newer
+// than every snapshot in the set with that term and index, even if the system
+// clock moved backwards since those snapshots were created.
+func (sn *SnapshotNamer) makeNameForSet(set SnapshotSet, term, index uint64, gen int64) string {
```
otoolep: Why does this function exist? Why not just change MakeName() to take a snapshot set?

This is an example of the problems of Agent-generated code. They have bad taste. :-(

snapshot/snapshot_namer.go
```diff
@@ -24,14 +25,41 @@ func NewSnapshotNamer(nowFn func() time.Time) *SnapshotNamer {
 // MakeName returns a name for the Snapshot, for the given the term, index, and
 // and generation. If gen is less than 1, then no generation is present in the name.
 func (sn *SnapshotNamer) MakeName(term, index uint64, gen int64) string {
```
otoolep: Alter this function to take a snapshot set. That's the missing piece I think.

```Go
func (sn *SnapshotNamer) MakeName(set SnapshotSet, term, index uint64, gen int64) string
```

goingforstudying-ctrl: Done in ab07da47. MakeName takes the set now, floor comes from the newest snapshot with the same term and index.

snapshot/snapshot_namer.go
goingforstudying-ctrl: Fair. Folded into MakeName in ab07da47, the helper is gone.

otoolep:
Great, thanks. I'll merge shortly.

PR #2819, opened 2026-09-22
Remove generation from snapshot names

otoolep:
The timestamp fix that went in with https://github.com/rqlite/rqlite/pull/2812 keeps the filename format the same, so preferable.

cc @rohanpadhye


rohanpadhye:
Amazing! (had to step through the various PRs to understand what is going on, but yeah the final scheme looks like it should work)

